PR #15, opened 2015-12-02
Updated django-guardian dependency to fix test errors with Python 2.7.

tjanez:
With Python 2.7 and django-guardian 1.3, tests on CentOS/Fedora were
failing with: "DoesNotExist: ContentType matching query does not exist."


mstajdohar:
Oh that was ugly

PR #16, opened 2015-12-02
Upgrade to Django 1.9

dblenkus:
Upgrade Django and all dependencies to latest versions.
Replace third-party JSON fields with Django's native JSONField.


tjanez:
Jenkins, test this please.

dblenkus:
@mstajdohar, @tjanez Tests are passing on Python 3.4 and 3.5, so the problem is only with Python 3.3. Do we want to support it?

tjanez:
I'm for dropping Python 3.3 support.

According to [PEP0398](https://www.python.org/dev/peps/pep-0398/) Python 3.3.5 was the last regular maintenance release before 3.3 entered security-fix only mode (it was released on March 8, 2014).

As for Linux distribution support:
- CentOS 7 has Python 3.4 (via EPEL)
- Debian 8 has Python 3.4
- Fedora 21 - 23 have Python 3.4
- Ubuntu 14.04 - 15.10 have Python 3.4

So I think it's safe to assume people will be able to run Resolwe on Python 3.4+.

mstajdohar:
Drop the Python 3.3 support.

mstajdohar:
Great job with the upgrade.

setup.py
```diff
@@ -14,16 +14,15 @@
     long_description=open('README.rst', 'r').read(),
     packages=find_packages(),
     install_requires=[
-        'django>=1.8,<1.9a1',
-        'djangorestframework>=3.1',
+        'django>=1.9,<1.10a1',
+        'djangorestframework>=3.3',
```
mstajdohar: Do we really need the 3.3 version or is this nice to have? The idea is that requirements in setup.py are as loose as possible so that other people can include our app into their Django projects even if they use older versions required modules. The same goes for django-filter and jsonschema. Of course it makes sense to raise the requirements if there were security fixes or added features that we need.

dblenkus: They introduced json field serializer in 3.3, so it is required.

We discussed this with @tjanez and decided to raise some of the requirements (not to the latest version, but to the latest minor version). The reason was that we have to test all combinations. And the second reason is that users can install resolwe in the virtualenv so it won't interfere with others' apps requirements.

mstajdohar: To the first point, that's great! To the third: if users wish to include resolwe in existing Django project, they already have some requirements and a virtualenv. So, looser is better, because some of their requirements may clash with ours. To the middle point :) It would be appropriate to test all combinations, but almost nobody does that in practice. Most open-source projecs test the latest requirements that match their setup.py. When they get bug reports from the community, they think about solutions: either to fix bugs or increase the requirements. If we would test 2 combinations: the highest versions that match the requirements and the lowest, that would already be pretty impressive. I don't think we have to worry about this much now, but we will have to in the future.

dblenkus:
@mstajdohar I've loosen some requirements, but included all that included Django 1.8 and 1.9

@tjanez I've tried "django~=1.9.0" notation and is working locally but failing on jenkins. Do we want to include all the latest features of pip, virtualenv,... or is it better to have "not so nice" code but support more environments?

mstajdohar:
In general we wish to support as much environments as possible. This will become important when we get users for the resolwe lib. Until then we have to be practical. We should limit our work to our needs and avoid extra work.

PR #21, opened 2015-12-14
Feature executors

dblenkus:


mstajdohar:
Fix tests. Add changes to change log. Improve the commit message (this will be easy after you update the change log).

kostko:
Can you describe what this pull request is implementing or link to the appropriate JIRA issue?

dblenkus:
It adds basic support for running processes. We have three concepts:
- **executor**: finds appropriate machine and sends task (i.e. celery)
- **expression engine**: compiles our code (at the moment we support bash with django template language) to be executable in bash
- **runtime**: environment where processes are run

resolwe/flow/engines/celery.py
```diff
@@ -0,0 +1,23 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import sys
+
+from . import BaseManager
+from ..tasks import celery_run
+
+try:
```
kostko: I see some issues with this:
* First, celery is already indirectly imported above when importing `..tasks`. So it will fail there.
* Second, you should not just print something on an `ImportError` and continue executing if the other code is not ready for this. Otherwise, the code could just fail later and the `print` output will simply be missed.

resolwe/flow/engines/celery.py
```diff
@@ -0,0 +1,23 @@
...
+    print('Please install Celery using `pip install celery`', file=sys.stderr)
+
+
+__all__ = ['manager']
+
+
+class Manager(BaseManager):
+
+    def run(self, data_id, script, run_sync=False, verbosity=1):
+        celery_run.async(self.executor, data_id, script, verbosity)
```
kostko: Looks like you are serializing executors here. This is ok, but I think it would be better to be explicit in what state should be serialized (using `__getstate__` on the executor classes etc.).

resolwe/flow/executors/__init__.py
```diff
@@ -1,13 +1,201 @@
...
+                process_progress=100,
+                finished=datetime.utcnow())
+        else:
+            Data.objects.filter(id=data_id).update(
+                status=Data.STATUS_ERROR,
+                process_progress=100,
+                process_rc=process_rc,
+                finished=datetime.utcnow())
+
+        # try:
```
kostko: What's up with this commented out code block?

resolwe/flow/engines/__init__.py
```diff
@@ -79,25 +83,8 @@ def communicate(self, run_sync=False, verbosity=1):
...
-                    inputs['proc'] = info  # add script info
-
-                    try:
-                        script = template.Template('{% load resource_filters %}{% load mathfilters %}' +
-                                                   script_template).render(template.Context(inputs))
-                    except template.TemplateSyntaxError as ex:
-                        data.status = Data.STATUS_ERROR
-                        data.process_error.append('Error in process script: {}'.format(ex))
-                        data.save()
+                    script = self.exprengines['dtlbash'].eval(data)
```
kostko: What happens if `dtlbash` expression engine is not configured in `FLOW_EXPRESSION_ENGINES`? Perhaps there should be a better exception than just `KeyError`.

mstajdohar: The expression that the processor use will be defined on the processor. Currently we only support the ```dtlbash```, but the platform was designed to be general, so that we can add more expression engines later. I do not think we need FLOW_EXPRESSION_ENGINES in settings at all? We should just load all available expression engines. Because ```dtlbash``` is available and because it is hardcoded, an exception can not happen here. When we will support another expression engine and read its name from the processor, we can also add the error handling.

kostko: > I do not think we need FLOW_EXPRESSION_ENGINES in settings at all? We should just load all available expression engines. Because dtlbash is available and because it is hardcoded, an exception can not happen here.

But currently, the expression engines are loaded based on the `FLOW_EXPRESSION_ENGINES` setting, so an exception is possible. See the above code which does:
```python
self.exprengines = self.load_exprengines(settings.FLOW_EXPRESSION_ENGINES)
```

mstajdohar: Yes, my suggestion is to change this to load all available expression engines.

dblenkus: I followed the concept of Django REST Framework, where you define define defaults in settings and you can override them later (i.e. in process).
@mstajdohar Where would you load expression engines from? An option is to check `exprengines` folders in all installed apps and load all `ExpressionEngine` classes. But this would be very time consuming, so it would be good to add global variable with already loaded engines (the same way as for processes in tests).

resolwe/flow/executors/__init__.py
dblenkus: @mstajdohar this is probably commented because there was no `purge` function at the time? So I should put it back?

resolwe/flow/engines/celery.py
dblenkus: fixed

resolwe/flow/engines/celery.py
dblenkus: I've added comment about serializing to `BaseFlowExecutor` class.

resolwe/flow/engines/__init__.py
mstajdohar: We have one expression engine now. Just load it. When we add the next one, we can extend the functionality. We will have to think about this more... Let's try to avoid premature optimization.

resolwe/flow/executors/__init__.py
mstajdohar: Yes

mstajdohar:
@dblenkus, have you addressed all the issues?

mstajdohar:
@dblenkus, did you write the changes to change log?

dblenkus:
Yes, I've fixed everything and just added change log.

PR #23, opened 2016-01-08
Ensure that permissions are assigned atomically

kostko:
Otherwise, the object might already be created but without permissions, which introduces a race condition with the query observers.


dblenkus:
Test are failing because of the new version of django-guardian. I've fixed version to the old one until they fix it. Please rebase to master.

kostko:
Rebased.

PR #25, opened 2016-01-11
Redirect subprocess' stderr to stdout in executors

tjanez:
Update _msg_out() to only dump stdout.txt.

### DIFF
--- a/resolwe/flow/executors/__init__.py
+++ b/resolwe/flow/executors/__init__.py
@@ -55,7 +55,7 @@ class BaseFlowExecutor(object):
         if verbosity >= 1:
             print('RUN: {} {}'.format(data_id, script))
 
-        self.data_id = data_id
+        self.data_id = str(data_id)
 
         output_path = os.path.join(settings.FLOW_EXECUTOR['DATA_PATH'], str(data_id))
         os.mkdir(output_path, 0o775)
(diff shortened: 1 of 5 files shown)


resolwe/flow/executors/__init__.py
```diff
@@ -55,7 +55,7 @@ def run(self, data_id, script, verbosity=1):
         if verbosity >= 1:
             print('RUN: {} {}'.format(data_id, script))
 
-        self.data_id = data_id
+        self.data_id = str(data_id)
```
mstajdohar: Is this OK for int id-s?

mstajdohar: We might wish to compare somewhere, but the id there might be in original form (int) that would not match the string here.

resolwe/flow/executors/docker.py
```diff
@@ -2,17 +2,30 @@
...
 
 class FlowExecutor(LocalFlowExecutor):
     def start(self):
+        container_image = settings.FLOW_EXECUTOR['CONTAINER_IMAGE']
+
+        if self.data_id != 'no_data_id':
+            container_name = 'resolwe_{}'.format(self.data_id)
+        else:
+            # set random container name for tests
+            rand_int = random.randint(1000, 9999)
```
mstajdohar: [UUID](https://docs.python.org/2/library/uuid.html) guaranties a unique id.

mstajdohar: Could some tests randomly fail on Jenkins because of this?

dblenkus: Docker container is not removed if you interrupt tests. So this is added just as convenience that you don't have to manually remove containers after each interrupted test. This will never happen on Jenkins. So I don't think that UUIDs are needed here.

resolwe/flow/executors/__init__.py
dblenkus: Agree, I will change it back.

dblenkus:
Included in #26

PR #28, opened 2016-01-21
Initial docs

mstajdohar:


docs/intro.rst
```diff
@@ -0,0 +1,7 @@
+===============
+Getting started
+===============
+
+TODO: Write about how to include Resolwe in a Django project and explain
+settings parameters. Create an example Django project in the ``docs/example``
```
dblenkus: Do we want to create another project and mantain it? We already have minimal project for running tests. Is it ok to add comments there?

mstajdohar: Here I would have a minimal project. I would have here a proper project (what the django-admin.py create project gives you). With the front-end you could play with and so on.

I don't know, but maybe I do not wish to clutter the test project with examples for the documentation.

dblenkus: Test project is exactly what you described as example project. It supports running Django development server, so you can play with the web UI.

mstajdohar: Well, I started with this project because I needed a place for the example processors that I used in the docs. I suggest you start writing the getting started and you will see what kinds of example files you will need to explain how to kick-off resolwe and all its settings. Maybe you will see the test project sufficient, maybe not. We can always shuffle this files around.

mstajdohar:
@dblenkus, I have amended the latest commit and rebased the PR.

mstajdohar:
The introduction docs are ready to accept.

PR #32, opened 2016-02-04
Include object permissions in API results

kostko:


resolwe/flow/views.py
```diff
@@ -214,6 +216,35 @@ def set_public_permissions(perm_type):
...
+            'public': format_permissions(permissions_public),
+        }
+
+    def get_serializer_class(self):
+        # Augment the base serializer class to include permissions information with objects.
+        base_class = super(ResolwePermissionsMixin, self).get_serializer_class()
+
+        class SerializerWithPermissions(base_class):
+            def to_representation(serializer, instance):
+                # TODO: These permissions queries may be expensive. Should we limit or optimize this?
```
kostko: Thoughts? This could be expensive in large lists. Perhaps we could limit it based on request query parameters.

PR #35, opened 2016-02-10
Add view to check if slug exists

dblenkus:


kostko:
> Maybe `slug_exists` route name would better describe what this method does?

Yeah, I agree with @hadalin on this, otherwise the pull request looks good to me.

dblenkus:
Renamed

hadalin:
I think [this line](https://github.com/genialis/genesis-platform/blob/master/requirements.txt#L2) should also be updated upon Resolwe update.

kostko:
It has already been updated ;-)

PR #38, opened 2016-02-21
A couple of fixes to enable running the tests of addon applications

tjanez:


resolwe/flow/tests/__init__.py
```diff
@@ -288,12 +295,12 @@ def assertFiles(self, obj, field_path, fn, compression=None):  # pylint: disable
...
             raise ValueError("Unsupported compression format.")
 
-        field = dict_dot(obj['output'], field_path)
+        field = dict_dot(obj.output, field_path)
         output = os.path.join(settings.FLOW_EXECUTOR['DATA_PATH'], str(obj.pk), field['file'])
         output_file = open_fn(output)
         output_hash = hashlib.sha256(output_file.read()).hexdigest()
 
-        wanted = os.path.join(self.current_path, 'outputs', fn)
+        wanted = os.path.join(self.files_path, fn)
```
dblenkus: Is this ok, to have outputs directly in the folder without any subfolder?

tjanez: I have no particular opinion about this, I just saw this being done in this way in [resolwe-bio](https://github.com/genialis/resolwe-bio/tree/master/resolwe_bio/tests/files) and assumed it is the way to go.

Perhaps one argument for all files being in the same directory is that some files will be an output of a process and at the same time input for several other processes.

@mstajdohar, @jkokosar do you have any thoughts on this?

PR #41, opened 2016-03-05
Fixes for render_descriptor

kostko:
Since d054afc76fd00f561b1766a45797797082cc3d11 the import process from the old system (`genesis_migrate`) is completely broken (some issues are introduced by that commit and some are just exposed by it). This pull request fixes some obvious issues with the mentioned commit, but there is also another problem.

Importing a dump currently fails with the following error when calling `hydrate_input_references`:
```
ValueError: invalid literal for int() with base 10: '54f70fb6fad58d20240baf97'
```

The issue seems to be that `render_descriptor` tries to resolve the references to imported data objects, but the Data's input JSON itself is not migrated at all, so all references are wrong. The migration process should also recursively fix the references before loading objects. But since Data objects may reference each other, we need to establish a proper ordering so we can save objects without any references first and then make our way towards objects that depend on others.

Probably the best way would be to construct a dependency graph in the first phase and then use a topological sort to get a suitable order. Then save the Data objects in order, record mappings from the old identifiers to the new identifiers and update JSON fields before saving the object.

### DIFF
--- a/resolwe/flow/models.py
+++ b/resolwe/flow/models.py
@@ -257,7 +257,7 @@ def render_descriptor(data):
     tmpl_vars = template.Context(tmpl_vars)
 
     # Set default values
-    for field_schema, _, path in iterate_schema(data.descriptor, data.descriptor_schema, 'descriptor'):
+    for field_schema, _, path in iterate_schema(data.descriptor, data.descriptor_schema.schema, 'descriptor'):
         if 'default' in field_schema:
             tmpl = field_schema['default']
             if field_schema['type'].startswith('list:'):
@@ -631,22 +631,40 @@ def dict_dot(d, k, val=None, default=None):
     if val is None and k == '':
         return d
 
+    def set_default(dict_or_model, key, default_value):
+        if isinstance(dict_or_model, models.Model):
+            if not hasattr(dict_or_model, key):
+                setattr(dict_or_model, key, default_value)
+
+            return getattr(dict_or_model, key)
+        else:
+            return dict_or_model.setdefault(key, default_value)
+
+    def get_item(dict_or_model, key):
+        if isinstance(dict_or_model, models.Model):
+            return getattr(dict_or_model, key)
+        else:
+            return dict_or_model[key]
+
+    def set_item(dict_or_model, key, value):
+        if isinstance(dict_or_model, models.Model):
+            setattr(dict_or_model, key, value)
+        else:
+            dict_or_model[key] = value
+
     if val is None and callable(default):
         # Get value, default for missing
-        # Ugly, but works for model.Data objects as well as dicts
-        # Does the same as:
-        # return reduce(lambda a, b: a.setdefault(b, default()), k.split('.'), d)
-        return reduce(lambda a, b: a.__setitem__(b, a[b] if b in a else default()) or a[b], k.split('.'), d)
+        return reduce(lambda a, b: set_default(a, b, default()), k.split('.'), d)
 
     elif val is None:
         # Get value, error on missing
-        return reduce(lambda a, b: a[b], k.split('.'), d)
+        return reduce(lambda a, b: get_item(a, b), k.split('.'), d)
 
     else:
         # Set value
         try:
             k, k_last = k.rsplit('.', 1)
-            dict_dot(d, k, default=dict)[k_last] = val
+            set_item(dict_dot(d, k, default=dict), k_last, val)
         except ValueError:
-            d[k] = val
+            set_item(d, k, val)
         return val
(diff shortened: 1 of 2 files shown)


dblenkus:
Yes, looks like I forgot to update references in input fields.

But, do you really need to complicate that much? As references are just strings in JSON field, it can be done in two passes. In first pass, which is already done, Data objects are just copied and ``id_mapping`` dict is created. So, I will just add the second pass and update all references. Or do you see any problem with this approach?

kostko:
The problem is that this happens during data import, when calling save. There are two options: either defer render_descriptor on save and do it as you suggested (you need to save to get the primary key value) or use a topological sort.
So, to do it as you suggested, you need to add an option to defer render_descriptor on save and then save all items first to get ids, update all references and call render_descriptor afterwards.
On 5 Mar 2016 11:56 pm, Domen Blenkuš <notifications@github.com> wrote:Yes, looks like I forgot to update references in input fields.

But, do you really need to complicate that much? As references are just strings in JSON field, it can be done in two passes. In first pass, which is already done, Data objects are just copied and id_mapping dict is created. So, I will just add the second pass and update all references. Or do you see any problem with this approach?

kostko:
I have implemented data reference migration, but there is another problem. It seems that there are inconsistent references in the old system as there exist references to Data objects, which no longer exist?

The question is what should be done with those? Currently, I replace them with `None`, which are then ignored when rendering the descriptor, but I don't think this is the best option.

mstajdohar:
`None` is OK in the DB. I do not think we can do more on this layer. The UI should probably be more informative, telling the user that some inputs were deleted.

dblenkus:
You can add ``if not field_schema['type'].startswith('data:') and not field_schema['type'].startswith('list:data:'):`` before calling ``render_descriptor``. Input fields of type ``data`` don't have default values so it is safe to do.

dblenkus:
The problem probably occurs in ``iterate_schema``, so this won't fix it :)

dblenkus:
Idea of ``render_descriptor`` is to generate some fields based on values on input fields (i.e. generate name "Alignment of Read xyz" where "Read xyz" is name of input read). As all this was already done on old platform we don't need to regenerate it. We can add bypass like ``if not self.descriptor: generate_descriptor(self)``

kostko:
> As all this was already done on old platform we don't need to regenerate it.

True, but I think that this should be regenerated anyway for consistency - it might happen that things will be slightly different on the new platform and it is good that the import process uses the latest definition.

Can you review the PR? The import process works for me after these fixes.

resolwe/flow/models.py
```diff
@@ -361,8 +361,10 @@ class Meta(BaseModel.Meta):
     descriptor = JSONField(default={})
 
     def save(self, *args, **kwargs):
+        force_render_descriptor = kwargs.pop('force_render_descriptor', False)
+        defer_render_descriptor = kwargs.pop('defer_render_descriptor', False)
 
-        if not self.pk:  # create
+        if force_render_descriptor or (not defer_render_descriptor and not self.pk):  # create
```
dblenkus: Is it on purpose to (not) set default values depending on ``*_render_descriptor`` values?

resolwe/flow/models.py
```diff
@@ -631,22 +637,40 @@ def dict_dot(d, k, val=None, default=None):
...
+        else:
+            dict_or_model[key] = value
+
     if val is None and callable(default):
         # Get value, default for missing
-        # Ugly, but works for model.Data objects as well as dicts
-        # Does the same as:
-        # return reduce(lambda a, b: a.setdefault(b, default()), k.split('.'), d)
-        return reduce(lambda a, b: a.__setitem__(b, a[b] if b in a else default()) or a[b], k.split('.'), d)
+        return reduce(lambda a, b: set_default(a, b, default()), k.split('.'), d)
```
dblenkus: +1 for updating ``dict_dot`` :)

dblenkus:
Do you have any example what can change in descriptor? It is just rendering a template with meta data of input objects, so it should (I think) all stay the same.

I agree to regenerate descriptors if it is trivial, but first major problem is that some inputs have been deleted. The second problem is that you want to set default (generated) values only to fields that hasn't been changed by user.

resolwe/flow/models.py
kostko: Oh, so you mean the second code block besides `render_descriptor`? I am not sure, so in your commit, you added this whole section, so I thought that this was somehow related and put a block around everything. If this is not the case, then the block below should be excluded.

dblenkus: Sorry if I was not clear enough. I meant to add extra if around just one line.

kostko:
> Do you have any example what can change in descriptor? It is just rendering a template with meta data of input objects, so it should (I think) all stay the same.

Ok, then we can just not render the descriptor in case it is already set. I have pushed an updated version implementing this change.

PR #42, opened 2016-03-07
WIP: Support sample

dblenkus:


mstajdohar:
LGTM

resolwe/flow/models.py
```diff
@@ -105,6 +105,13 @@ class Meta(BaseModel.Meta):
 
     data = models.ManyToManyField('Data')
 
+    #: collection descriptor schema
+    descriptor_schema = models.ForeignKey('DescriptorSchema', blank=True, null=True, on_delete=models.PROTECT)
```
mstajdohar: We probably want to clear the `descriptor` field if `descriptor_schema` is deleted.

resolwe/flow/static/flow/processorSchema.json
```diff
@@ -26,6 +26,10 @@
             "description": "Processor category",
             "type": "string"
         },
+        "flow_collection": {
+            "description": "Collection to which object belong",
```
mstajdohar: Maybe change to: Auto-create a collection for the corresponding dataflow

resolwe/flow/models.py
dblenkus: I copied this line from Data model. So do I have to fix it there to?

mstajdohar: Yes. Please check all models.PROTECT references and think if such fix is appropriate.

dblenkus: At the moment we can not delete `Descriptor` if it is referenced in any Data object, so we have consistent descriptors and their schemes. If we change `models.PROTECT`, that means that we will have to handle descriptors without schemes. Do we really want this?

kostko: Why would you want this?

mstajdohar: @dblenkus you are right about this. Let's not allow the deletion of DescriptorSchema if referenced in Data objects.

resolwe/flow/models.py
```diff
@@ -373,6 +393,47 @@ def save(self, *args, **kwargs):
...
+
+                collection_query = Collection.objects.filter(
+                    descriptor_schema__slug=self.process.flow_collection, data__id__in=input_objects).distinct()
+
+                if collection_query.count() == 1:
+                    collection = collection_query.first()
+                else:
+                    des_schema = DescriptorSchema.objects.get(slug=self.process.flow_collection)
+
+                    # XXX: generate meaningful name
```
dblenkus: @mstajdohar what is the best way to generate sample names?

mstajdohar: If not pushed explicitly, names should be set by the process. The bioinformatician knows what is meaningful, and it will be different for each process.

mstajdohar:
LGTM

PR #45, opened 2016-03-10
Add ability to filter lines when comparing test output files

tjanez:
Extend `ProcessorTestCase.assertFiles()` with a filter parameter that is a function for filtering the contents of output files.
Add two test cases for this new feature.

### DIFF
--- a/resolwe/flow/tests/test_test_framework.py
+++ b/resolwe/flow/tests/test_test_framework.py
@@ -1,0 +1,51 @@
+# pylint: disable=missing-docstring
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import io
+import mock
+import unittest
+
+try:
+    import builtins  # py3
+except ImportError:
+    import __builtin__ as builtins  # py2
+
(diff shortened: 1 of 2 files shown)


resolwe/flow/tests/test_test_framework.py
```diff
@@ -0,0 +1,54 @@
...
+            output2_file = io.BytesIO(b"some line\ndate: 2015-10-31\n")
+            open_mock = mock.MagicMock(side_effect=[output1_file, output2_file])
+            with mock.patch.object(builtins, 'open', open_mock):
+                dummy_case = ProcessTestCase.__new__(ProcessTestCase)
+                dummy_case.files_path = ""
+                dummy_case._debug_info = lambda _: ""
+                dummy_case.assertEqual = self.assertEqual
+                obj_mock = mock.MagicMock()
+                def date_in_line(x):
+                    if x.startswith(b"date"):
```
dblenkus: Write this as:```return x.startswith(b"date")```

PR #48, opened 2016-04-01
WIP: Fix error in data create

mstajdohar:
To be merged with @dblenkus's branch.

### DIFF
--- a/resolwe/flow/models.py
+++ b/resolwe/flow/models.py
@@ -420,19 +420,17 @@ class Data(BaseModel):
                     collection = collection_query.first()
                 else:
                     des_schema = DescriptorSchema.objects.get(slug=self.process.flow_collection)
-
-                    col_name = "{}_{}".format(self.descriptor['name'], self.flow_collection)
-                    unique_col_name = col_name
+                    unique_slug = self.name
                     i = 1
-                    while Collection.objects.filter(name=unique_col_name).exists():
-                        unique_col_name = "{}_{}".format(col_name, i)
+                    while Collection.objects.filter(slug=slugify(unique_slug)).exists():
+                        unique_slug = "{} {}".format(self.name, i)
                         i += 1
 
                     collection = Collection.objects.create(
                         contributor=self.contributor,
                         descriptor_schema=des_schema,
-                        name=unique_col_name,
-                        slug=slugify(unique_col_name),
+                        name=self.name,
+                        slug=slugify(unique_slug),
                     )
 
                 collection.data.add(self)
--- a/resolwe/flow/views.py
+++ b/resolwe/flow/views.py
@@ -135,23 +135,35 @@ class ResolweCreateDataModelMixin(ResolweCreateModelMixin):
     """
     def create(self, request, *args, **kwargs):
         collections = request.data.get('collections', [])
-        if len(collections) != 1:
-            return Response({'collections': 'Exactly one id required on create.'},
-                            status=status.HTTP_400_BAD_REQUEST)
+        
+        for collection_id in collections:
+            try:
+                collection = Collection.objects.get(pk=collection_id)
+            except Collection.DoesNotExist:
+                return Response({'collections': ['Invalid pk "{}" - object does not exist.'.format(collection_id)]},
+                                status=status.HTTP_400_BAD_REQUEST)
 
-        try:
-            collection = Collection.objects.get(pk=collections[0])
-        except Collection.DoesNotExist:
-            return Response({'collections': ['Invalid pk "{}" - object does not exist.'.format(collections[0])]},
-                            status=status.HTTP_400_BAD_REQUEST)
-
-        if not request.user.has_perm('add_collection', obj=collection):
-            if request.user.is_authenticated():
-                raise exceptions.PermissionDenied
-            else:
-                raise exceptions.NotFound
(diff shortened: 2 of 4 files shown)


resolwe/flow/models.py
```diff
@@ -420,19 +420,17 @@ def save(self, *args, **kwargs):
...
+                        unique_slug = "{} {}".format(self.name, i)
                         i += 1
 
                     collection = Collection.objects.create(
                         contributor=self.contributor,
                         descriptor_schema=des_schema,
-                        name=unique_col_name,
-                        slug=slugify(unique_col_name),
+                        name=self.name,
+                        slug=slugify(unique_slug),
```
mstajdohar: @dblenkus maybe use [django_unique_slugify](https://github.com/defcube/django-unique-slugify) or similar.

mstajdohar: If name not given, the template should be defined in the process object. We should add a field `name_data_default` to processes. Does anyone (@jkokosar, @kostko, @dblenkus) have a better suggestion for the field name?

resolwe/flow/views.py
```diff
@@ -135,23 +135,35 @@ class ResolweCreateDataModelMixin(ResolweCreateModelMixin):
...
-            return Response({'collections': ['Invalid pk "{}" - object does not exist.'.format(collections[0])]},
-                            status=status.HTTP_400_BAD_REQUEST)
-
-        if not request.user.has_perm('add_collection', obj=collection):
-            if request.user.is_authenticated():
-                raise exceptions.PermissionDenied
-            else:
-                raise exceptions.NotFound
+        
+        for collection_id in collections:
```
mstajdohar: Collections are allowed in REST API, but the data object is not added to them. @dblenkus?

mstajdohar: I have implemented the REST API so that users do not have to list collections, but they can if they want to.

mstajdohar: Probably not the most efficient implementation.

dblenkus: I've already fixed this on my branch. The problem here is how to construct response if user has permissions on all collections except one (or maybe one does not exist).
Quick solution is to disable setting any collections on create and then make separate requests.

dblenkus: One more question is how to make requests for adding data objects to collections. Do you have to post list of all collections? In this case you can (accidentally override updated list on server). The more secure way would be to create special endpoints, ie. `collection/<id>/add_data` and `collection/<id>/remove_data`.
What do you think @mstajdohar?

mstajdohar: I vote to fail the whole transaction.

resolwe/flow/models.py
```diff
@@ -420,19 +420,17 @@ def save(self, *args, **kwargs):
...
                 else:
                     des_schema = DescriptorSchema.objects.get(slug=self.process.flow_collection)
-
-                    col_name = "{}_{}".format(self.descriptor['name'], self.flow_collection)
-                    unique_col_name = col_name
+                    unique_slug = self.name
                     i = 1
-                    while Collection.objects.filter(name=unique_col_name).exists():
-                        unique_col_name = "{}_{}".format(col_name, i)
+                    while Collection.objects.filter(slug=slugify(unique_slug)).exists():
```
dblenkus: +1

resolwe/flow/models.py
dblenkus: Thanks, I will check it.

resolwe/flow/views.py
mstajdohar: I do not expect this feature (creating new data on multiple collections simultaneously) will be used much.

dblenkus: I agree that failing whole transaction is easiest solution. But do we want to return just first error or all of them?
We have to thing what to support and what not. If something won't be used much and is to complicated to think through all possibilities then I don't see a reason not to disable it.

mstajdohar:
@dblenkus, maybe you should merge this with your branch and do a unified PR.

resolwe/flow/views.py
mstajdohar: The web devs (@kostko, @hadalin) should tell how they would like to append data objects to collections at create and what they would want in return. Adding a data object to a one collection is definitely used in GenBoard. The zero collections alternative is used by @JureZmrzlikar in resolwe-bio-py. I do not expect a case with multiple collections in the near future. But I do not like exceptions and it is quite ugly to me to support just one element in an array.

resolwe/flow/models.py
kostko: How was the name generated on the old platform?

dblenkus: Template was defined in static schema (now descriptor schema) and generated in `render_static` function. We also have `render_descriptor` function in resolwe, so we probably just have to copy name from `descriptor.name` to `name`.

dblenkus: This is for the name of `Data` object. But here we generate name of `sample`, which is a concept that wasn't supported on old platform.

tjanez:
@mstajdohar, @dblenkus: Please, rebase this or the other branch on top of `master` branch and add a CHANGELOG entry.

mstajdohar:
I think @dblenkus will change this completely with his updated code.

PR #50, opened 2016-04-11
Fix datacreate

dblenkus:


resolwe/permissions/tests/test_data.py
```diff
@@ -62,9 +62,9 @@ def test_post(self):
         d = Data.objects.get(pk=resp.data['id'])
         self.assertTrue(datetime.now() - d.modified < timedelta(seconds=1))
         self.assertTrue(datetime.now() - d.created < timedelta(seconds=1))
-        self.assertEqual(d.status, 'RE')
-        self.assertEqual(d.started, None)
-        self.assertEqual(d.finished, None)
+        self.assertEqual(d.status, 'OK')
+        self.assertTrue(datetime.now() - d.started < timedelta(seconds=1))
```
mstajdohar: We should stop using just `datetime.now`. Read about [Django's Time Zones](https://docs.djangoproject.com/en/1.9/topics/i18n/timezones/). We should make resolwe invariant to Django settings and when time zone support not enabled use:
```
import datetime
now = datetime.datetime.now()
```
but when time zone support is enabled (USE_TZ=True) use:
```
from django.utils import timezone
now = timezone.now()
```

mstajdohar: Sorry, not relevant here, because this is a test with specific time zone setting. Will delete.

mstajdohar:
Maybe also fix the `datetime.utcnow()` in [executors](https://github.com/genialis/resolwe/blob/master/resolwe/flow/executors/__init__.py).

We should stop using simply `datetime.now`/`datetime.utcnow`. Read about [Django's Time Zones](https://docs.djangoproject.com/en/1.9/topics/i18n/timezones/). We should make resolwe invariant to Django settings and when time zone support not enabled use:
```
import datetime
now = datetime.datetime.now()
```
but when time zone support is enabled (USE_TZ=True) use:
```
from django.utils import timezone
now = timezone.now()
```

mstajdohar:
We should use [django-autoslug](https://github.com/neithere/django-autoslug) instead of ours `def unique_slug(cls, slug):` or we will have problems with slug field length.

resolwe/flow/models.py
```diff
@@ -420,21 +422,22 @@ def save(self, *args, **kwargs):
...
                         contributor=self.contributor,
                         descriptor_schema=des_schema,
-                        name=unique_col_name,
-                        slug=slugify(unique_col_name),
+                        name=self.name,
+                        slug=slugify(unique_slug),
                     )
 
+                    for permission in list(zip(*collection._meta.permissions))[0]:
+                        shortcuts.assign_perm(permission, collection.contributor, collection)
```
mstajdohar: What if this fails? Can it?

resolwe/flow/views.py
```diff
@@ -135,23 +136,51 @@ class ResolweCreateDataModelMixin(ResolweCreateModelMixin):
...
+                                status=status.HTTP_400_BAD_REQUEST)
+
+            if not request.user.has_perm('add_collection', obj=collection):
+                if request.user.is_authenticated():
+                    raise exceptions.PermissionDenied
+                else:
+                    raise exceptions.NotFound
+
+        process_slug = request.data.get('process', None)
+        process_quey = Process.objects.filter(slug=process_slug).order_by('version')
```
mstajdohar: Typo, `process_query`.

resolwe/flow/views.py
```diff
@@ -135,23 +136,51 @@ class ResolweCreateDataModelMixin(ResolweCreateModelMixin):
...
-        if not request.user.has_perm('add_collection', obj=collection):
+        if not request.user.has_perm('view_process', obj=process):
             if request.user.is_authenticated():
                 raise exceptions.PermissionDenied
             else:
                 raise exceptions.NotFound
 
-        return super(ResolweCreateDataModelMixin, self).create(request, *args, **kwargs)
+        resp = super(ResolweCreateDataModelMixin, self).create(request, *args, **kwargs)
+        manager.communicate()
```
mstajdohar: Do we need try/catch here?

mstajdohar:
Fix tests.

mstajdohar:
Supported the timezones in executor in #51.

resolwe/flow/models.py
dblenkus: I think that it cannot fail, as non of the inputs is controlled by user.

resolwe/flow/views.py
dblenkus: Fixed.

resolwe/flow/views.py
dblenkus: At the moment you get `500 error` (without traceback if `DEBUG` is `False`). I think that `500` is appropriate error, but we can add helpful message. I'll add task for this and do it later, because it is not crucial for the moment.

tjanez:
Next time, please add a CHANGELOG entry for all note-worthy changes.

PR #55, opened 2016-04-12
Add endpoints for adding/removing Data to/from Collection

dblenkus:


mstajdohar:
@kostko, @dblenkus says it's not the most elegant code, but it works. Could you make some improvements?

dblenkus:
The problem is that `descriptor_schema` is serialized with our custom serializer and I cannot find nice way to reference already existing `descriptor_schema` when creating `Data`/`Collection`.

kostko:
Can you comment on the part of the diff that you find problematic?

resolwe/flow/views.py
```diff
@@ -182,6 +204,12 @@ def perform_create(self, serializer):
             collection = Collection.objects.get(pk=c)
             collection.data.add(instance)
 
+        ds_slug = self.request.data.get('descriptor_schema', None)
+        if ds_slug:
+            descriptor_schema = DescriptorSchema.objects.filter(slug=ds_slug).order_by('version').last()
+            instance.descriptor_schema = descriptor_schema
+            instance.save()
```
dblenkus: `DescriptorSchema` is added in `perform_create`. It would be probably better if `descriptor_schema` is passed to `DataSerializer` and it handles it on initial save.

PR #57, opened 2016-04-13
Fix second analysis step in flow collection

mstajdohar:
Fixes the problem with Bowtie 2 in the primary analysis after annotation.
Create collection without slug fixes the New Collection error in the Annotation page.


mstajdohar:
Do not accept yet. Breaks MACS14.

PR #58, opened 2016-04-13
Make Celery engine work

tjanez:


mstajdohar:
Would be great if you added the celery requirement and settings for the test project so developers could use it.

dblenkus:
It is a bit complicated to use it in development, will need to add few extra features.

PR #64, opened 2016-04-15
Don't set error status if rc is 0

dblenkus:


mstajdohar:
Was that the premature ERROR bug?

dblenkus:
I think so. Genome process is using `re-save proc.rc` a lot.

PR #66, opened 2016-04-19
Render data name after resolved

mstajdohar:
I also fixed all the other tests that failed or did not clean after themselves.


resolwe/flow/engines/__init__.py
```diff
@@ -88,7 +88,7 @@ def communicate(self, run_sync=False, verbosity=1):
                         continue
 
                     data.status = Data.STATUS_WAITING
-                    data.save()
+                    data.save(render_name=True)
```
mstajdohar: Ugly

resolwe/flow/models.py
```diff
@@ -384,26 +391,40 @@ class Meta(BaseModel.Meta):
...
 
-            # default values for INPUT
+            # Default values for INPUT
             for field_schema, fields, path in iterate_schema(self.input, self.process.input_schema, ''):
                 if 'default' in field_schema and field_schema['name'] not in fields:
                     dict_dot(self.input, path, field_schema['default'])
 
+        elif kwargs.get('render_name', False):
+            render_name(self)
+            del kwargs['render_name']
```
mstajdohar: Ugly

mstajdohar:
I did some ugly hacks to implement the desired functionality. I would be happy if someone improved them.

resolwe/permissions/tests/test_data.py
```diff
@@ -1,8 +1,18 @@
 # pylint: disable=missing-docstring
 from __future__ import absolute_import, division, print_function, unicode_literals
```
mstajdohar: Wanted to make resolwe tests compatible with `USE_TZ = True`.

resolwe/flow/models.py
dblenkus: Define `save` function as `save(self, render_name=False, *args, **kwargs):` and it will be a lot nicer.

Django [`save` function](https://docs.djangoproject.com/en/1.9/_modules/django/db/models/base/#Model.save) already has few such arguments

dblenkus: Other option would be to define separate `Data.render_name` function and call it before calling `save`.

tjanez:
@mstajdohar, could you please split unrelated changes into separate commits?

tjanez:
@mstajdohar, I added `six`  depency as part of #69.

resolwe/flow/tests/test_executors.py
```diff
@@ -27,7 +27,8 @@ class DockerExecutorTestCase(unittest.TestCase):
     @mock.patch('os.chmod')
     @mock.patch('os.chdir')
     @mock.patch('resolwe.flow.executors.Data.objects.filter')
-    def test_run_in_docker(self, data_filter_mock, chdir_mock, chmod_mock, mkdir_mock):
+    @mock.patch('resolwe.flow.executors.Data.objects.get')
```
tjanez: Was this mock object accidentally added since it is not used inside the test function?

mstajdohar: It is used in the function called by this function. It crashes without this mock.

tjanez: Aha, ok. Thanks for explaining!

mstajdohar:
Thank you for comments. They were addressed accordingly.

PR #67, opened 2016-04-19
Add ability to use a custom executor command

tjanez:
Use it by specifying the `FLOW_EXECUTOR['COMMAND']` setting.
Change test project's settings to allow setting a custom executor command via `RESOLWE_EXECUTOR_COMMAND` environment variable.

### DIFF
--- a/resolwe/flow/executors/local.py
+++ b/resolwe/flow/executors/local.py
@@ -3,8 +3,11 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import logging
 import os
+import shlex
 import subprocess
 import time
+
+from django.conf import settings
 
 from resolwe.flow.executors import BaseFlowExecutor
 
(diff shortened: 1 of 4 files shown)


resolwe/flow/executors/local.py
```diff
@@ -21,7 +21,8 @@ def __init__(self):
         self.kill_delay = 5
 
     def start(self):
-        self.proc = subprocess.Popen(['/bin/bash'],
+        command = settings.FLOW_EXECUTOR.get('COMMAND', '/bin/bash')
+        self.proc = subprocess.Popen([command],
```
dblenkus: `[command]` -> `shlex.split(command)` would be more general.

tjanez: Thanks! I'll fix that.

dblenkus:
LGTM

PR #69, opened 2016-04-20
Make Resolwe work with Python 3 again

tjanez:
Add explicit requirement on six package to setup.py


tjanez:
@mstajdohar, @dblenkus, if we accept this PR before the others, we can easily make a bugfix release 1.1.1 which restores Python 3 support.

If we plan to relase 1.2.0 before the end of the week, then it doesn't matter so much.

dblenkus:
LGTM

PR #71, opened 2016-04-26
WIP: Systematically support API filters

dblenkus:


mstajdohar:
Rebase to [PR-75](https://github.com/genialis/resolwe/pull/75), make sure tests pass.

mstajdohar:
Maybe this should now be done with Haystack, see [Query Field Lookups in DRF Haystack docs](https://drf-haystack.readthedocs.io/en/latest/advanced_usage.html#query-field-lookups) and [Field Lookups in Haystack docs](https://django-haystack.readthedocs.io/en/latest/searchqueryset_api.html?highlight=lookups#field-lookups).

hadalin:
It comes down to performance because the API is similar and I'm not sure if there's any performance gain for this. We would have to test.

PR #73, opened 2016-05-04
Improve Tox configuration

tjanez:


tox.ini
```diff
@@ -4,8 +4,11 @@ skip_missing_interpreters = True
 
 [testenv]
 install_command = pip install --process-dependency-links {opts} {packages}
-deps = .[docs,test]
 commands =
+# install documentation and testing requirements
+# NOTE: Don't use 'deps = .[docs,test]' tox option since we want Tox to install
+# the package from sdist first
+    pip install --process-dependency-links .[docs,test]
```
dblenkus: Is `--process-dependency-links` really necessary?

tox.ini
```diff
@@ -4,8 +4,11 @@ skip_missing_interpreters = True
 
 [testenv]
 install_command = pip install --process-dependency-links {opts} {packages}
```
dblenkus: You can probably also remove this and use default value (`pip install {opts} {packages}`).

tox.ini
tjanez: Currently, no. However, using `--process-dependency-links` in both places in `tox.ini` ensures that a developer can easily add a git version of a dependency without knowing all the details of Tox.

dblenkus: I think that `master` branch should depend only on officially released packages and tests should reflect that. Or is there an use-case where we want to reference git dependency?

tjanez: I fully agree that the `master` branch should only depend on officially released packages. What I was thinking was more a convenience for developers that would sometimes need to bump a certain dependency to the latest git version while developing on a personal feature branch.

Anyway, I think we can omit `--process-dependency-links` here and I'll update the PR.

PR #77, opened 2016-05-10
Set default values before anything else

dblenkus:


resolwe/flow/models.py
```diff
@@ -349,27 +349,27 @@ def __init__(self, *args, **kwargs):
 
     def save(self, render_name=False, *args, **kwargs):
         # Generate the descriptor if one is not already set.
```
mstajdohar: Remove or move the comment.

mstajdohar:
Minor revision, otherwise LGTM.

PR #78, opened 2016-05-10
Enable spawned processes in executors

dblenkus:


dblenkus:
@jkokosar please test if this works.

mstajdohar:
LGTM

PR #81, opened 2016-05-12
Ensure a recent pip is installed when Tox creates a new virtualenv

tjanez:
This fixes a problem with installing Resolwe in a new virtualenv on CentOS 7.


tjanez:
It appears this is not necessary.

PR #82, opened 2016-05-13
Support creating Process on API

dblenkus:


mstajdohar:
LGTM

resolwe/permissions/tests/test_tool.py
```diff
@@ -35,8 +43,8 @@ def test_get_list_admin(self):
         self.assertEqual(len(resp.data), 2)
 
     def test_post(self):
-        resp = self._post({'name': 'New process'}, self.admin)
-        self.assertEqual(resp.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)
+        resp = self._post(self.post_data, self.admin)
+        self.assertEqual(resp.status_code, status.HTTP_201_CREATED)
 
     def test_patch(self):
```
mstajdohar: We need to allow patch and test for HTTP_200_OK here. Development would be a real pain otherwise.

resolwe/permissions/tests/test_tool.py
```diff
@@ -35,12 +43,12 @@ def test_get_list_admin(self):
...
         resp = self._patch(1, {'name': 'Hacked process'}, self.admin)
-        self.assertEqual(resp.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)
+        self.assertEqual(resp.status_code, status.HTTP_200_OK)
```
mstajdohar: PATCH works in test but not Python SDK yet.

mstajdohar: @dblenkus, any ideas why?

PR #86, opened 2016-05-19
Docker executor: add ability to set a dynamic uid and gid

tjanez:


resolwe/flow/executors/docker.py
```diff
@@ -18,14 +18,18 @@ def __init__(self, *args, **kwargs):
...
+        }
 
         if self.data_id != 'no_data_id':
-            container_name = 'resolwe_{}'.format(self.data_id)
+            data_id = self.data_id
         else:
             # set random container name for tests
-            rand_int = random.randint(1000, 9999)
-            container_name = 'resolwe_test_{}'.format(rand_int)
+            data_id = 'test_{}'.format(random.randint(1000, 9999))
```
mstajdohar: Why not have a global counter that increases for each test data?

mstajdohar:
Very nice work. One minor comment, otherwise LGTM.

resolwe/flow/tests/__init__.py
```diff
@@ -59,23 +59,6 @@ def _register_schemas():
         SCHEMAS_FIXTURE_CACHE['descriptor_schemas'] = list(DescriptorSchema.objects.all())
 
 
-flow_executor_settings = settings.FLOW_EXECUTOR.copy()
-# since we don't know what uid/gid will be used inside Docker executor, others
-# must have all permissions on the data directory
-flow_executor_settings['DATA_DIR_MODE'] = 0o777
-# create a temporary upload directory so we can give it relaxed permissions and
-# make it convenient for use by Docker
-flow_executor_settings['UPLOAD_PATH'] = os.path.join(flow_executor_settings['UPLOAD_PATH'], 'test_upload')
```
dblenkus: We should probably keep this.

mstajdohar: I like that test upload files go to a separate directory. I would do the sample thing for test data folders. This is just my opinion, you can decide how you want.

resolwe/flow/executors/docker.py
tjanez: Yea, we could improve this. But since this is not a regression from what we have, I suggest to fix this later.

mstajdohar: OK

resolwe/flow/tests/__init__.py
tjanez: I've kept the separate test upload directory.

PR #87, opened 2016-05-24
Fix tests

dblenkus:


resolwe/flow/tests/__init__.py
```diff
@@ -199,23 +199,22 @@ def run_process(self, process_slug, input_={}, assert_status=Data.STATUS_DONE, r
...
-
-                        # since we don't know what uid/gid will be used inside Docker executor,
-                        # we must give others read and write permissions
-                        os.chmod(new_path, 0o666)
-                        fields[field_schema['name']] = {
-                            'file': file_name,
-                            'file_temp': file_name,
-                        }
-                        break
+                file_name = os.path.basename(fields[field_schema['name']])
```
mstajdohar: Does this work is you call tests in genialis-base? The current implementation allowed to call resolwe and resolwe_bio tests from the genialis-base project.
```
./manage.py test resolwe
```
I think you people using this Django apps has to be able to run tests on them in any Django package.

dblenkus: Yes, it will work in genialis-base. Only requirement is to set ``self.files_path``in each test class. But this is already required by ``assertFile`` and ``assertJSON``functions, so this change doesn't add any extra complexity to the testing procedure.

Here is example from resolwe-bio: https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/tests/utils.py#L31

mstajdohar:
LGTM

PR #92, opened 2016-06-01
Improve permissions system

dblenkus:


dblenkus:
Jenkins, test this please

resolwe/permissions/shortcuts.py
```diff
@@ -23,48 +61,138 @@ def get_user_group_perms(user_or_group, obj):
...
+                user_filters['{}__content_object'.format(related_name)] = obj
+
+            user_perms_qs = perms_qs.filter(**user_filters)
+            user_perms = list(chain(user_perms_qs.values_list("codename", flat=True)))
+
+    group_perms_qs = Group.objects.filter(**group_filters)
+    group_perms = list(chain(group_perms_qs.values_list(
+        "pk", "name", "{}__permission__codename".format(group_rel_name))))
+
+    # TODO: Optimize by writing custom Django's aggregate expresion
```
kostko: What exactly should be optimized here? You just need to ensure that all the returned items are ordered by `(pk, name)` and then you can group in Python in a single pass, which is optimal. I am not sure that generating an array in the database would be much difference unless there are really a lot of rows returned (in which case the overhead of repeating pk and name would be noticeable).

kostko:
LGTM

resolwe/permissions/shortcuts.py
dblenkus: There can be quite a few groups with permissions on some frequently used objects (i.e. `Process` or `DescriptorSchema`). But I agree that this is not a big problem if results are ordered properly. So I will add `.order_by("pk")` (`name` can be omitted as it is unique for `pk`) and remove the comment.

PR #94, opened 2016-06-05
Validate JSON fields

dblenkus:
@miha: This is first version of validation. Any comments are appreciated. Some concerns are marked with `XXX`.


resolwe/flow/models.py
```diff
@@ -372,6 +374,15 @@ def save(self, render_name=False, *args, **kwargs):
         if not self.descriptor:
             render_descriptor(self)
 
+        if create:
+            validate_schema(self.input, self.process.input_schema)
+        if self.descriptor_schema:
+            validate_schema(self.descriptor, self.descriptor_schema.schema)
+        elif self.descriptor and self.descriptor != {} and self.descriptor != "{}":
+            raise ValueError("`descriptor_schema` must be defined if `descriptor` is given")
+        if self.status == Data.STATUS_DONE:
```
mstajdohar: We should probably always check this. The process is saving output fields even before the Data is DONE. We do not want conflicting values in DB.

resolwe/flow/models.py
```diff
@@ -583,6 +594,81 @@ def validation_schema(name):
...
+TYPE_SCHEMA = validation_schema('type')
+
+
+def validate_schema(instance, schema):
+    """Check if DictField values are consistent with our data types.
+
+    Perform basic JSON schema validation and our custom validations:
+
+      * check that required fields are given
+      * check if ``basic:file:`` and ``list:basic:file`` fields matches
```
mstajdohar: "matches" -> "match"

resolwe/flow/models.py
```diff
@@ -583,6 +594,81 @@ def validation_schema(name):
...
+        regex given in schema (only if ``validate_regex`` is defined in
+        schema for coresponding fields)
+      * check that referenced ``Data`` objects (in ``data:<data_type>``
+        and  ``list:data:<data_type>`` fields) exists and are of type
+        ``<data_type>``
+      * check that referenced ``Storage`` objects (in ``basic:json``
+        fields) exists
+
+    :param list instance: Instance to be validated
```
mstajdohar: Does this work? We used to have:
```
:param instance: ...
:type instance: list
```

resolwe/flow/models.py
```diff
@@ -583,6 +594,81 @@ def validation_schema(name):
...
+            if type_ == 'list:basic:file:' and 'validate_regex' in _schema:
+                for file_ in field:
+                    validate_file(file_['file'], _schema['validate_regex'])
+
+            if type_ == 'basic:json:' and not Storage.objects.filter(pk=field).exists():
+                raise ValidationError(
+                    "Referenced `Storage` object does not exist (id:{})".format(field))
+
+            if type_.startswith('data:'):
+                validate_data(field, type_[5:])  # remove "data:" from `type_`
```
mstajdohar: We should probably allow Data objects that do not exist, but they should be of the right type if they are present. Say you run a differential expression, 10 cases vs 10 controls. Later you remove some of the input data. The differential expression result is still valid. You can not access the original objects any more, but you probably do not have to. If you would remove the Data object from the list (to make it valid), you could get a false impression that the differential expression was only done on the subset of data. This should be handled correctly in UI: we should write "(deleted)" next to the link to the data objects that do not exist. And they should not be links, only labels.

mstajdohar:
```
type_[5:]
```

You could trim this "smartly" in the `validate_data ` function. It is easy to detect if it starts with list or not. Any it would make the calls more elegant.

resolwe/flow/models.py
```diff
@@ -583,6 +594,81 @@ def validation_schema(name):
...
+        fields) exists
+
+    :param list instance: Instance to be validated
+    :param list schema: Schema for validation
+    :rtype: None
+    :raises ValidationError: if ``instance`` doesn't match schema
+      defined in ``schema``
+
+    """
+    def validate_file(filename, reg):
```
mstajdohar: Don't forget about file size.

resolwe/flow/models.py
dblenkus: This is exactly what this test is for: https://github.com/genialis/resolwe/pull/94/files/d223e543ff0af5405ff3ba085af6ae060ecdaa47#diff-1a3a69110496bbde71fce111ce3c6848R140
Input validation is performed only on create, so all inputs must exists when you create Data object, but you can delete them later.

resolwe/flow/models.py
dblenkus: Yes, it works for simple types. Here are some examples: http://www.sphinx-doc.org/en/stable/domains.html#info-field-lists

resolwe/flow/tests/test_validation.py
```diff
@@ -0,0 +1,375 @@
...
+        instance = {'date': '2000-12-31'}
+        validate_schema(instance, schema)
+
+        instance = {'date': '2000/01/01'}
+        validate_schema(instance, schema)
+
+        instance = {'date': '2000 04 31'}
+        validate_schema(instance, schema)
+
+        instance = {'date': '2000.06.21'}  # XXX: Is this ok?
```
mstajdohar: I would support / and - only.

mstajdohar: I maybe even just [ISO](https://en.wikipedia.org/wiki/ISO_8601) data format standard.

mstajdohar: Consult @lukaw3d how does JavaScript's Date object bind to UI and how does it serialize in HTTP request.

lukaw3d: `Sequencing date` is currently POSTed as "2016-06-16".
The date `created` is currently received from server as "2016-05-18T18:13:39.606296Z"

mstajdohar: Note that one is Resolwe field `basic:date`, the other is Django's model field `models.DateTimeField`.

resolwe/flow/models.py
```diff
@@ -583,32 +627,121 @@ def validation_schema(name):
...
+        if id_ in file_name:
+            file_name = file_name[file_name.find(id_) + len(id_):]  # remove id from filename
+
+        return os.path.join(settings.FLOW_EXECUTOR['DATA_DIR'], id_, file_name)
+
+    def hydrate_storage(storage_id):
+        if re.match('^[0-9a-fA-F]{24}$', str(storage_id)) is None:
+            print("ERROR: basic:json value in {} not ObjectId but {}.".format(name, storage_id))
+
+        return Storage.objects.get(pk=storage_id)
```
mstajdohar: Return a lazy object.

resolwe/flow/models.py
```diff
@@ -349,6 +351,35 @@ def __init__(self, *args, **kwargs):
         super(Data, self).__init__(*args, **kwargs)
         self._original_name = self.name
 
+    def save_storage(self, instance, schema):
+        """Save basic:json values to a Storage collection."""
+
+        validator = {val['name']: val for val in schema}
+        for key in instance:
```
mstajdohar: You should use `iterate_fields` to avoid the duplication of the iteration logic.

resolwe/flow/models.py
```diff
@@ -583,32 +627,121 @@ def validation_schema(name):
...
     """
+    def hydrate_file(file_name):
+        id_ = "{}/".format(data.id)  # needs trailing slash
+        if id_ in file_name:
+            file_name = file_name[file_name.find(id_) + len(id_):]  # remove id from filename
+
+        return os.path.join(settings.FLOW_EXECUTOR['DATA_DIR'], id_, file_name)
+
+    def hydrate_storage(storage_id):
+        if re.match('^[0-9a-fA-F]{24}$', str(storage_id)) is None:
```
mstajdohar: Remove this 2 lines.

mstajdohar:
Tests fail because validation works! :) We test Data PATCH with a field that is not in the schema.

resolwe/flow/models.py
```diff
@@ -583,32 +627,121 @@ def validation_schema(name):
...
     for field_schema, fields in iterate_fields(output, output_schema):
         name = field_schema['name']
         value = fields[name]
         if 'type' in field_schema:
             if field_schema['type'].startswith('basic:file:'):
-                fn = value['file']
-                id_ = "{}/".format(data.id)  # needs trailing slash
-                if id_ in fn:
-                    fn = fn[fn.find(id) + len(id_):]  # remove id from filename
+                value['file'] = hydrate_file(value['file'])
```
mstajdohar: Add file size if there is no size key in the dict. Maybe write `hydrate_size()`?

mstajdohar:
We should run a schema validation when saving processes. I was able to save a field that does not exist to the process' `output_schema`.

mstajdohar:
Please fix tests.

mstajdohar:
In addition to `file_size` please support `dir_size`.

mstajdohar:
Check that files and directories saved in `basic:file` and `basic:dir` fields actually exist. This may be supported for files, but not for directories.

mstajdohar:
Test `basic:file`, `list:basic:file`, `basic:dir` and `list:basic:dir` in `input`, `output` and `descriptor` fields. Janez had a validation bug with the `basic:dir` field in `output` because `size` was required. Did you support the computation of directory size?

mstajdohar:
Rebase to master.

PR #96, opened 2016-06-09
Fix permissions url paths

dblenkus:


mstajdohar:
Ugly. Is this workaround needed for both DRF 3.3.2 and DRF 3.3.3?

dblenkus:
Yes, I've tested with DRF 3.3.3, but also doesn't work. Looks like two methods cannot be registered on same url path, although they have different methods. I will write an issue on DRF and try to find better solution. This is (hopefully) just a workaround, so front-end can be finished.

dblenkus:
@mstajdohar fixed

resolwe/flow/views.py
```diff
@@ -219,18 +219,30 @@ class ResolwePermissionsMixin(object):
     """
 
     def _fetch_user(self, query):
-        """Get user by ``pk`` or ``email``. Return ``None`` if doesn't exist."""
+        """Get user by ``pk`` or ``username``. Return ``None`` if doesn't exist."""
+        User = get_user_model()
+
         try:
```
mstajdohar: Too many queries and too complicated. Maybe define a `get_by_id_or_username(value)` method on the User's object manager. If the value is int, query by id, else by the username.

dblenkus:
@mstajdohar I've reduced number of queries. I won't add it to UserManager as it is default Django's and it doesn't make sense to subclass it for such additions. Another problem is that query is a string even if it is id,but I've added nicer check.

mstajdohar:
@dblenkus do you ever sleep?

mstajdohar:
LGTM

PR #100, opened 2016-06-20
Implement data purge

kostko:


kostko:
Test (`test_purge`) passes locally for me, but fails on Jenkins for some reason.

cc @tjanez @dblenkus

kostko:
It seems that Jenkins was configured incorrectly, passing:
```
export RESOLWE_EXECUTOR_COMMAND="sudo docker"
```

instead of using `/bin/bash` as the executor (since tests use the `local` executor). I've fixed the Jenkins configuration and now the test passes.

kostko:
Is there a `re-save-file` equivalent for directories? I could just `echo` some output in tests, but this doesn't seem portable.

dblenkus:
Yes, there is `re-save-dir` in latest `resolwe-runtime-utils`. You have to pull latest docker image to get it. You can check whole list of commands [here](https://github.com/genialis/resolwe-runtime-utils/blob/master/setup.py#L82).

tjanez:
@kostko, good job at finding the cause for the test error. I've also updated the configuration of the [*resolwe-master* test project](https://ci.genialis.com/job/resolwe-master/).

There is still an issue though:
[`resolwe.flow.tests.test_executors.DockerExecutorTestCase.test_run_in_docker`](https://ci.genialis.com/job/resolwe-pull_requests/576/TOXENV=py34,label=master/testReport/resolwe.flow.tests.test_executors/DockerExecutorTestCase/test_run_in_docker/) test should not be skipped. Skipping occurs because the `docker info` command doesn't have required permissions to access the Docker daemon. It must be prefixed by `sudo`, e.g. `sudo docker info` and then it works.

The solution is probably to set `FLOW_EXECUTOR['COMMAND']` to `"sudo docker"` only for the `test_run_in_docker` test?

tjanez:
Thinking a little more about the issue, we probably can't just hard-code `FLOW_EXECUTOR['COMMAND']` to `"sudo docker"` for the `test_run_in_docker` test since different developer/testing environments would need different commands.

I see 2 options:
- We could extend the settings to allow one to define executor-specific `'COMMAND'`.
This would be more general but I wonder if there would ever be such a use case in practice since one would probably just use one executor in production.
- We could put a little "magic" in the `test_run_in_docker` test to go through a list of viable possibilities, e.g. `"docker"`, `"sudo docker"`, ... and only skip the test if none of the hard-coded options would work.

@dblenkus, @kostko, what do you think?

dblenkus:
We have to make sure that any test can run docker (not just `test_run_in_docker`). I.e. purge test is also failing because of docker.
Is there any problem with defining `RESOLWE_EXECUTOR_COMMAND` on per machine base?

kostko:
> Is there any problem with defining RESOLWE_EXECUTOR_COMMAND on per machine base?

So the problem is defining `RESOLWE_EXECUTOR_COMMAND` with the incorrect executor. For example, for `local` it has to be `/bin/bash` (and not `sudo docker` or similar), while for `docker` it has to be a proper way to execute the docker command (with or without sudo).

So the purge test doesn't care if it is run with docker or with the local executor. It works either way. But the executor that is used must be configured correctly.

If I understand correctly, the docker test requires the docker executor? Perhaps the suggestion by @tjanez with having a per-executor command configuration (one for local, another one for docker) would be best.

dblenkus:
Executor settings are overridden in [docker test](https://github.com/genialis/resolwe/blob/master/tests/settings.py). So we can just add appropriate command in there. It is not the nicest way, but is is only used in tests. There is no plan to support more than one simultaneous executor in production.

dblenkus:
Fixed link: https://github.com/genialis/resolwe/blob/master/resolwe/flow/tests/test_executors.py#L56

kostko:
> So we can just add appropriate command in there.

But which one is appropriate? Different systems require different commands.

tjanez:
I'm more in favour of my second proposal:

> We could put a little "magic" in the `test_run_in_docker` test to go through a list of viable possibilities, e.g. `"docker"`, `"sudo docker"`, ... and only skip the test if none of the hard-coded options would work.

There is also a sub-variant of this proposal, a special setting that sets the executor's `COMMAND` just for this test. What do you think?

dblenkus:
Good point :). It is only needed in one test, so we can, as Tadej suggested, iterate through different commands in that test and don't support this otherwise.

kostko:
I've added some more tests for purge. @dblenkus can you review?

dblenkus:
It is really important that purge is working exactly as expected. So write some more tests to make sure that it handles each field type as it should.

It's been discussed with @mstajdohar to move part of the code, where it is determined which files to delete and which not, to a separate function. Then you can write tests without creating data objects (but leave the existing test to test the whole workflow), just pass output schema to the new function and check the result. This way you can easily make minimal examples for each field type.

Also make sure that purge:
- works when `data_id` is passed to it (this is the most common use-case) - make sure that other data folder are not touched
- files are deleted only when processing is done

@mstajdohar is it possible that files are referenced in `descriptor`? Or are they only in `output`?

mstajdohar:
Files will be references in the descriptor as well.

mstajdohar:
Make sure purge is called when a data object becomes DONE (in the executor?) and when the data object is deleted.

tjanez:
Also, don't forget to add a change log entry.

kostko:
> It's been discussed with @mstajdohar to move part of the code, where it is determined which files to delete and which not, to a separate function. Then you can write tests without creating data objects (but leave the existing test to test the whole workflow), just pass manually created output_schema to the new function and check the result. This way you can easily make minimal examples for each field type.

If you expect this function to only take as input the output and its schema, it can only return the reverse: which files not to remove. But to check which files to actually remove, you actually need to access the filesystem to check, which files are there. This means that at least the data path must exist and have some files, unless you are suggesting that we also mock filesystem access and fake which files are there.

> So write some more tests to make sure that it handles each field type as it should.

So which field types are missing in the test? We currently test:
* `basic:file:`
* `basic:dir:`
* `list:basic:file:`
* `list:basic:dir:`

And for each of those we also test if `refs` are handled correctly.

The only one I see as missing is `basic:url`, which I find strange to have there, since it is supposed to represent an URL and not a local file? Or do we support `file://` URLs as well?

> files are deleted only when processing is done

Can you elaborate on this one? When could they be deleted? While the processor is still running?

dblenkus:
Maybe the function can take the output, it's schema and path to the top-level directory. 

I see two options for tests:
* mock the file system
* create some real directories and files (maybe with [`tempfile`](https://docs.python.org/3/library/tempfile.html)) and run all tests against it

I prefer the second option, but you can choose whichever you think is better.

I know that current test covers all options. But do you really know that file is kept because it is listed in `basic:file:` field? Or is there some other coincidence that prevent file from being deleted?
I propose to make separate example for every field type and check its behavior isolated from others.

kostko:
Ok. What about the `basic:url:` type? How can that reference local files?

kostko:
Ok I've added some more detailed tests. For each field we now separately test three things:
* If the `get_purge_objects` returns only specific paths (and nothing else).
* If purge is correctly executed after a test processor is run, with references in its `output`.
* If purge is correctly executed after a test processor is run, with references in its `descriptor`.

BTW, how can there be an unreferenced Storage object, since we require all Storage objects to have foreign keys to respective Data objects and the database enforces referential constraints? Is this a leftover from the previous MongoDB store? I think we can safely remove purging of Storage references.

mstajdohar:
Purge should ignore `basic:url:`. Nice work with the tests.

dblenkus:
Part that handles Storage object is a leftover from MongoDB. I think that you can safely delete it. `output` is not meant to be changed after Data object is done, so reference to storage shouldn't change. @mstajdohar what do you think?

mstajdohar:
1.) You can delete Storage leftovers. We should handle that with rules in DB. 
2.) The output field should not change after status is DONE or ERROR. The input field is set upon create and should not be modified. The descriptor can be changed at any time. You do not have to optimize for this. We do not enforce this rules in the model yet, do we @dblenkus?

dblenkus:
1) This is already implemented. When Data object is deleted, all Storages that reference it will also be deleted.
2) No, we enforce this on API level. I don't see any advantage of enforcing this on model level. All standard workflows go through API, so it is protected. And if you want to change protected field, you will do it somehow :)

resolwe/flow/tests/test_purge.py
```diff
@@ -0,0 +1,223 @@
...
+                output_schema=[field_schema],
+                descriptor={},
+                descriptor_schema=[]
+            )
+
+            for filename in not_removed:
+                self.assertNotIn(os.path.realpath(os.path.join(simulated_root, filename)), unreferenced_files)
+
+            for filename in removed:
+                filename = os.path.realpath(os.path.join(simulated_root, filename))
```
dblenkus: Why do you use `realpath`? It changes the path (`/var/folders/6q/x6bxkpk51xz37hkr06crmk9h0000gn/T/tmpGvPbS2/remove_this/` -> `/private/var/folders/6q/x6bxkpk51xz37hkr06crmk9h0000gn/T/tmpGvPbS2/remove_this`) and causes tests to fail on OS X. Without it tests fail because trailing slashes, but paths are correct.

resolwe/flow/tests/test_purge.py
```diff
@@ -0,0 +1,223 @@
...
+        self.assertFilesRemoved(data, *removed)
+        self.assertFilesNotRemoved(data, *not_removed)
+
+        # Test descriptor.
+        descriptor_schema = DescriptorSchema.objects.create(
+            slug=get_random_string(6),
+            contributor=self.admin,
+            schema=[field_schema]
+        )
+        data = self.create_and_run_processor(
```
dblenkus: Is there any advantage of running two processes for each field? Have you tried how much time can you save without this?
If there is a big impact, you can probably just test the output of `get_purge_files` for each field type. And then write only one test with running processes, as there is the same procedure for all fields once you get list of files/directories to delete.

dblenkus: And if you don't need to run the processes, you can create a separate class for these tests which inherits from `unittest.TestCase` which will speed them up quite a bit.

dblenkus:
Don't forget to add tests mentioned above:
 - works when `data_id` is passed to it - make sure that other data folders are not touched
 - files are deleted only when processing is done

kostko:
> works when data_id is passed to it - make sure that other data folders are not touched

This should be caught by validation of result from `get_purge_files`, no?

> files are deleted only when processing is done

Not sure what this test should check? I mean, when can they be deleted if not only then?

resolwe/flow/tests/test_purge.py
kostko: This really doesn't take a lot of time, for me all the data purge tests take around 1.5sec. And I think that running two independent processors (one specifically for `output` and another one specifically for `descriptor`) is best.

Also we need to run the processor to check if data purge actually gets called after a processor completes.

resolwe/flow/tests/test_purge.py
kostko: > Without it tests fail because trailing slashes, but paths are correct.

This is the exact reason why realpath is used. I didn't think about having the temporary directory inside some symlinked location. I'll change this to just strip slashes instead when comparing.

kostko:
> This should be caught by validation of result from `get_purge_files`, no?

But yes, the actual purge could remove something else. I'll add a test for this.

mstajdohar:
2.) I agree to leave this at the API level. Some times you wish to "cheat the system" and manually prepare/migrate some data on the server directly. So purge should always check input, output and descriptor fields.

kostko:
Wait, why also `input` fields? Those come from other data objects?

mstajdohar:
Sorry, you are right about the input fields.

dblenkus:
> works when data_id is passed to it

Test should check that if there are 2 data objects and  `data_purge(data_ids=[1])` is called only files from first object are deleted

> files are deleted only when processing is done

If you call purge manually while some processes are running, their files shouldn't be deleted even if they are not referenced in database.

dblenkus:
And maybe add another test to ensure that nothing is deleted if `purge_data` is called with `delete=False`.

kostko:
> If you call purge manually while some processes are running, their files shouldn't be deleted even if they are not referenced in database.

By "while some processes are running" you mean to check that the purge data function correctly excludes Data items which are not yet DONE/ERROR?

resolwe/flow/tests/test_purge.py
dblenkus: Yes, I agree that we must run processes for few tests. But I think it is good not to run them too much or without any reason. They add a lot to the complexity of the tests. Not only that the use some time, tests are also way more complex and harder to maintain.

I think that [this](https://github.com/genialis/resolwe/blob/master/resolwe/flow/tests/test_validation.py) is good example of how to write tests (and run processes). First class is used fore some general (e2e) tests and second one is for thorough testing validation function.

dblenkus:
Yes

resolwe/flow/tests/test_purge.py
kostko: The `assertFieldWorks` checks multiple things, but only for a specific field type. And then `test_general` contains a more complex example with multiple fields on the same processor.

I am not sure why do you think the first part is more complex? I wanted to make sure that when you call `assertFieldWorks`, this tests everything for that field. And the different checks that are performed (test simulated operation, test output and test descriptor) are documented in the function. I don't want to see that you need to write two different test functions (or more) for each field, because this means that it is harder to extend and some test cases can be missed when adding a new field.

This makes it easy to extend test cases:
* When adding a new field, one just needs to call `self.assertFieldWorks` with the correct parameters.
* When extending tests for each field, one just needs to modify the `assertFieldWorks` to include additional checks for each field, without the need to modify all the tests.

dblenkus: I understand your arguments and agree with them.

On the other hand it is way harder to maintain tests, as running a process is a complex operation and any change in it reflects in also fixing all tests.
I've just experienced that when I added validation of JSON fields and had to make all processes in tests pass it.

Because of that I try to separate tests which run processes and write unit tests without them.

But you can leave it as is if you feel that changing it will compromise the quality of tests.

kostko: Ok, I see what you mean and yes in this case it makes sense. I will refactor the tests into two separate classes.

kostko:
I've refactored the existing tests (one class for tests which run processors, another one for simple unit tests and a mixin for common cases) and added new tests for the remaining cases.

dblenkus:
@mstajdohar looks like some processes are referencing local files in `list:basic:url:view:` field. Should it be really ignored by purge?
https://github.com/genialis/resolwe-bio/pull/75/files#diff-3b9b7cd957ddb749ccb11eec1acc8d97R195

kostko:
> looks like some processes are referencing local files in list:basic:url:view: field

This sounds completely wrong to me, because in that case that value is not an URL at all, so why is it stored in an URL field?

mstajdohar:
The value can be a html file or the url. In case of `:view` the file/url is opened in iframe. In case of `:link` the file/url opens in another tab. Can support `:file:html:view`, `:file:html:link` and for urls `:url:view` and `:url:link`. If you decide to keep the current solution, you should not purge the file and its references if there is a file in the `:url:*` fields.

kostko:
But still, why is there a non-url in a field with type url? I do not agree with having file paths inside url fields as it doesn't make any sense unless the file urls are clearly prefixed with `file:` and thus form an actual URL. Is there no validation for URL fields?

URL syntax is as follows:
```
 scheme:[//[user:password@]host[:port]][/]path[?query][#fragment]
```

And the URL fields should not allow anything that does not conform to this syntax.

mstajdohar:
The only use of the url field so far has been to view the quality control. The output of the quality control analysis is a web-page and this is probably why we have named the field `basic:url`. In our mind this was a bunch of files hosted on our file server that should be linked or opened in an iframe. We probably wanted to make the solution general to reference other files as well. In the beginning we probably had full urls to the files. A lot has changed since the first version of our platform. We have probably later changed the links it to relative urls, with the introduction of the purge function. Finally we even clipped out the path prefix when we made the platform more general. I agree that now the `basic:file:iframe/link` would be a more appropriate structure. You can support the `basic:file:iframe`, remove the `basic:url` field altogether and change the field type in the processes in resolwe-bio. The `basic:url:view` can probably be substituted simply by `basic:file`. Some small changes will be needed in the dynamic directive, also.

kostko:
> You can support the basic:file:iframe, remove the basic:url field altogether and change the field type in the processes in resolwe-bio. The basic:url:view can probably be substituted simply by basic:file. Some small changes will be needed in the dynamic directive, also.

Ok, so all `basic:file:*` specializations are already supported by purge (I will add a test for this, just to be sure). I can also update the other repositories.

kostko:
Oh, I can't test `basic:file:iframe:` as this fails validation. Can I add this to the type schema, @dblenkus?

dblenkus:
Yes, just don't forget to remove `basic:url` fields.

kostko:
Hm, but probably `basic:url:` fields are still useful for referencing external websites (eg. in descriptors), so I am not sure if we should just remove them? Perhaps just replace them with the new `basic:file:iframe/link` where appropriate (this is, when referencing an internal file, not an external URL)?

dblenkus:
Ok, I agree, then just update the regex for url field.

tjanez:
@kostko, good job with tests' coverage!

Would it make sense to make a test that checks if [error is correctly caught and logged when `data_purge()` raises an exception](https://codecov.io/gh/genialis/resolwe/compare/f4588ef7172288fb416a24aaa9c76ae3fef53de5...95d7b6db2b7187db77afb8beb069a2d05699de92#7265736F6C77652F666C6F772F6578656375746F72732F5F5F696E69745F5F2E7079-232)?

kostko:
Sure, I just need to find a way to generate an exception ;-)

kostko:
Done, logging is now also covered.

dblenkus:
LGTM

PR #108, opened 2016-07-01
Fix pylint

dblenkus:


tjanez:
@dblenkus, good job on fixing all pylint and pep8 errors!

Two comments:
- change the first commit message to "Remove deprecated pylint options"
- add a change log entry saying that pylint and pep8 errors are fixed

PR #110, opened 2016-07-06
Ensure testing utilities are included in the built/installed package

tjanez:
Move Flow's testing utilities from `resolwe.flow.tests` to `resolwe.flow.utils.test` and from `resolwe.permissions.tests.base` to `resolwe.permissions.utils.test`.
Properly include documentation for testing utilities and data purge in the Reference documentation.

@dblenkus, I have a question regarding using absolute imports vs. explicit relative imports. Do we have any preferences? Currently, we mainly use absolute imports, but in some places we use explicit relative imports.


dblenkus:
LGTM, I use relative imports for the same level or if it make sense.

tjanez:
@dblenkus, thanks for the review.

I'll merge this PR even though patch coverage doesn't meet the target since this will make Resolwe Bioinformatics work again with the latest Resolwe *master* branch.

PR #112, opened 2016-07-12
Make ProcessTestCase.assertJSON compatible with Python 3

jkokosar:


docs/CHANGELOG.rst
```diff
@@ -59,6 +59,7 @@ Fixed
 - Set collection name to avoid warnings in test output
 - Improve Python 3 compatibility
 - Fix setting descriptor schema on create
+- Fixed compatibility of ``assertJSON`` with Python3
```
tjanez: This is redundant, as it is already covered two lines above ("Improve Python 3 compatibility").

tjanez:
A comment regarding the commit message:
- Be more descriptive about the changed function, so write "ProcessTestCase.assertJSON" instead of "assertJSON".
- Use "Python 3" instead of "Python3".

tjanez:
@jkokosar, how did you trigger the problem? Or in other words, do you have reproducer we could use as a base for the test?

jkokosar:
@tjanez The error was ``'TypeError: the JSON object must be str, not 'bytes'`` when validating the JSON storage field content. Test files used by ``ProcessTestCase.assertJSON``are ``.gz`` files read by ``gzip.open()`` which produces bytes stream. This worked in Python 2 but is not compatible with Python 3.

tjanez:
@jkokosar, thanks for explaining the issue.

I think the fix could even be simpler. `gzip.open` opens files in [`rb` (i.e. read-only binary) mode by default](https://docs.python.org/3/library/gzip.html#gzip.open). You just need to set the `mode` to `rt` and the file will be opened as `str`, which will be happily consumed by `json.load`.

There's still an issue with the test coverage. I can help you write the test, if you want.

jkokosar:
@tjanez Great suggestion! Updated and tested using Resolwe-bio tests. Please do advise on writing the Resolwe test for this function.

resolwe/flow/utils/test.py
```diff
@@ -424,7 +424,7 @@ def assertJSON(self, obj, storage, field_path, file_name):  # pylint: disable=in
 
             self.fail(msg="Output file {} missing so it was created.".format(file_name))
 
-        with gzip.open(file_path) as f:
+        with gzip.open(file_path, 'rt') as f:
```
tjanez: I prefer being explicit, so use `mode='rt'` in in place of `'rt'`.

tjanez:
@jkokosar, I'll write the test and add the commit to the PR.

tjanez:
@dblenkus, I've added the tests for the `ProcessTestCase.assertJSON` function. Please, review them.

I've had an issue when using [`io.BytesIO` objects](https://docs.python.org/3/library/io.html#binary-i-o) with Python 2.7's version of `gzip.open`.
It turns out `gzip.open` only gained support for *filename* being a file object in [Python 3.3+](https://docs.python.org/3/library/gzip.html#gzip.open).

I see two options to work-around this limitation:
- skip the tests on Python 2
- make the tests work with real files in a temporary directory

For the time being, I've chosen the first option, because I like the speed and simplicity of using `io.BytesIO` objects.

tjanez:
@jkokosar, please see the diff of the first patch. I've spotted another instance where things failed on Python 3.

resolwe/flow/tests/test_test_framework.py
```diff
@@ -49,3 +56,94 @@ def date_in_line(x):
...
+            new_gzipped_json_file = io.BytesIO()
+
+            join_mock = mock.MagicMock(side_effect=[new_gzipped_json_file])
+            isfile_mock = mock.MagicMock(return_value=False)
+            with mock.patch.object(os.path, 'join', join_mock):
+                with mock.patch.object(os.path, 'isfile', isfile_mock):
+                    with six.assertRaisesRegex(self, AssertionError,
+                                               'Output file .* missing so it was created.'):
+                        ProcessTestCase.assertJSON(dummy_case, obj_mock, storage_mock, '', 'foo.gz')
+            self.assertEqual(gzipped_json_file.getvalue(), new_gzipped_json_file.getvalue())
```
dblenkus: It will be better (more explicit), to read and unzip value the content of `new_gzipped_json_file` and assert it with `example_json`.

tjanez: @dblenkus, thanks for your comment. I've changed the test to compare the unzipped JSON with `example_json`. Please, re-review!

resolwe/flow/tests/test_test_framework.py
```diff
@@ -49,3 +56,94 @@ def date_in_line(x):
                     return x.startswith(b"date")
 
                 ProcessTestCase._assert_file(dummy_case, obj_mock, "", "", filter=date_in_line)
+
+        @unittest.skipIf(sys.version_info.major < 3, "Needs Python 3 version of the gzip module")
```
dblenkus: Use `six.PY2`

tjanez: Done

dblenkus:
LGTM

mstajdohar:
Nice work @tjanez!

PR #113, opened 2016-07-12
Sphinx extension for automatic process documentation

JureZmrzlikar:


tjanez:
Please, add a change log entry and improve the commit message accordingly.

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,393 @@
...
+        Find the starting line of process definition - where slug is defined
+
+        :param path/str fname: Filename with processes
+        :param string slug: slug of the process
+        :return: line where the process definiton starts
+        :rtype: int
+        """
+        with open(fname) as file_:
+            for i, line in enumerate(file_):
+                if 'slug: ' + slug in line:
```
dblenkus: There is no guarantee that there will be only one space between `'slug:'` and actual `slug`. Use `if re.search('slug:\s*{}'.format(line), line):` instead.

JureZmrzlikar:
Do not accept yet, there is an issue with hyperlinks to GitHub...

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
+"""
+Sphinx extension for semi-autmoatic documentation of Resolwe processes.
+
+This module introduces new directive that enables user to document a
```
tjanez: "new directive" -> "a new directive"

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+
+    .. autoprocess::
+
+    .. autoprocess:: category_name
+
+    .. autoprocess:: process_slug
+
+The first option documents *all* processes.
+
+The second one documents just the ones that have *category* equal to
```
tjanez: Since you have 'ones' afterwards, I would use "The second option documents..."

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+`category_name`. Since some categories have subcategories (analyses,
```
tjanez: Shouldn't you put `` `category_name` `` in double backticks?

I would change the "Since some categories ..." to "This means that subcategories (e.g. ``` ``analyses:alignment`` ```, ``` ``analyses:variants`` ```) of a category (e.g. ``` ``analyses`` ```) will *not* be documented."

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
+"""
+Sphinx extension for semi-autmoatic documentation of Resolwe processes.
```
tjanez: "semi-autmoatic" -> "semi-automatic"

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+    .. autoprocess:: process_slug
+
+The first option documents *all* processes.
+
+The second option documents just the ones that have *category* equal to
+`category_name`. Since some categories have subcategories (analyses,
+analyses:alignment, analyses:variants), this means that subcategories
+will *not* be documented.
+
+Third option decuments only one process: the one with *slug* equal to
```
tjanez: "decuments" -> "documents"

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+`process_slug`.
```
tjanez: Same as before, shouldn't you put `` `process_slug` `` in double backticks?

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+import six
+import yaml
+
+from docutils import nodes
+from docutils.parsers.rst import Directive
+
+
+class AutoProcessDirective(Directive):
+    """
+    Directive used to automatically document Resolwe process.
```
tjanez: "Resolwe process" -> "a Resolwe process"

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+    final_argument_whitespace = False
+    option_spec = None
+
+    def filter_processes(self, content, categories, processes_dict):
+        """
+        Filter processes list by the content of the directive
+
+        Content can be:
+
+            * None - if there is not content, all processes are returned.
```
tjanez: "not content" -> "no content"

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+            * Category - if the content of the directive is a string matching
+              an element in categories list, only processes with this
+              category will be returned.
+            * Process slug - if the content of directive is a valid process
```
tjanez: "directive" -> "the directive"

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+              slug only corresponding process is returned. Actually, a list
```
tjanez: "slug" -> "slug,"
"corresponding process" -> "the corresponding process"

I would simplify the text "Actually, a list..." with "... corresponding process is returned (packed in a list)."

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+              with only this process is returned.
+            * If none of this options, then raise a ValueError
```
tjanez: This is more a thing of one's taste but I think this line could be omitted entirely since you open the enumeration with "Content can be:" which implies that if the content is not any of the enumerated options, an error will be raised.

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+
+        :param list content: content of the directive
+        :param set categories: set of all categories
+        :param dict processes_dict: dict of all source_base_url: processes pairs
+        :rtype: list
```
tjanez: Shouldn't you also include `:returns:` and describe what the method returns?

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+        """
+        if not content:
+            return processes_dict
+        elif len(content) > 1:
+            raise NotImplementedError("Only one argument can be passes to autoprocess.")
+        content = content[0]
+        # Make a filtered dict of all process that have name == content
+        querry_for_slug = dict(pair for pair in processes_dict.items() if pair[1]['slug'] == content)
```
tjanez: I think the code would be simpler if you would use:
```python
{url, process for url, process in six.iteritems(processes_dict) if process['slug'] == content}
```

I used `six.iteritems` instead of `.items()` since the latter is inefficient on Python 2.

tjanez: I think `querry_for_slug` is a bit awkward. What about `filter_by_slug` or `filtered_by_slug` or just `filtered_dict`?

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+
+        :param list content: content of the directive
+        :param set categories: set of all categories
+        :param dict processes_dict: dict of all source_base_url: processes pairs
+        :rtype: list
+        """
+        if not content:
+            return processes_dict
+        elif len(content) > 1:
+            raise NotImplementedError("Only one argument can be passes to autoprocess.")
```
tjanez: "passes" -> "passed"

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+        content = content[0]
+        # Make a filtered dict of all process that have name == content
+        querry_for_slug = dict(pair for pair in processes_dict.items() if pair[1]['slug'] == content)
+        if len(querry_for_slug) == 1:
+            return querry_for_slug
```
tjanez: Shouldn't you return a `list` not a `dict`?

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+
+        elif content in categories:
+            filtered_dict = dict(
```
tjanez: Similar to above, use dict comprehesion and unpack items into keys and values.

resolwe/flow/utils/docs/autoprocess.py
tjanez: Please, also document raised exceptions with [`:raises ...:`](http://www.sphinx-doc.org/en/stable/domains.html#info-field-lists).

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+            return filtered_dict
+
+        raise ValueError("Invalid argument given to autoprocess directive.")
+
+    def get_process_definition_start(self, fname, slug):
+        """
+        Find the starting line of process definition - where slug is defined
+
+        :param path/str fname: Filename with processes
+        :param string slug: slug of the process
```
tjanez: What about "Process' slug"?

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+        :return: line where the process definiton starts
+        :rtype: int
+        """
+        with open(fname) as file_:
+            for i, line in enumerate(file_):
+                if re.search(r'slug:\s*{}'.format(slug), line):
+                    return i + 1
+        # In case starting line is not found just return first line
+        return 1
```
tjanez: Hmm... wouldn't it be better to either document this behaviour in the docstring or raise an exception instead or returning 1?

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+
+    def get_all_processes(self, process_dir, source_base_url):
+        """
+        Return a dictionary of all processes in `process_dir`
+
+        Dictionary key is an URL's to process source code and value is
+        process as it is loaded from yaml file.
+
+        :param path/str process_dir: Parent directory of all process files
```
tjanez: Same as above regarding `path/str`. My suggestion for the description is "Path to the directory where to search for processes".

resolwe/flow/utils/docs/autoprocess.py
dblenkus: It is not crucial feature, just nice to have, so I'm against raising error.

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+        with open(fname) as file_:
+            for i, line in enumerate(file_):
+                if re.search(r'slug:\s*{}'.format(slug), line):
+                    return i + 1
+        # In case starting line is not found just return first line
+        return 1
+
+    def get_all_processes(self, process_dir, source_base_url):
+        """
+        Return a dictionary of all processes in `process_dir`
```
tjanez: I wouldn't hard-code `` `process_dir` `` and rather use "Return a dictionary of all processes in the given directory". The ``process_dir`` parameter is already nicely described below.

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+            """Read yaml file"""
+            with open(fname) as file_:
+                return yaml.load(file_)
+
+        all_processes = {}
+        for yaml_file in all_process_files:
+            processes_in_file = read_yaml_file(yaml_file)
+            for process in processes_in_file:
+
+                # This section finds the line in file where the
```
tjanez: This comment/description is probably redundant since `get_process_definition_start()` is already nicely documented?

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+                # defintion of the process starts. (there are
+                # multiple process definition in some files).
+                startline = self.get_process_definition_start(yaml_file, process['slug'])
+
+                # Put together URL to starting line of process definition.
+                cutted_yaml_path = yaml_file.split(process_dir)
+                source_url = source_base_url + cutted_yaml_path[1] + '#L' + str(startline)
+                chosen_process = [proc for proc in processes_in_file if proc['slug'] == process['slug']]
```
tjanez: What about replacing the code below with:

```python
process_slug = process['slug']
if any(proc['slug'] == process_slug for proc in processes_in_file):
    raise ValueError("More than one process with the same slug: {}".format(process_slug))
all_processes[source_url] = process
```

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+            processes_in_file = read_yaml_file(yaml_file)
+            for process in processes_in_file:
+
+                # This section finds the line in file where the
+                # defintion of the process starts. (there are
+                # multiple process definition in some files).
+                startline = self.get_process_definition_start(yaml_file, process['slug'])
+
+                # Put together URL to starting line of process definition.
+                cutted_yaml_path = yaml_file.split(process_dir)
```
tjanez: No need for this, it is actually a bit deceiving since `str.split()` is usually not meant for splitting path names.

Just use `os.path.basename(yaml_file)` in place of `cutted_yaml_path[1]` on the next line.

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+                for filename in fnmatch.filter(filenames, extension):
+                    all_process_files.append(os.path.join(root, filename))
+
+        def read_yaml_file(fname):
+            """Read yaml file"""
+            with open(fname) as file_:
+                return yaml.load(file_)
+
+        all_processes = {}
+        for yaml_file in all_process_files:
```
tjanez: Why not use `process_file` instead of `yaml_file` for consistency?

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+                if re.search(r'slug:\s*{}'.format(slug), line):
+                    return i + 1
+        # In case starting line is not found just return first line
+        return 1
+
+    def get_all_processes(self, process_dir, source_base_url):
+        """
+        Return a dictionary of all processes in `process_dir`
+
+        Dictionary key is an URL's to process source code and value is
```
tjanez: This description should go to the `:return:` info field. Also, the current wording is a bit hard to understand. Try to use something like: ":return: Dictionary of processes where keys are URLs pointing to processes' source code and values are processes' definitions parsed from YAML files"

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+        process as it is loaded from yaml file.
+
+        :param path/str process_dir: Parent directory of all process files
+        :param url/string source_base_url: Base of the url to source-code repository
```
tjanez: Same as above regarding `url/string`.

What about using `base_source_url` as the parameter name and "Base URL of the source code repository with process definitions" as the description?

resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -0,0 +1,395 @@
...
+                if pair[1].get('category', 'uncategorized') == content)
+            return filtered_dict
+
+        raise ValueError("Invalid argument given to autoprocess directive.")
+
+    def get_process_definition_start(self, fname, slug):
+        """
+        Find the starting line of process definition - where slug is defined
+
+        :param path/str fname: Filename with processes
```
tjanez: I think `path/str` type definition is not accepted by Sphinx. Just use `str` since that is the type of the parameter and perhaps change the description to something like "Path of the file where process is defined"?

tjanez:
@JureZmrzlikar, I'll stop with the review for now and wait for your feedback and improvements.

Ping when you are ready to continue.

mstajdohar:
Please write unit tests.

mstajdohar:
@JureZmrzlikar will not have time for this, so I will take over. @tjanez, I will address your comments. I propose we accept this without tests and write tests when we have more resources.

JureZmrzlikar:
Tests are almosat written and @tjanez  comments can be adressed relatively quickly. I can do it until todays evening. Or just hand you the work i did already.

mstajdohar:
I will close this in favor of #121. I had to quick fix a py3 related issue. @tjanez, please have another review pass of #121 even though it has been merged.

PR #115, opened 2016-07-15
Add 'name' process tamplate tag

jkokosar:


mstajdohar:
Write tests.

resolwe/flow/templatetags/process_fields.py
```diff
@@ -6,10 +6,31 @@
...
 
 
+def get_data_attr(data, attr):
+    if isinstance(data, dict):
+        # `Data` object's id is hydrated as `__id` in expression engine
+        data = data['__id']
+
+    data_obj = Data.objects.filter(id=data)
+
+    if not data_obj.exists():
```
mstajdohar: We call `exists` and `first`. Are we doing 2 queries to DB or does Django handle this efficiently? Wouldn't it be more efficient to just call `first` with a try/catch?

resolwe/flow/templatetags/process_fields.py
```diff
@@ -6,10 +6,31 @@
 """
 from django import template
 
+from resolwe.flow.models import Data
+
 
 register = template.Library()  # pylint: disable=invalid-name
 
 
+def get_data_attr(data, attr):
```
mstajdohar: Start private functions with an underscore.

resolwe/flow/tests/test_templatetags.py
```diff
@@ -0,0 +1,70 @@
...
+
+from django.conf import settings
+from django.contrib.auth import get_user_model
+from django.test import TestCase
+
+from resolwe.flow.managers import manager
+from resolwe.flow.models import Process, Data
+
+
+class ProcessFieldsTagsTest(TestCase):
```
mstajdohar: A long test for such small functionality. Couldn't you just create a Data obj and call Django's render function? What about border cases, e.g. no data object?

dblenkus: Problem is that `Data` objects are hydrated in a special way before rendering (i.e. `id` becomes `__id`). I can fake that, but I've decided to run real process as this is one of the significant things that are being tested.

resolwe/flow/templatetags/process_fields.py
dblenkus: +1

PR #116, opened 2016-07-19
Determine Storage pemissions based on referenced Data

dblenkus:


resolwe/permissions/shortcuts.py
```diff
@@ -194,3 +198,165 @@ def format_permissions(perms):
         })
 
     return perms_list
+
+
+# based on guardian.shortcuts.get_objects_for_user
```
mstajdohar: You could write this in the function's docstring.

resolwe/permissions/shortcuts.py
```diff
@@ -194,3 +198,165 @@ def format_permissions(perms):
...
+def get_objects_for_user(user, perms, klass=None, use_groups=True, any_perm=False,
+                         with_superuser=True, accept_global_perms=True, perms_filter='pk__in'):
```
mstajdohar: Are just the lines that include `perms_filter` are different from the original implementation in REST framework?

resolwe/permissions/filters.py
```diff
@@ -0,0 +1,23 @@
...
+class ResolwePermissionsFilter(BaseFilterBackend):
+
+    def filter_queryset(self, request, queryset, view):
+        user = request.user
+
+        app_label = queryset.model._meta.app_label
+        model_name = queryset.model._meta.model_name
+
+        kwargs = {}
+        if model_name == 'storage':
```
kostko: Could this be changed so that it (`model_name` and `perms_filter`) is configured on the viewset instead of being hardcoded here?

tests/settings.py
```diff
@@ -117,7 +117,7 @@
         'rest_framework.authentication.SessionAuthentication',
     ),
     'DEFAULT_FILTER_BACKENDS': (
-        'rest_framework.filters.DjangoObjectPermissionsFilter',
+        'resolwe.permissions.filters.ResolwePermissionsFilter',
```
mstajdohar: I have this crazy idea. Why don't we leave the original `DjangoObjectPermissionsFilter`. This would work as intended for all our models except Storage, right. For Storage it would return all objects. Then, just for Storage, we add another StoragePermissionsFilter after all the other filters. This filter can be specific for Storage but much shorter then `get_objects_for_user `, and it can filter-out the objects missing permissions.

resolwe/permissions/tests/test_shortcuts.py
```diff
@@ -176,3 +188,435 @@ def test_user_permissions(self):
...
+        self.assertEqual(resp.status_code, status.HTTP_404_NOT_FOUND)
+
+    def test_not_allowed_methods(self):
+        self.assertFalse(hasattr(StorageViewSet, 'update'))
+        self.assertFalse(hasattr(StorageViewSet, 'partial_update'))
+        self.assertFalse(hasattr(StorageViewSet, 'destroy'))
+        self.assertFalse(hasattr(StorageViewSet, 'create'))
+
+
+# tests copied from guardina.testapp.tests.test_shortcuts
```
mstajdohar: Are this tests any different from the original? Could we simply modify the Tox script to run them directly?

tests/settings.py
kostko: Interesting idea, I also don't like that currently the storage use case is hard-coded in the filter. So we should either generalize the filter (make the filter configurable from the viewset as I suggested in the comment above) or do as Miha suggests here.

mstajdohar:
Done reviewing, @dblenkus please review comments.

resolwe/permissions/tests/test_shortcuts.py
dblenkus: Yes, tests are the same (I've only deleted few of them that are difficult to run and are not relevant to us). The only difference is that different `get_objects_for_user` function is used, so their tests cannot be run directly.

resolwe/permissions/shortcuts.py
dblenkus: Yes. There are two commits, so you can easily check what has been changed.

tests/settings.py
dblenkus: I was also considering this option, but came to the conclusion that we would have two pieces of code with exactly the same functionality (only difference is a join when filtering permissions - second commit in this PR).
So if we write our own function we should use it for all models. But why should we write it, if it is already written and we just need to modify it a bit?

resolwe/permissions/filters.py
dblenkus: Agree, I will check what can be done.

PR #118, opened 2016-07-21
Make ProcessTestCase._debug_info() robust to non-ascii stdout.txt

jkokosar:
Fix for error:
`  File "/Users/janez/resolwe/resolwe/flow/utils/test.py", line 442, in _debug_info
    msg += fn.read()
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 2017: ordinal not in range(128)`

### DIFF
--- a/resolwe/flow/utils/test.py
+++ b/resolwe/flow/utils/test.py
@@ -9,6 +9,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import copy
 import hashlib
 import gzip
+import io
 import json
 import os
 import shutil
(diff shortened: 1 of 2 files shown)


docs/CHANGELOG.rst
```diff
@@ -109,6 +109,7 @@ Fixed
 - Ensure Tox installs the package from sdist
 - Pass all Resolwe's environment variables to Tox's testing environment
 - Ensure tests gracefully handle unavailability of Docker
+- Fix reading stdout.txt containing non-ascii characters
```
tjanez: This is inserted at the wrong place.

Also, I think this small change is not worthy of a change log entry.

resolwe/flow/utils/test.py
```diff
@@ -438,7 +439,7 @@ def _debug_info(self, data):
         path = os.path.join(settings.FLOW_EXECUTOR['DATA_DIR'], str(data.pk), "stdout.txt")
         if os.path.isfile(path):
             msg += "\nstdout.txt:\n" + 11 * "-" + "\n"
-            with open(path, 'r') as fn:
+            with codecs.open(path, 'r', encoding='utf-8') as fn:
```
tjanez: I would prefer if we use `io.open`.

For more details see [*io.open vs. codecs.open*](https://mail.python.org/pipermail/python-list/2015-March/687124.html).

tjanez:
@dblenkus, please review the added test.

PR #119, opened 2016-07-22
Support exporting data from processes

dblenkus:


resolwe/flow/executors/__init__.py
```diff
@@ -97,9 +103,10 @@ def run(self, data_id, script, verbosity=1):
 
         self.data_id = data_id
 
+        data_dir = settings.FLOW_EXECUTOR['DATA_DIR']
         dir_mode = getattr(settings, 'FLOW_EXECUTOR', {}).get('DATA_DIR_MODE', 0o755)
```
mstajdohar: We'll have a lot of this. Maybe we should write a helper function, e.g.:
```
settings_value(**keys)
```
that you would call in this case as:
```
settings_value('FLOW_EXECUTOR', 'DATA_DIR')
```
It would return the value on the arbitrary dictionary depth, say:
```
settings_value('TEMPLATES', 'OPTIONS', 'context_processors')
```
@tjanez what do you think? You usually have good suggestions for naming such functions.

dblenkus: We should add default settings and then not care if they are define or not. [This](https://docs.djangoproject.com/en/1.9/topics/settings/#custom-default-settings) may be interesting, if it is possible to define per-app defaults. Otherwise [DRF](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/settings.py) has such implementation, but it looks a bit complicated to me.

@kostko do you have any suggestion?

dblenkus:
@mstajdohar exposed proposal is not a part of this PR, so we should discuss it separately.

Is this ok to merge?

resolwe/flow/executors/__init__.py
```diff
@@ -140,6 +147,16 @@ def run(self, data_id, script, verbosity=1):
 
                         for obj in iterjson(line[3:].strip()):
                             spawn_processors.append(obj)
+                    elif line.strip().startswith('export'):
+                        file_name = line[6:].strip()
+
+                        export_folder = settings.FLOW_EXECUTOR['UPLOAD_DIR']
+                        export_path = os.path.join(export_folder, str(data_id))
```
mstajdohar: I do not like using data IDs. I would rather rename the files to some random UUID name when we move them to the upload dir.

resolwe/flow/executors/__init__.py
```diff
@@ -140,6 +147,16 @@ def run(self, data_id, script, verbosity=1):
...
+
+                        if not os.path.isdir(export_path):
```
mstajdohar: If you use a random name, this is not necessary.

resolwe/flow/executors/__init__.py
```diff
@@ -140,6 +147,16 @@ def run(self, data_id, script, verbosity=1):
 
                         for obj in iterjson(line[3:].strip()):
                             spawn_processors.append(obj)
```
mstajdohar: This should include the original `name` and random `temp_name`. Then the `hydrate_spawned_files` would not be needed anymore.

resolwe/flow/management/commands/register.py
```diff
@@ -50,12 +50,13 @@ def valid(self, instance, schema):
             self.stderr.write("        val. value: {}".format(ex.validator_value))
             return False
 
-    def find_schemas(self, schema_path, filters=None, schema_type='process'):
+    def find_schemas(self, schema_path, filters=None, schema_type='process', verbosity=1):
```
mstajdohar: We should use a logger.

resolwe/flow/management/commands/register.py
```diff
@@ -86,7 +87,7 @@ def find_schemas(self, schema_path, filters=None, schema_type='process'):
 
         return schema_matches
 
-    def register_processes(self, process_schemas, user, force=False):
+    def register_processes(self, process_schemas, user, force=False, verbosity=1):
```
mstajdohar: We should use a logger.

mstajdohar:
Done reviewing, @dblenkus please review comments.

PR #120, opened 2016-07-25
Hydrate contributor data

kostko:


resolwe/flow/serializers.py
```diff
@@ -120,6 +152,7 @@ class CollectionSerializer(ResolweBaseSerializer):
     descriptor_schema = DescriptorSchemaSerializer(required=False)
 
     slug = serializers.CharField(read_only=False, required=False)
+    contributor = ContributorSerializer()
```
dblenkus: Add this to all serializers. Maybe to `ResolweBaseSerializer`?

dblenkus:
Add change log entry and one minor comment. Otherwise LGTM.

dblenkus:
And add tests for all possibilities in `to_internal_value` function

kostko:
Done.

kostko:
Just noticed that we can't really test the correctness of `to_internal_value` as `ResolweCreateModelMixin` always overrides the contributor value.

kostko:
I've fixed tests so that we test the serializer's `to_internal_value` directly.

mstajdohar:
LGTM. Please merge.

hadalin:
@kostko, @mstajdohar Just noticed that `genialis-base/genesis/app_state/models/AppState` model also has a contributor filed, but does not extend resowle base model. Is this ok or should we leave it as it is.

dblenkus:
Please, don't discuss private repos in public ones.

We decided a while ago that all models should be subclasses of `BaseModel`, so this should be handled automatically (serializers also should be subclasses of `ResolweBaseSerializer`). Only think that should be added is `.prefetch_related('contributor')` in viewset. But this is only for performance matter, would work anyway.

PR #124, opened 2016-07-29
Set Data status and proc_error on ValidationError

dblenkus:


mstajdohar:
The solution is a bit awkward is the sense that the validation function writes the error to DB to the same object it validates. This triggers a validation within validation. The code seams to work as expected, but please add extensive tests and a note in change log.

resolwe/flow/models.py
```diff
@@ -479,6 +479,35 @@ def add_dependency(value):
...
+            else:
+                validate_schema(self.output, output_schema, path_prefix=path_prefix,
+                                test_required=False)
+        except ValidationError as ex:
+            # if object doesn't have ``pk`` assigned and doesn't validate, it
+            # won't be stored in database, so we don't have to do anything
+            if self.pk:
+                # current object is already corupted, so we have to get fresh
+                # copy (last that has passed validation)
+                data = Data.objects.get(pk=self.pk)
```
mstajdohar: @kostko, what do you think of this solution. So the validation function is validation a data object. If the validation fails, it saves an error message to the same data object in the database.

kostko: This looks wrong to me. Why do you update the previously valid object in the database when a new update to this object fails? To me it looks like this should just raise a validation error (and abort the save) and the actual error handling should happen somewhere else (inside the process executor?).

mstajdohar: I agree. Please move the error handling to the process executor. The model should raise ValidationError. I will give an example where the current solution behaves strange. Say you wish to change the data object directly on the server in the Django shell. You make a typo in annotations and run `save()`. Instead of getting the ValidationError, the data object would be saved with an ERROR status.

dblenkus: @mstajdohar we discussed this before and agreed that this would be the best place to handle this error, as otherwise it should be handled on (too) many places.
I don't see a problem updating old object, because only "metadata" is updated (actual data, i.e. input and output, aren't touched).
In your example, `ValidationError` would be (re)raised, and the nice thing about this is that Data object would stay consistent even when updating it through shell ( = `status` and `error` would be updated automatically as you probably wouldn't handle it manually).

What do you (and @kostko) think?

kostko: > as otherwise it should be handled on (too) many places.

Where must this be handled besides the process executor?

> In your example, ValidationError would be (re)raised, and the nice thing about this is that Data object would stay consistent even when updating it through shell ( = status and error would be updated automatically as you probably wouldn't handle it manually).

Yes, this is even stranger. An exception gets raised during `save`, but the object is still mutated? This does not interact well with transaction handling. And I don't see this as consistent - there is no error, because the object was not saved (the save was aborted). But the object now shows an error just because some later operation failed on it (and was aborted)?

So in the general case this looks like some totally unexpected behavior when manipulating a database model. It does make sense when you look at it from the process executor's perspective: the process tried to send invalid data to the executor, so the process itself has been aborted and the Data's status reflects that error condition. But, to me this seems like something that only makes sense when implemented inside the executor and not in any other situation.

But in the general case, the erroneous data may come from anywhere and it is not always the correct way to just mutate the Data object so it now displays an error, related to the process. Perhaps the `save` caller wants to handle the validation error differently?

dblenkus: You are right, I've overlooked this. It only makes sense to write error in the database if invalid state is already in the database. So the only such cases are:  
* status is set to `DONE` and validation process changes
* `DescriptorSchema` changes

in all other scenarios it doesn't make sense to write error in the database as invalid data won't be saved.

I'll fix that.

mstajdohar: I know we discussed this, but I was not comfortable with our solution. Thanks @kostko for reminding us of bad practices.

PR #126, opened 2016-08-23
Fix updating DescriptorSchema through API

dblenkus:


resolwe/flow/tests/test_api.py
```diff
@@ -70,38 +80,65 @@ def tearDown(self):
         Data.objects.all().delete()
         Process.objects.all().delete()
         Collection.objects.all().delete()
+        DescriptorSchema.objects.all().delete()
```
kostko: BTW, any special reason why you are not using Django's database-aware test case class here? As it should run everything in transactions and thus clean up stuff automatically after each test?

kostko:
LGTM

resolwe/flow/tests/test_api.py
dblenkus: Fixed, probably just a left over from the old platform, where Mongo was used :)

PR #127, opened 2016-08-24
Support passing Storage object in file

dblenkus:


mstajdohar:
Nice work Domen! Please also mention this in the docs.

dblenkus:
@mstajdohar Added

docs/proc.rst
```diff
@@ -643,6 +643,18 @@ For example:
 
 will store the ``{ JSON blob }`` to the ``etc`` field.
 
+.. note::
```
tjanez: May suggestion for minor improvements:

```
.. note::
    
    Printing a lot ot data to standard output can cause problems when using
    the Docker executor due to its current implementation. Therefore, it is
    advised to save big JSON blobs to a file and only pass the file name to the
    ``re-save`` function.

    For example:

    .. code-block:: bash

        command_that_generates_large_json > json.txt
        re-save etc json.txt
```

resolwe/flow/tests/test_models.py
```diff
@@ -292,6 +300,32 @@ def test_save_storage(self):
...
+            name='Test data',
+            contributor=self.user,
+            process=self.proc,
+        )
+
+        data.output = {'json_field': 'json.txt'}
+        data.status = Data.STATUS_DONE
+
+        isfile_mock = MagicMock(return_value=True)
+        if six.PY3:
```
tjanez: I think a nicer and more general solution is to use `io.BytesIO`, open the stream with `open` and use the `t` mode to open the stream in text mode. Then, one can just use `json.dump` rather than `json.dumps`.

The thing is that `json.dumps` will serialize the given object to a string, which creates problems when trying to write Python 2 & 3 compatible code.

My proposed solution would be:
```python
json_file = io.BytesIO()
with open(json_file, mode='wt') as f:
    json.dump({'foo': 'bar'}, f)
# set seek position of the binary stream object back to 0
json_file.seek(0)

isfile_mock = MagicMock(return_value=True)
open_mock = MagicMock(return_value=json_file)
with patch.object(os.path, 'isfile', isfile_mock):
    with patch.object(builtins, 'open', open_mock):
        data.save()
```

I used a similar thing in [`test_test_framework`'s tests](https://github.com/genialis/resolwe/blob/master/resolwe/flow/tests/test_test_framework.py#L64-L89).

Note that I would also drop `_mock` from the `json_file_mock` variable name since the variable is not a Mock object but rather a real in-memory file.

docs/CHANGELOG.rst
```diff
@@ -12,6 +12,7 @@ Unreleased
 Added
 -----
 - Auto-process style, type tree and category index
+- JSON is loaded from file if it passed to ``basic:json:`` field
```
tjanez: Minor comment. I would find this easier to grasp:

```
- Support loading JSON from a file if the string passed to the ``basic:json:``
  field is a file.
```

tjanez:
Expect for the inline comments, this looks good to me.

As a side question, do you know why are classes in `test_models.py` named so differently, i.e..:
- `DataModelTest`
- `HydrateFileSizeUnitTest`
- `StorageModelTestcase`
- `UtilsTestCase`

dblenkus:
These are only names :). They are not used anywhere, but we should unify them anyway.

tjanez:
Ok, fine with me, we can do than later.

PR #128, opened 2016-09-01
Check documentation style

mstajdohar:
Check documentation style and refactor linters.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,13 +66,10 @@ setup(
         'pyyaml>=3.11',
         'jsonschema>=2.4.0',
         'six>=1.10.0',
+        'Sphinx>=1.4.6',
     ],
     extras_require = {
         'docs':  [
-            # XXX: Intersphinx URLs are broken when using non-default inventory
-            # (like in Django's case) with Sphinx 1.4.3+
-            # Bug report: https://github.com/sphinx-doc/sphinx/issues/2789
-            'Sphinx==1.4.2',
             'sphinx_rtd_theme',
         ],
         'package': [
@@ -80,14 +77,14 @@ setup(
             'wheel',
         ],
         'test': [
-            'coverage>=3.7.1',
-            'pep8>=1.6.2',
-            'pylint>=1.4.3',
-            'testfixtures>=4.10.0',
             'check-manifest',
+            'coverage>=4.2',
+            'pycodestyle>=2.0.0',
+            'pydocstyle>=1.0.0',
+            'pylint>=1.6.4',
             'readme_renderer',
             'resolwe-runtime-utils>=1.1.0',
-            'Sphinx==1.4.2',
+            'testfixtures>=4.10.0',
         ],
     },
 
--- a/.pep8rc
+++ b/.pep8rc
@@ -1,5 +1,0 @@
-[pep8]
-max-line-length=119
-# E127 continuation line over-indented for visual indent - chhecked by pylint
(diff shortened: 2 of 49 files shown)


setup.py
```diff
@@ -72,22 +72,23 @@
             # XXX: Intersphinx URLs are broken when using non-default inventory
             # (like in Django's case) with Sphinx 1.4.3+
             # Bug report: https://github.com/sphinx-doc/sphinx/issues/2789
-            'Sphinx==1.4.2',
+            'sphinx==1.4.2',
```
tjanez: I would leave `Sphinx` with a capital `S` since browsing to https://pypi.python.org/pypi/sphinx redirects you to https://pypi.python.org/pypi/Sphinx, which indicates that `Sphinx` is the correct name to use.

setup.py
```diff
@@ -72,22 +72,23 @@
             # XXX: Intersphinx URLs are broken when using non-default inventory
```
tjanez: I saw that [Sphinx 1.4.6 was released](http://www.sphinx-doc.org/en/stable/changes.html#release-1-4-6-released-aug-20-2016), which fixes sphinx-doc/sphinx#2789 which we encountered.

Would you remove the comment and bump the requirement?

setup.py
```diff
@@ -72,22 +72,23 @@
...
-            'testfixtures>=4.10.0',
             'check-manifest',
+            'coverage>=4.2',
+            'pycodestyle>=2.0.0',
+            'pydocstyle>=1.0.0',
+            'pylint>=1.6.4',
             'readme_renderer',
             'resolwe-runtime-utils>=1.1.0',
-            'Sphinx==1.4.2',
+            'sphinx==1.4.2',
```
tjanez: Same as above

mstajdohar:
Thanks @tjanez, issues fixed. @dblenkus, can you have a look? Let's make this PR a priority and not accept the other PRs before this one. Otherwise we will have conflicts all over the place.

setup.py
```diff
@@ -69,25 +69,23 @@
...
-            'testfixtures>=4.10.0',
             'check-manifest',
+            'coverage>=4.2',
+            'pycodestyle>=2.0.0',
+            'pydocstyle>=1.0.0',
+            'pylint>=1.6.4',
             'readme_renderer',
             'resolwe-runtime-utils>=1.1.0',
-            'Sphinx==1.4.2',
+            'Sphinx>=1.4.6',
```
tjanez: I just noticed that we list `Sphinx` as a requirement twice, once as a `docs` extra requirement and once as a `test` extra requirement. This was erroneously introduced in commit 66e452cbb31827ea7ba464e82270fab4defc2036. Please, remove it from `test` extra requirements.

mstajdohar: It is needed in tests because we develop and test sphinx extensions. Also, we need it in docs to build the actual documentation.

mstajdohar: Maybe we should add it to install_requires and remove it from tests and docs.

tjanez: @mstajdohar, thanks for the explanation, I forgot about the autoprocess Sphinx extension.

I agree with your last comment, we should move it to `install_requires` since we explicitly import Sphinx from [`resolwe/flow/utils/docs/autoprocess.py`](https://github.com/genialis/resolwe/blob/master/resolwe/flow/utils/docs/autoprocess.py#L39).

PR #129, opened 2016-09-08
Add list:basic:integer: field

dblenkus:


resolwe/flow/static/flow/typeSchema.json
```diff
@@ -241,19 +241,19 @@
...
-        "list_data": {
+        "list_integer": {
             "type": "object",
             "required": ["type", "value"],
             "additionalProperties": false,
             "properties": {
                 "type": {
                     "type": "string",
-                    "pattern": "^list:data:[a-z0-9:]+:$"
+                    "pattern": "^list:(basic:integer|data:[a-z0-9:]+):$"
```
kostko: Why is the `data:[a-z0-9:]+:` (without `list:`) present in both `string` and `integer` types, but for lists, it is only present under `list_integer` and not `list_string`? This seems inconsistent.

kostko:
LGTM

resolwe/flow/static/flow/typeSchema.json
dblenkus: Thanks for pointing this out. It is left-over from the old platform, where ids were uuids. I've fixed it and added test

kostko: Yes, I've seen the fix and the test, everything looks good now.

PR #133, opened 2016-10-11
Update dependencies

dblenkus:


resolwe/flow/views.py
```diff
@@ -532,4 +532,4 @@ class StorageViewSet(mixins.RetrieveModelMixin,
 
     queryset = Storage.objects.all().prefetch_related('contributor')
     serializer_class = StorageSerializer
-    filter_fields = ('contributor', 'name', 'created', 'modified', 'slug', 'json')
```
tjanez: Why did you put this in the same commit as updating of dependencies?

setup.py
```diff
@@ -46,23 +46,18 @@
         ]
     },
     install_requires=[
```
tjanez: Great job with updating dependencies and removing some old pinnings :+1: !

resolwe/flow/views.py
dblenkus: Because this fails with new `djangorestframework-filters`. But as discussed with @lukaw3d this is not used and it also doesn't make sense to filter by `json`field.
`Storage` is always accessed through `Data` object so we don't need any advanced filtering, only desired functionality is to allow to download part ot the `json` filed

PR #135, opened 2016-10-14
Add get_or_create endpoint for Data objects

dblenkus:


resolwe/flow/migrations/0011_calculate_checksum.py
```diff
@@ -0,0 +1,27 @@
...
+# Generated by Django 1.9.8 on 2016-10-13 09:15
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+from resolwe.flow.utils import get_data_checksum
+
+
+def calculate_checksum(apps, schema_editor):
+    # We can't import the Person model directly as it may be a newer
```
kostko: What `Person` model? And why this comment, this is something that is known when doing migrations as you should always use historic versions of models in migrations.

resolwe/flow/migrations/0010_fix_jsonfields.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('flow', '0009_data_parents'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
kostko: What exactly has changed here? I don't see this in any of the models changes below?

resolwe/flow/utils/__init__.py
```diff
@@ -11,3 +11,14 @@
 
 """
 from __future__ import absolute_import, division, print_function, unicode_literals
+
+import hashlib
+
+
+def get_data_checksum(proc_input, proc_slug, proc_version):
+    """Compute checksum of processor inputs, name and version."""
+    hash_ = hashlib.sha1()
```
kostko: Why the `_`, since `hash` is not a reserved keyword, similar to `id`, so you can override it without any issue? If this bothers you, use something like `checksum`.

dblenkus: It's a built-in function and pylint is not happy with overriding it. I've renamed it to `checksum`

resolwe/flow/migrations/0011_calculate_checksum.py
dblenkus: Ops, I've forgot to delete the comment from Django's documentation :)

PR #136, opened 2016-10-19
Use default values to calculate checksum

dblenkus:


hadalin:
Works from the frontend. Super fast also 🎉

resolwe/flow/views.py
```diff
@@ -215,7 +216,14 @@ def create(self, request, *args, **kwargs):
         # perform "get_or_create" if requested - return existing object
         # if found
         if kwargs.pop('get_or_create', False):
-            checksum = get_data_checksum(request.data.get('input', {}), process.slug, process.version)
+            process_input = request.data.get('input', {})
```
mstajdohar: Shouldn't we compute the checksum in the Data model's `save()` function?

PR #138, opened 2016-10-24
Refactor expression engines

kostko:


docs/CHANGELOG.rst
```diff
@@ -22,6 +22,8 @@ Added
 Changed
 -------
 - Refactor linters, check PEP 8 and PEP 257
+- Split expression engines into expression engines and execution engines
+- Use Jinja2 instead of Django Template syntax
```
hadalin: Out of curiosity, is this to put logic into templates or ...?

kostko: No, this is because we will require some Jinja2 API to implement inline expressions correctly in workflows. And it also reduces coupling between processes and Django, which is IMO a good thing.

kostko: Because previously, we had things like `templatetags`, which were used in processes. This looked quite hacky to me. Now, filters for use in process definition expressions are defined using separate modules:
```python
FLOW_EXPRESSION_ENGINES = [
    {
        'ENGINE': 'resolwe.flow.expression_engines.jinja',
        'CUSTOM_FILTERS': [
            'resolwe_bio.expression_filters.sample',
        ]
    },
]
```

resolwe/flow/managers/base.py
```diff
@@ -134,38 +153,67 @@ def load_executor(self, executor_name):
                 # If there's some other error, this must be an error in Django
                 raise
 
-    def load_exprengines(self, exprengine_list):
-        """Load expression engines."""
-        exprengines = {}
+    def _load_engines(self, class_name, base_module, engines):
```
dblenkus: Is there an easy way to move this function out of the class and also use it to load manager and executor?

resolwe/flow/models.py
```diff
@@ -273,22 +272,10 @@ class Meta(BaseModel.Meta):
...
-        raise KeyError("`RESOLWE_CUSTOM_TEMPLATE_TAGS` setting must be a list.")
-
-    template_headers.extend(
-        ['{{% load {} %}}'.format(template_tag) for template_tag in custom_template_tags]
-    )
+    """Render template using the Jinja expression engine."""
+    from .managers import manager
 
-    return template.Template(''.join(template_headers) + template_string).render(context)
+    return manager.get_expression_engine('jinja').evaluate_block(template_string, context)
```
dblenkus: Why is `jinja` hardcoded here?

resolwe/flow/engine.py
```diff
@@ -0,0 +1,16 @@
...
+
+class BaseEngine(object):
+    """Base class for all engines in Resolwe workflow."""
+
+    name = None
+
+    def __init__(self, manager, settings=None):
+        """Construct a Resolwe engine."""
+        self.manager = manager
+        self.settings = settings or {}
```
kostko: But then all invocations would share the same instance? I know that we are not modifying `settings` but I think that using `None` and constructing a new dictionary on each invocation is good practice.

resolwe/flow/expression_engines/jinja/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def _register_custom_filters(self):
+        """Register any custom filter modules."""
+        custom_filters = self.settings.get('CUSTOM_FILTERS', [])
+        if not isinstance(custom_filters, list):
+            raise KeyError("`CUSTOM_FILTERS` setting must be a list.")
+
+        for filter_module_name in custom_filters:
+            try:
+                filter_module = import_module('{}'.format(filter_module_name))
```
kostko: To avoid saying `str(x)` or `unicode(x)`, which could be different among Python 2/3. I've seen this elsewhere in Resolwe so I used a similar pattern here.

resolwe/flow/expression_engines/jinja/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+                    raise TypeError
+            except (AttributeError, TypeError):
+                raise ImproperlyConfigured(
+                    "Filter module '{}' does not define a 'filters' dictionary".format(filter_module_name)
+                )
+            self._environment.filters.update(filter_map)
+
+    def evaluate_block(self, template, context=None):
+        """Evaluate a template block."""
+        if context is None:
```
kostko: See above.

resolwe/flow/expression_engines/jinja/filters.py
```diff
@@ -0,0 +1,56 @@
...
+    """Return `name` of `Data`."""
+    return _get_data_attr(data, 'name')
+
+
+def id(obj):  # pylint: disable=redefined-builtin,invalid-name
+    """Return ``id`` key of dict."""
+    return obj['__id']
+
+
+def type(obj):  # pylint: disable=redefined-builtin
```
kostko: I've just copied this over from previous `templatetags` and did not change the implementation at all. So, should we change it?

resolwe/flow/models.py
kostko: Well, not sure how this would be determined otherwise? Previously this always used DTL. We currently don't have any process-wise definition of which expression engine should be used for fields like `data_name`. In CWL, there are the `requirements` that a process can define, but we don't have that (only a language for the program, but this is something different).

One possible way would be to add something like the `requirements` from CWL, so the process definition would have to say something like this at the top level:
```yaml
requirements:
    - expression-engine:jinja
```

And then everything would use that expression engine. We could also then just have the `bash` execution engine instead of `bash-jinja` and reuse this configuration also for expressions inside the program. This would also bring our implementation closer to CWL. What do you think?

/cc @mstajdohar @tjanez

resolwe/flow/managers/base.py
kostko: Good idea. Done in the latest push.

resolwe/flow/expression_engines/jinja/filters.py
kostko: Done.

resolwe/flow/expression_engines/jinja/__init__.py
tjanez: What about [`six.u()`](https://pythonhosted.org/six/#six.u)?

More compact and *probably* faster...

resolwe/flow/models.py
tjanez: I'm in favour of @kostko's proposal.

1. I think it's cleaner to specify the expression engine for the whole process definition.

2. Being closer to CWL is a nice bonus.

kostko: Done.

resolwe/flow/execution_engines/bash/__init__.py
```diff
@@ -0,0 +1,38 @@
+"""An execution engine that outputs bash programs."""
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from django.conf import settings
+
+from resolwe.flow.execution_engines import BaseExecutionEngine, ExecutionError
```
mstajdohar: `bash` is a submodule of `execution_engines`, however we do not import it in `execution_engines` (yet). Could this introduce a circular import in the future?

resolwe/flow/managers/base.py
```diff
@@ -92,80 +90,49 @@ def communicate(self, run_sync=False, verbosity=1):
                         data.save()
                         continue
 
-                    script = self.exprengines['dtlbash'].eval(data)
-                    if script is None:
+                    try:
+                        execution_engine = data.process.run.get('language', 'bash')
+                        # TODO: Make it an error if there is no explicit language.
+                        program = self.get_execution_engine(execution_engine).evaluate(data)
```
mstajdohar: What happens when an invalid execution engine (say `python`) is defined in the process? Write a test and handle the error.

resolwe/flow/execution_engines/bash/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+
+            script_template = data.process.run.get('program', '')
+
+            # Get the appropriate expression engine. If none is defined, do not evaluate
+            # any expressions.
+            expression_engine = data.process.requirements.get('expression-engine', None)
+            if not expression_engine:
+                return script_template
+
+            return self.get_expression_engine(expression_engine).evaluate_block(script_template, inputs)
```
mstajdohar: What happens if the given expression_engine does not exist?

resolwe/flow/models.py
```diff
@@ -271,24 +270,20 @@ class Meta(BaseModel.Meta):
...
-        '{% load process_fields %}',
-        '{% load mathfilters %}',
-    ]
 
-    custom_template_tags = getattr(settings, 'RESOLWE_CUSTOM_TEMPLATE_TAGS', [])
-    if not isinstance(custom_template_tags, list):
-        raise KeyError("`RESOLWE_CUSTOM_TEMPLATE_TAGS` setting must be a list.")
+def render_template(process, template_string, context):
+    """Render template using the specified expression engine."""
+    from .managers import manager
```
mstajdohar: Why don't you import the manager at the top of the file?

resolwe/flow/execution_engines/bash/__init__.py
kostko: Why would you import it there? The execution engines need to be defined in the settings file and are registered by the manager.

resolwe/flow/models.py
kostko: Because this would cause an `ImportError` due to circular imports.

resolwe/flow/execution_engines/bash/__init__.py
mstajdohar: I don't know. To be safe I prefer base classes in base.py, exceptions in exceptions.py and in __init__.py just import what I need in the namespace.

resolwe/flow/models.py
mstajdohar: I think we had problems with this kind of avoiding circular imports and Celery before. When processes spawn other processes, the manager is called on the worker. Some Python modules does not copy correctly to workers and import fails. Can you try if this PR works for the spawned process use case (_e.g.,_ demultiplexing) with Celery. @dblenkus, can you give more information about the problems we had?

mstajdohar: I do not know the mechanics exactly, but somehow Celery copies the dependent modules of the script that it runs on the worker node. Say you run execution engine on the worker (`celery_run.delay(self.executor, data_id, script, verbosity)
 `). The engine that runs on the worker node spawns new processes. This would trigger a run of the manager on the worker node. The manager then resolves dependencies and runs the spawned processes. Celery somehow knows it has to copy the `managers/celery.py` to the worker node so it can be imported and executed. However you import the function in a method (like here) to avoid circular dependencies instead of on the top, this somehow messes up Celery. 

@dblenkus should know more about this and if we have a test for this in resolwe. Here is a test for demultiplexing: https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/tests/processes/test_demultiplex.py.

kostko:
All tests pass though?

resolwe/flow/models.py
kostko: All tests pass though?

resolwe/flow/managers/base.py
kostko: Done.

resolwe/flow/execution_engines/bash/__init__.py
kostko: Done.

resolwe/flow/models.py
tjanez: @kostko: As discussed, please find out how Celery sends tasks to workers and report the findings here.

resolwe/flow/expression_engines/jinja/__init__.py
tjanez: @kostko, ping

resolwe/flow/expression_engines/jinja/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    def evaluate_inline(self, expression, context=None):
+        """Evaluate an inline expression."""
+        if context is None:
+            context = {}
+
+        try:
+            compiled = self._environment.compile_expression(expression)
+            return compiled(**context)
+        except jinja2.TemplateError as error:
+            raise EvaluationError('{}'.format(error))
```
tjanez: @kostko, ping

resolwe/flow/expression_engines/jinja/__init__.py
kostko: Fixed.

resolwe/flow/models.py
tjanez: For future reference:

Everything in the call to [`celery_run()`](https://github.com/genialis/resolwe/blob/master/resolwe/flow/tasks.py#L14) must be picklable/serializable.

For `data_id`, `script` and `verbosity`, this is not an issue, since they are `int`, `str` and `str` (respectively). The `executor` argument is of type [`BaseFlowExecutor`](https://github.com/genialis/resolwe/blob/master/resolwe/flow/executors/__init__.py#L65) which is a more complex type and we must ensure that it remains serializable. Note that it is pickled "by reference" which means that Celery workers must run the same `BaseFlowExecutor` code as the manager process.

PR #139, opened 2016-11-03
Fix filters

dblenkus:


mstajdohar:
Add at least one test. For instance, test if filtering processes by slug indeed returns one process. Before this fix it returned all the processes.

mstajdohar:
Fix the linter error.

lukaw3d:
There is still an issue with `'genesis.base.filters.JsonOrderingFilter'`

kostko:
What kind of an issue?

lukaw3d:
This works: `api/presample?limit=5`
But ordering doesn't: `/api/presample?ordering=-created&limit=5`
Unless I remove `genesis.base.filters.JsonOrderingFilter`

kostko:
@lukaw3d does this problem occur specifically with this PR or even before it is applied? And does this only apply to the `created` field or also to others?

lukaw3d:
@kostko This PR specifically, but any ordering (not `created` specifically)

kostko:
Oh, ok, because I tested without this PR and the ordering works fine :-) I am not sure how the JSON filter could break anything, because it ignores all fields which don't have a dot (`.`) in their names? I'll take a look.

kostko:
BTW, what exactly does this PR fix?

lukaw3d:
@kostko OH no, I tested it wrong. It isn't this PR specifically

kostko:
Why didn't you say that this raises an exception on the new version? This makes the error apparent - there was a package upgrade which caused the API to change.

mstajdohar:
> BTW, what exactly does this PR fix?

When we upgraded djangorestframework-filters to 0.9.0 the filters stopped working. For instance, when you would try to filter samples by slug, the server would return all samples instead of just one.

kostko:
> When we upgraded djangorestframework-filters to 0.9.0 the filters stopped working.

And not only this. The upgrade also broke the internal filters API and this is why the ordering is not working.

tjanez:
So, we need at least two tests:
- test for filtering by slug
- test for ordered filtering

PR #149, opened 2016-11-11
Improve error message for exported files

dblenkus:


resolwe/flow/executors/__init__.py
```diff
@@ -56,8 +56,8 @@ def iterjson(text):
 def hydrate_spawned_files(filename, data_id):
     """Hydrate spawned files' paths."""
     if filename not in EXPORTED_FILES_MAPPER:
-        raise KeyError('All files referenced in spawned processes must be exported using'
-                       '`re-export` command.')
+        raise KeyError('All files referenced in spawned processes must be exported using '
```
dblenkus: `re-export` is not used in spawned process, but for spawned process, so I modified the message a bit.

resolwe/flow/executors/__init__.py
```diff
@@ -56,8 +56,7 @@ def iterjson(text):
 def hydrate_spawned_files(filename, data_id):
     """Hydrate spawned files' paths."""
     if filename not in EXPORTED_FILES_MAPPER:
-        raise KeyError('All files referenced in spawned processes must be exported using'
-                       '`re-export` command.')
+        raise KeyError('Use `re-export` to prepare following file for spawned process: {}'.format(filename))
```
mstajdohar: I would suggest:
```
'Use `re-export` to prepare the file for spawned process: {}'.format(filename)
```

PR #150, opened 2016-11-11
Add function for checking object permissions

dblenkus:


mstajdohar:
Rebase on to master and write tests.

resolwe/flow/views.py
```diff
@@ -456,20 +456,24 @@ class CollectionViewSet(ResolweCreateModelMixin,
     permission_classes = (permissions_cls,)
     filter_class = CollectionFilter
 
+    def _check_object_permissions(self, collection, user):
```
mstajdohar: Consider renaming to `_check_collection_permissions`.

dblenkus: Naming is intentional, because you have to check the permissions on Sample in SampleViewSet and you just overwrite this function.

PR #157, opened 2016-11-21
Add workflow execution engine

kostko:


resolwe/flow/execution_engines/workflow/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+        if expression_engine is not None:
+            expression_engine = self.get_expression_engine(expression_engine)
+
+        # Parse steps.
+        steps = data.process.run.get('program', None)
+        if steps is None:
+            return
+
+        if not isinstance(steps, list):
+            raise ExecutionError('Workflow program must be a list of steps.')
```
mstajdohar: Should we enforce this in process schema as well?

resolwe/flow/execution_engines/workflow/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+            # Process all input variables.
+            step_input = step.get('input', {})
+            if not isinstance(step_input, dict):
+                raise ExecutionError('Incorrect definition of step "{}", input must be a dictionary.'.format(
+                    step_id
+                ))
+
+            data_input = {}
+            for input_name, input_value in step_input.items():
+                if isinstance(input_value, string_types):
```
mstajdohar: Some input values can be in lower levels of the dict, _e.g._ in `alignment-star` part of the input is given as:
```
{
    'filtering': {
        'outFilterType': 'Normal',
        'outFilterMultimapNmax': 2,
        'outFilterMismatchNmax': 10
    }
    ...
} 
```

resolwe/flow/execution_engines/workflow/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+
+class ExecutionEngine(BaseExecutionEngine):
+    """An execution engine that supports workflow specifications."""
+
+    name = 'workflow'
+
+    def get_output_schema(self, process):
+        """Return any additional output schema for the process."""
+        return [
+            {'name': 'steps', 'type': 'list:data:'},
```
mstajdohar: Should this be of type `data:workflow`?

resolwe/flow/execution_engines/workflow/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+
+            # TODO: What about permissions?
+
+            context['steps'][step_id] = data_object.pk
+
+        # Immediately set our status to done and output all data object identifiers.
+        data.output = {
+            'steps': list(context['steps'].values()),
+        }
+        data.status = Data.STATUS_DONE
```
mstajdohar: Should the workflow fail if a step fails? And should it have the lowest status of the steps? For instance, if the first step is `Data.STATUS_DONE` and the second is `Data.STATUS_WAITING` the workflow object should have status `Data.STATUS_WAITING`. We'll need this when we'll support tracking workflow execution in the UI.

resolwe/flow/execution_engines/workflow/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+                data_input[input_name] = input_value
+
+            # Create the data object.
+            data_object = Data.objects.create(
+                process=process,
+                contributor=data.contributor,
+                input=data_input,
+            )
+
+            # TODO: What about permissions?
```
mstajdohar: @dblenkus ?

resolwe/flow/tests/processes/workflow.yml
```diff
@@ -0,0 +1,59 @@
...
+    program:
+      - id: step1
+        run: test-example-1
+        input:
+          param1: world
+          param2: true
+      - id: step2
+        run: test-example-2
+        input:
+          param1: '{{ steps.step1 }}'
```
mstajdohar: Consider inputs in lower levels of the dict.

resolwe/flow/execution_engines/workflow/__init__.py
kostko: No. Because these are the data objects that the steps can output and they can be of arbitrary types. I needed to change the type schema in order to allow definition of a generic data type that can match any data object.

resolwe/flow/execution_engines/workflow/__init__.py
kostko: Good point!

resolwe/flow/execution_engines/workflow/__init__.py
kostko: Permissions will be copied from the parent.

resolwe/flow/execution_engines/workflow/__init__.py
kostko: @dblenkus mentioned that we currently do the same for spawned processes? Is this correct? I think we should be consistent.

mstajdohar: Maybe, maybe not. In spawned processes the correct implementation would probably that the original process gets `STATUS_DONE` if at least one of the spawned processes gets `STATUS_DONE`. Consider this use case. The sequencer outputs a single file that includes reads in multiple lanes. We demultiplex this original file into several files, then we spawn Data objects, one for each of this files. Say one of the lanes had issues (maybe no reads) and the Data object gets `STATUS_ERROR`. All the other Data objects with `STATUS_DONE` are still useful in subsequent analysis. In workflow we have the reverse. If one step fails all the subsequent steps will fail, because they depend on it. So the workflow as a whole would fail.

kostko: But is it as you described with spawned processes? So the manager actually tracks the spawned processes and updates the status of the parent process? I am not sure this is the case currently.

mstajdohar: No, we do not do this yet. I was just thinking what would be a correct solution. Maybe what we have now is a correct solution for the spawned processes. If you consider spawned processes independent of the original process, the original process should get `STATUS_DONE` immediately after the processes are spawned successfully. The original process "does not care" what happens with its children. They may live or die, his job is done.

kostko: Ok, so in this case we need to add some kind of tracking of child processes to the manager and make it modular so different execution engines can do this in different ways.

resolwe/flow/execution_engines/workflow/__init__.py
kostko: The problem with the monolithic schema is that it makes everything non-modular as schema for all supported execution engines must be defined in one place.

Because the definition of the program depends on the execution engine and only the execution engine defines the schema. This is why I updated the schema to make the program be of any type (string, object, array).

resolwe/flow/execution_engines/workflow/__init__.py
mstajdohar: Conceptually, spawned processes start independent workflow that never merge. The opposite are concepts like workflow, branch and merge, map/reduce...

kostko: Yes, but in any case, we need support for tracking child processes. Perhaps in a similar way to how we already track dependencies between data objects (through inputs), but here the execution engines would explicitly declare some data objects as being children. What do you think?

resolwe/flow/execution_engines/workflow/__init__.py
mstajdohar: Hm, just thinking, what if we would introduce execution engine schemas? Your solution is fine though.

resolwe/flow/execution_engines/workflow/__init__.py
mstajdohar: I'll have to sleep on this.

mstajdohar: Maybe something like this. We add a model Link with fields parent, child, type (input/output, spawned, workflow). This would also support fast queries for inputs.

resolwe/flow/execution_engines/workflow/__init__.py
kostko: Yes, this would probably be the best thing. Then we would just need to merge the schemas based on the configured execution engine before performing validation.

resolwe/flow/managers/base.py
```diff
@@ -104,7 +104,8 @@ def communicate(self, run_sync=False, verbosity=1):
                         # program must not be set to None as then the process will be stuck in waiting state.
                         program = ''
 
-                    data.status = Data.STATUS_WAITING
+                    if data.status != Data.STATUS_DONE:
```
dblenkus: Add a comment with explanation for this check.

resolwe/flow/execution_engines/workflow/__init__.py
dblenkus: What if we add `get_status` method to the `Data` model. It can just return the status for normal objects and calculate it for workflows (and maybe cache `DONE` and `ERROR` in status field)? Workflow already has a list of it's children, so it shouldn't be a problem to calculate the status.
Only drawback is that `status` field won't be updated if it will be called directly.

resolwe/flow/managers/base.py
kostko: Done.

resolwe/flow/tests/processes/workflow.yml
kostko: Done.

resolwe/flow/execution_engines/workflow/__init__.py
kostko: Done.

PR #158, opened 2016-11-23
Order objects on API endpoints

dblenkus:


kostko:
Tests are failing?

mstajdohar:
> Tests are failing?

I think because of the problems with our JIRA/Jenkins server. Tests passed on Travis.

resolwe/flow/views.py
```diff
@@ -451,10 +452,16 @@ class CollectionViewSet(ResolweCreateModelMixin,
                         viewsets.GenericViewSet):
     """API view for :class:`Collection` objects."""
 
-    queryset = Collection.objects.all().prefetch_related('descriptor_schema', 'contributor')
+    queryset = Collection.objects.all().prefetch_related(
+        'descriptor_schema',
+        'contributor',
+        Prefetch('data', queryset=Data.objects.all().order_by('created'))
```
mstajdohar: Created is not OK, we should order by ID. Objects may be created within the same microsecond and in this cases the order among them will be random.

resolwe/flow/views.py
```diff
@@ -451,10 +452,16 @@ class CollectionViewSet(ResolweCreateModelMixin,
...
-    queryset = Collection.objects.all().prefetch_related('descriptor_schema', 'contributor')
+    queryset = Collection.objects.all().prefetch_related(
+        'descriptor_schema',
+        'contributor',
+        Prefetch('data', queryset=Data.objects.all().order_by('id'))
+    )
     serializer_class = CollectionSerializer
     permission_classes = (permissions_cls,)
     filter_class = CollectionFilter
+    ordering_fields = ('created', 'modified', 'name')
```
mstajdohar: Should we create db indexes on this fields?

resolwe/flow/views.py
```diff
@@ -521,6 +530,8 @@ class DataViewSet(ResolweCreateDataModelMixin,
     serializer_class = DataSerializer
     permission_classes = (permissions_cls,)
     filter_class = DataFilter
+    ordering_fields = ('created', 'modified', 'started', 'finished', 'name')
```
mstajdohar: Should we create indexes on this fields?

mstajdohar:
Fix tests.

dblenkus:
I've just remove one test, because if we order by id, there is no way to change order and it cannot be tested if it works.

mstajdohar:
> I've just remove one test, because if we order by id, there is no way to change order and it cannot be tested if it works.

:) 

You should test if the id field is of type auto incremental integer. If we change the type of IDs sometime in the future this sort will fail. But wait, then even your current test is valid.

mstajdohar:
Tests still fail.

PR #165, opened 2016-11-29
Add Entity

dblenkus:


resolwe/flow/models.py
```diff
@@ -691,7 +736,36 @@ class Meta(BaseModel.Meta):
...
+            ("view_entity", "Can view entity"),
+            ("edit_entity", "Can edit entity"),
+            ("share_entity", "Can share entity"),
+            ("download_entity", "Can download files from entity"),
+            ("add_entity", "Can add data objects to entity"),
+            ("owner_entity", "Is owner of the entity"),
+        )
+
+    #: list of collections to which entity belongs
+    collections = models.ManyToManyField(Collection)
```
mstajdohar: I thought Entity would contain data objects, like Sample does now.

resolwe/flow/views.py
```diff
@@ -500,6 +501,97 @@ def remove_data(self, request, pk=None):
...
 
 
+class EntityViewSet(CollectionViewSet):
+    """API view for entities."""
+
+    filter_class = EntityFilter
+    serializer_class = EntitySerializer
+
+    queryset = Entity.objects.annotate(
+        latest_date=Max('data__modified')
```
mstajdohar: Add index on the `modified` field.

resolwe/flow/views.py
```diff
@@ -500,6 +501,97 @@ def remove_data(self, request, pk=None):
...
+    ).prefetch_related(
+        'descriptor_schema',
+        'contributor'
+    ).order_by('-latest_date')
```
mstajdohar: Maybe this will be too slow and we should have a _latest_date_ field that we update through signals each time a data object changes.

resolwe/flow/models.py
```diff
@@ -483,6 +484,42 @@ def add_dependency(value):
                 for data in value:
                     add_dependency(data)
 
+    def create_entity(self):
+        """Create entity if `flow_collection` is defined in process.
+
+        Following rules applies for adding `Data` object to `Entity`:
+        * Only add `Data object` to `Entity` if process has defined
+        `flow_collwection` field
+        * Add object to existing `Entity`, if all perents that are part
```
kostko: `perents` -> `parents`

resolwe/flow/models.py
```diff
@@ -535,7 +572,23 @@ def save(self, render_name=False, *args, **kwargs):
...
             # is why a transaction block is needed and the save method must be called first.
-            self.save_dependencies(self.input, self.process.input_schema)  # pylint: disable=no-member
+            if create:
+                self.save_dependencies(self.input, self.process.input_schema)  # pylint: disable=no-member
+
+        if create:
+            self.create_entity()
+
+    def delete(self, *args, **kwargs):
+        """Delete Data object and its Entity if it becomes empty."""
```
kostko: I am not sure if it is ok to do this by overriding `delete`. In bulk deletes, it can happen that this method is not called. Use the `post_delete` signal instead, which is called even for bulk deletes.

resolwe/flow/models.py
kostko: It does contain data objects as it is a collection.

resolwe/flow/views.py
dblenkus: Same query was quick enough in sample view. I've added prefetch for `data` objects and db index for `modified` field, so now it should be even quicker. I think that we can optimize this when problem will appear.

dblenkus:
Fixed

mstajdohar:
@kostko please check if you comments were addressed and allow merge.

PR #167, opened 2016-11-30
Add option to order processes by version

### LINKED ISSUE #166: Ordering of processes does not work
Tested on the QA server:

https://qa.genialis.com/api/process?slug=alignment-bwa-mem&ordering=-version

For the above query, the `alignment-bwa-mem` processes should be sorted by version, descending. The version 16777232 should be at the top, but comes second.

dblenkus:
Fixes #166

### DIFF
--- a/resolwe/flow/tests/test_ordering.py
+++ b/resolwe/flow/tests/test_ordering.py
@@ -1,0 +1,35 @@
+# pylint: disable=missing-docstring
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from django.contrib.auth import get_user_model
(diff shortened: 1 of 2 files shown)


mstajdohar:
Test would be nice. This used to work but was broken during refactoring recently.

dblenkus:
But test wouldn't catch this problem, because there would be no test for option that we forgot to add. This is just simple list that is passed to Viewset and there are tests in django rest framework that ensures that ordering works. I think that such test would only duplicate them and add unneeded   complexity to our test.

dblenkus:
Added test

PR #169, opened 2016-12-05
Workflow improvements

kostko:
* Add missing workflow input support.
* Add `data_by_slug` filter.

### DIFF
--- a/resolwe/flow/tests/test_expression_filters.py
+++ b/resolwe/flow/tests/test_expression_filters.py
@@ -50,6 +50,7 @@ class ProcessFieldsTagsTest(TestCase):
                 {'name': 'basename', 'type': 'basic:string:'},
                 {'name': 'subtype', 'type': 'basic:string:'},
                 {'name': 'yesno', 'type': 'basic:string:'},
+                {'name': 'datalookup', 'type': 'basic:integer:'},
             ],
             run={
                 'language': 'bash',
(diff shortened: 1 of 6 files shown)


resolwe/flow/tests/test_expression_filters.py
```diff
@@ -60,6 +61,7 @@ def test_templatetags(self):
 re-save basename "{{ '/foo/bar/moo' | basename }}"
 re-save subtype "{{ 'data:test:inputobject:' | subtype('data:') }}"
 re-save yesno "{{ true | yesno('yes', 'no') }}"
+re-save datalookup "{{ 'input-data-object' | lookup_data_by_slug }}"
```
mstajdohar: How about just `by_slug`?

mstajdohar: > But we can remove the `lookup_` prefix.

kostko: Done.

dblenkus:
@mstajdohar Where do we need this? 
We should be careful with such features, because slug is not always exactly what you set - suffix can be added to it to make it unique.

kostko:
> @mstajdohar Where do we need this?

@jkokosar needs this when constructing workflows to reference existing data objects. I think this was also part of the specification.

mstajdohar:
One example is for automating the primary analysis. Janez will prepare certain data objects on the platform (genome, gene model, ontology...) and then in the configuration specify them as inputs to the pipeline.

Another example is that we wish to make the GO Enrichment Analysis process easy to use. Based on the inputs it can figure out which annotation files it should use (human or mouse). The process would expect either human-gaf or mouse-gaf data object on the platform. If they would not be available, it would fail.

Specifically, when you run the GO Enrichment Analysis all that is really needed on input are gene IDs. We can then figure out automatically if the IDs are mouse or human. Based on this we can find the appropriate GAF on the platform automatically, so there is no need for the user to specify it.

kostko:
Can we merge this?

mstajdohar:
I agree. @dblenkus?

kostko:
I've now renamed `inputs` to `input`.

PR #177, opened 2016-12-15
WIP: Add elastic app

dblenkus:


resolwe/elastic/indices.py
```diff
@@ -0,0 +1,122 @@
...
+import six
+
+import elasticsearch_dsl as dsl
+
+from guardian.models import GroupObjectPermission, UserObjectPermission
+
+from resolwe.flow.models import dict_dot
+
+
+class NoValue(object):
```
kostko: Please explain that this is used to differentiate missing attributes from `None` (another approach would be to catch `AttributeError` instead of having a custom guard type).

resolwe/elastic/indices.py
```diff
@@ -0,0 +1,122 @@
...
+    obj = None
+    object_type = None
+    mapping = {}
+    document_class = None
+    _preprocessed = False
+
+    def __init__(self, obj):
+        self.obj = obj
+
+    def filter(self):
```
kostko: Add documentation for all methods.

resolwe/elastic/indices.py
```diff
@@ -0,0 +1,122 @@
...
+                continue  # this fields are handled separately
+
+            get_value_function = getattr(self, 'get_{}_value'.format(field), None)
+            if get_value_function:
+                setattr(document, field, get_value_function())
+                continue
+
+            if field in self.mapping:
+                if callable(self.mapping[field]):
+                    setattr(document, field, self.mapping[field]())
```
kostko: I think it would be better if the callable is called with the object (`self.obj`) as argument.

resolwe/elastic/indices.py
```diff
@@ -0,0 +1,122 @@
...
+                    continue
+
+                if isinstance(self.mapping[field], six.string_types):
+                    object_attr = dict_dot(self.obj, self.mapping[field], default=self.mapping[field])
```
kostko: I don't think this is a good approach as typos will automatically change to default values (e.g. saying `fo.bar` instead of `foo.bar` will silently set all indexed values for that field to the string `'fo.bar'`). In order to set constants we should force them to be wrapped in lambda expressions or figure out a different way.

resolwe/elastic/indices.py
```diff
@@ -0,0 +1,122 @@
...
+                else:
+                    object_attr = self.mapping[field]
+                if callable(object_attr):
+                    setattr(document, field, object_attr())
```
kostko: As above, pass `self.obj` as argument.

resolwe/elastic/indices.py
```diff
@@ -0,0 +1,122 @@
...
+                else:
+                    setattr(document, field, object_attr)
+                continue
+
+            object_value = dict_dot(self.obj, field, default=NoValue())
```
kostko: Why using `NoValue` instead of propagating `AttributeError`/`KeyError`?

resolwe/elastic/indices.py
```diff
@@ -0,0 +1,122 @@
...
+            setattr(document, field, object_value)
+
+        permissions = self.get_permissions()
+        document.users_with_permissions = permissions['users']
+        document.groups_with_permissions = permissions['groups']
+
+        document.save()
+
+    def get_permissions(self):
+        return {
```
kostko: How expensive is this when doing indexing in bulk? Can we optimize during bulk reindex?

resolwe/elastic/management/commands/es_index.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class Command(BaseCommand):
+    """Build ElasticSearch indexes for data objects."""
+
+    help = "Build ElasticSearch indexes for data objects."
+
+    def handle(self, *args, **options):
+        """Command handle."""
+        for data in Data.objects.all():
```
kostko: Can we avoid hardcoding which models to index? Instead, the list of registered index classes should be used to determine the set of models to index.

resolwe/elastic/signals.py
```diff
@@ -0,0 +1,76 @@
...
+        return
+
+    if not perm.content_type.name in ['data', 'sample']:
+        return
+
+    obj = perm.content_type.get_object_for_this_type(id=perm.object_pk)
+    add_index(obj)
+
+
+@receiver(post_save, sender=Data)
```
kostko: I don't like these hardcoded signals. Can we make it so that the first time a new index references a new model, the appropriate signals are connected?

resolwe/elastic/indices.py
```diff
@@ -0,0 +1,122 @@
...
+    _preprocessed = False
+
+    def __init__(self, obj):
+        self.obj = obj
+
+    def filter(self):
+        pass
+
+    def preprocess_object(self):
+        pass
```
mstajdohar: Raise NotImplementedError exception.

resolwe/elastic/management/commands/es_index.py
```diff
@@ -0,0 +1,27 @@
+""".. Ignore pydocstyle D400.
```
mstajdohar: Consider renaming to `elastic_index.py`.

resolwe/elastic/viewsets.py
```diff
@@ -0,0 +1,58 @@
...
+from elasticsearch_dsl.query import Q
+
+from django.conf import settings
+from django.contrib.auth import get_user_model
+
+from rest_framework.viewsets import GenericViewSet
+
+
+class ElasticSearchViewSet(GenericViewSet):
+    """Django REST Framework viewset
```
mstajdohar: One line comment.

dblenkus: Yes, I will improve documentation :)

resolwe/elastic/indices.py
dblenkus: At the moment each object is processed individually, so I don't see easy option to optimise this. But this is definitely something that we have to work on in the future. I will add a comment about this.

resolwe/elastic/management/commands/es_index.py
dblenkus: Ok, I will finish everything else and then work on this.

resolwe/elastic/indices.py
kostko: BTW, is there a reason why a constant value should be indexed? Wouldn't it be the same for all documents?

resolwe/elastic/indices.py
mstajdohar: Then change the docs accordingly,
> To work properly, subclass of this class must ovverride following methods and attributes: ... object_type and process_object

resolwe/elastic/builder.py
```diff
@@ -10,33 +10,99 @@
...
+
+    """
+
+    #: list of index builders
+    indexes = []
+
+    def __init__(self):
+        """Initialize index builder object."""
+        # Set dafault connection for ElasticSearch
+        # TODO: Reestablish connection if broken!
```
kostko: Is this not handled automatically (e.g. retry API request)? Because AFAIK connections are not persistent, each request is done via HTTP?

dblenkus: It should be, but the connection stopped working when ElasticSearch was restarted.

resolwe/elastic/builder.py
```diff
@@ -0,0 +1,108 @@
...
+            indexes_path = '{}.elastic_indexes'.format(app_config.name)
+            try:
+                indexes_module = import_module(indexes_path)
+
+                for attr_name in dir(indexes_module):
+                    attr = getattr(indexes_module, attr_name)
+                    if inspect.isclass(attr) and issubclass(attr, BaseIndex) and attr is not BaseIndex:
+                        self.indexes.append(attr())
+            except ImportError:
+                pass  # no `elastic_indexes` in app
```
kostko: Can we differentiate this from "importing that specific index caused an ImportError"? Because this one should be reraised.

resolwe/elastic/builder.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def trigger(self, obj=None):
```
kostko: I am not sure if `trigger` is the best name for this, perhaps it should be `build`, `index` or `update_object` (to be consistent with `remove_object` below).

resolwe/elastic/indices.py
```diff
@@ -0,0 +1,200 @@
...
+class BaseDocument(dsl.DocType):
+    """Base document class to build ElasticSearch documents.
+
+    This is standard ``elasticsearch-dsl`` ``DocType`` class with
+    already added fields for handling permissions.
+
+    """
+
+    #: list of user ids with view permission on the object
+    users_with_permissions = dsl.String()
```
kostko: Shouldn't there be a `multi=True` here as it accepts an array of strings (also below)?

resolwe/elastic/indices.py
```diff
@@ -0,0 +1,200 @@
...
+
+        permissions = self.get_permissions(obj)
+        document.users_with_permissions = permissions['users']
+        document.groups_with_permissions = permissions['groups']
+
+        document.save()
+
+    def run(self, obj=None):
+        """Main function for running indexes."""
+        if obj and obj not in self.queryset:
```
kostko: Does this execute a database query just to determine if an object matches the filter? Can this be improved?

resolwe/elastic/viewsets.py
```diff
@@ -0,0 +1,69 @@
...
+from django.conf import settings
+from django.contrib.auth import get_user_model
+
+from rest_framework.viewsets import GenericViewSet
+
+
+__all__ = ('ElasticSearchViewSet',)
+
+
+class ElasticSearchViewSet(GenericViewSet):
```
kostko: Should this viewset also include methods for handling most common search requests, so that the user needs to only override a few specific things and common stuff like ordering and pagination are handled correctly? An even better approach would be to maybe have this as a mixin.

resolwe/elastic/builder.py
dblenkus: Yes, urllib3 raised error that connection is broken.

resolwe/elastic/builder.py
dblenkus: Done.

resolwe/elastic/builder.py
dblenkus: I'll name it `build`, because it triggers whole build process and `remove_object` only removes single object, so there should be some difference in the name.

resolwe/elastic/indices.py
dblenkus: Yes, database query is executed, but I don't see any easier way to implement this. This is used only for single object (if `obj` is defined), so it shouldn't slow the process too much.

resolwe/elastic/viewsets.py
dblenkus: Agree, I've changed it to the mixin.

resolwe/elastic/indices.py
kostko: Ok, if it is just for a single object (and never used in loops with a large number of objects) then it is ok.

resolwe/elastic/indices.py
```diff
@@ -0,0 +1,209 @@
...
+class BaseDocument(dsl.DocType):
+    """Base document class to build ElasticSearch documents.
+
+    This is standard ``elasticsearch-dsl`` ``DocType`` class with
+    already added fields for handling permissions.
+
+    """
+
+    #: list of user ids with view permission on the object
+    users_with_permissions = dsl.String(many=True)
```
kostko: Shouldn't it be `multi` and not `many`? At least based on [this](https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/field.py#L49).

resolwe/elastic/indices.py
```diff
@@ -0,0 +1,209 @@
...
+        class_name = type(self).__name__
+        if not self.object_type:
+            raise RuntimeError('`object_type` must be defined in {}'.format(class_name))
+
+        if not self.document_class:
+            raise RuntimeError('`document_class` must be defined in {}'.format(class_name))
+
+        self._index_name = self.document_class()._get_index()  # pylint: disable=not-callable,protected-access
+
+    def filter(self):
```
kostko: In which case is `filter` used, now that we have `queryset`?

resolwe/elastic/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from elasticsearch_dsl.connections import connections
+
+from django.conf import settings
+
+elasticsearch_host = getattr(settings, 'ELASTICSEARCH_HOST', 'localhost')  # pylint: disable=invalid-name
+elasticsearch_port = getattr(settings, 'ELASTICSEARCH_PORT', 9200)  # pylint: disable=invalid-name
+connections.create_connection(hosts=['{}:{}'.format(elasticsearch_host, elasticsearch_port)])
```
kostko: Any reason why this must be done here and not in the `ready` method?

resolwe/elastic/builder.py
```diff
@@ -0,0 +1,139 @@
...
+    #: list of index builders
+    indexes = []
+
+    #: list of registered signals
+    signals = []
+
+    def __init__(self):
+        """Initialize index builder object."""
+        # Set dafault connection for ElasticSearch
+        # elasticsearch_host = getattr(settings, 'ELASTICSEARCH_HOST', 'localhost')
```
kostko: What's up with this commented block?

resolwe/elastic/builder.py
```diff
@@ -0,0 +1,139 @@
...
+        # elasticsearch_port = getattr(settings, 'ELASTICSEARCH_PORT', 9200)
+        # # TODO: Reestablish connection if broken!
```
kostko: So currently connections are not reestablished? Would it be hard to add this?

resolwe/elastic/indices.py
```diff
@@ -0,0 +1,236 @@
...
+        object_type = type(obj).__name__.lower()
+        return '{}_{}'.format(object_type, self.get_object_id(obj))
+
+    def process_object(self, obj):
+        """Process current object and push it to the ElasticSearch."""
+        document = self.document_class(meta={'id': self.generate_id(obj)})  # pylint: disable=not-callable
+
+        for field in document._doc_type.mapping:  # pylint: disable=protected-access
+            if field in ['users_with_permissions', 'groups_with_permissions']:
+                continue  # this fields are handled separately
```
kostko: `this fields` -> `These fields`

kostko:
BTW, this PR seems to decrease coverage?

resolwe/elastic/__init__.py
dblenkus: Thanks for spotting this, Ive forgot to revert the changes when performing some tests.

resolwe/elastic/builder.py
dblenkus: Same as above.

resolwe/elastic/builder.py
dblenkus: Looks like it has started to work by itself :)

dblenkus:
@kostko I've added a commit with fixes.

I will add tests in separate pull request, so we can start testing this.

PR #181, opened 2016-12-22
Export RESOLWE_API_HOST env var in executor

mstajdohar:
TODO: localhost will not work for development, because the server does not run within docker.

### DIFF
--- a/resolwe/flow/managers/base.py
+++ b/resolwe/flow/managers/base.py
@@ -8,6 +8,7 @@ Abstract Manager
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import logging
+import os
 
 from django.conf import settings
 from django.db import IntegrityError, transaction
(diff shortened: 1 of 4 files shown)


kostko:
> localhost will not work for development, because the server does not run within docker.

This can be fixed by using the IP of the docker bridge interface in development mode instead of `localhost`.

resolwe/flow/managers/base.py
```diff
@@ -65,6 +66,15 @@ def __init__(self):
         execution_engines = getattr(settings, 'FLOW_EXECUTION_ENGINES', ['resolwe.flow.execution_engines.bash'])
         self.execution_engines = self.load_execution_engines(execution_engines)
 
+    def _include_environment_variables(self, program):
+        """Define environment variables."""
+        env_vars = {
+            'RESOLWE_API_HOST': getattr(settings, 'RESOLWE_API_HOST', 'localhost'),
+        }
+
+        export_commends = ['export {}={}'.format(key, value) for key, value in env_vars.items()]
```
dblenkus: Should this be escaped and wrapped in quotes?

tests/settings.py
```diff
@@ -10,6 +10,8 @@
 
 SECRET_KEY = 'secret'
 
+RESOLWE_API_HOST = 'localhost'
```
mstajdohar: > This can be fixed by using the IP of the docker bridge interface in development settings instead of localhost.

I guess here. Is there a quick way to read the IP of the docker bridge interface?

kostko: Not exactly "quick", but here is one example using the `ioctl` syscall:
```python
import socket
import fcntl
import struct

SIOCGIFADDR = 0x8915

sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
docker0_address = socket.inet_ntoa(fcntl.ioctl(
    sock.fileno(),
    SIOCGIFADDR,
    struct.pack('256s', 'docker0'),
)[20:24])

kostko: Not sure if it works on non-Linux.

mstajdohar: I think @tjanez has some other ideas to support this for tests. He'll write more about it.

mstajdohar:
We will solve this in separate PR.

PR #183, opened 2017-01-04
Trim auto generated data names

### LINKED ISSUE #179: Process crashes if rendered data name is too long
After process is done, names of all child objects are re-rendered. If any of the names is longer than 100 characters (limit specified in name field), process crashes.

Names of data object should probably be trimmed to that limit. Increasing of the limit should also be considered (in combination with trimming).

dblenkus:
Closes #179

### DIFF
--- a/resolwe/flow/models.py
+++ b/resolwe/flow/models.py
@@ -595,13 +595,19 @@ class Data(BaseModel):
         template_context = inputs
 
         try:
-            self.name = render_template(
+            name = render_template(
                 self.process,
                 self.process.data_name,  # pylint: disable=no-member
                 template_context
             )
         except EvaluationError:
-            self.name = '?'
+            name = '?'
+
+        name_max_len = self._meta.get_field('name').max_length
+        if len(name) > name_max_len:
+            name = name[:(name_max_len - 3)] + '...'
+
+        self.name = name
 
 
 class DescriptorSchema(BaseModel):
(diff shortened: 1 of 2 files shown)


resolwe/flow/models.py
```diff
@@ -595,13 +595,19 @@ def _render_name(self):
...
+            name = render_template(
                 self.process,
                 self.process.data_name,  # pylint: disable=no-member
                 template_context
             )
         except EvaluationError:
-            self.name = '?'
+            name = '?'
+
+        name_max_len = self._meta.get_field("name").max_length
```
kostko: Single quotes.

resolwe/flow/tests/test_models.py
```diff
@@ -127,6 +127,13 @@ def test_name(self):
         self.assertEqual(second.name, 'User\' data name')
         self.assertTrue(second.named_by_user)
 
+    def test_trim_name(self):
+        process = Process.objects.create(contributor=self.user, data_name='test' * 50)
```
mstajdohar: Should we make the test invariant of the field length:
```
name_max_len = Data._meta.get_field("name").max_length
data_name='a' * (name_max_len + 1) 
process = Process.objects.create(contributor=self.user, data_name= data_name)
```

resolwe/flow/models.py
dblenkus: Fixed.

resolwe/flow/tests/test_models.py
dblenkus: Tests shouldn't include too much logic, so I think that it is not necessary to complicate it. If the limit will be changed, `assert(len, 100)` will fail and test will have to be fixed.

PR #187, opened 2017-01-09
Preparatory work for adding a custom LiveServerTestCase for process tests that use Resolwe API

tjanez:
Contains:
- Upgrade to Sphinx 1.5.1+
- Remove `RESOLWE_API_HOST` setting from test project's settings since it is not really used anywhere
- Expose `check_docker()` test utility function 
- Add `check_installed()` test utility function
- Add support for configuring the network mode of Docker executor

I think it will be easier to review this commit by commit.

PR #192, opened 2017-01-16
Add support for Docker image requirements

kostko:


resolwe/flow/executors/__init__.py
```diff
@@ -122,6 +123,14 @@ def run(self, data_id, script, verbosity=1):
 
         self.data_id = data_id
 
+        # Fetch data instance to get any executor requirements.
+        if data_id != 'no_data_id':
```
kostko: Yes, this is used during some tests, which do a lot of mocking. I don't like it, but without this, these other tests will fail.

mstajdohar: I also dislike this. We should not complicate production code, to support one weird test case. How about you fix [this test](https://github.com/genialis/resolwe/blob/master/resolwe/flow/tests/test_executors.py#L81) to refer to an actual data object?

kostko: I am not sure why this test is done like it is. I think it would be better to just use the `ProcessTestCase` which uses actual data objects (something like I do in the new `test_executor_requirements`). @dblenkus?

kostko: I would wait for #187 to be merged as it introduces some refactoring around the Docker test utilities.

dblenkus: I don't remember exactly, but I think that this hack was introduced to support "get or create", where we don't want to process new objects in standard way. However I don't now if this is still used.

tjanez: > I think it would be better to just use the `ProcessTestCase` which uses actual data objects (something like I do in the new `test_executor_requirements`).

I'm stumbled across a similar thing recently. At first sight, `ProcessTestCase` seems a bit heavy-weight. Was this the reason not to use it?

tjanez:
@kostko, as discussed privately, I've uploaded Docker images needed by Resolwe's Docker executor tests to [Docker Hub](https://hub.docker.com/r/resolwe/test/).

Please, also replace current uses of `centos` Docker image with `resolwe/test:base` and make it the default in settings.

resolwe/flow/executors/__init__.py
kostko: I have created a `with_docker_executor` decorator, which handles setting up the Docker executor for a specific unit test. It is now used both for testing if Docker executor works and for testing if specifying Docker requirements works.

kostko:
Can you do another quick pass due to test updates?

mstajdohar:
Looks good.

resolwe/flow/utils/test.py
```diff
@@ -79,6 +80,31 @@ def check_docker():
     return available, reason
 
 
+def with_docker_executor(method):
```
tjanez: Very nice!

resolwe/flow/tests/test_executors.py
```diff
@@ -151,3 +110,13 @@ def test_workflow(self):
         self.assertEqual(step2_data.input['param2']['a'], step1_data.pk)
         self.assertEqual(step2_data.input['param2']['b'], 'hello')
         self.assertEqual(step2_data.output['out1'], 'simon says: hello world')
+
+    @with_docker_executor
```
tjanez: Is there any particular reason to put these two tests in the same class as above?

I would prefer to put them in a separate class, or rename existing class to something more general.

resolwe/flow/executors/__init__.py
tjanez: Great!

kostko:
> Could you put adding with_docker_executor() and test-docker process, and refactoring of the DockerExecutorTestCase in a separate commit (also note the addition of with_docker_executor() in the change log)?
This way adding Docker executor requirements will be clearer.

Done.

> Do you plan to implement local executor requirements in a separate commit?

Yes.

resolwe/flow/tests/test_executors.py
tjanez: Please, disregard, I've been deceived by GitHub's file trimming 😠 .

PR #194, opened 2017-01-18
Ignore objects of incorrect type during index

kostko:


mstajdohar:
A lint error:
```
************* Module resolwe.elastic.indices
resolwe/elastic/indices.py:211: [W0212(protected-access), BaseIndex.build] Access to a protected member _meta of a client class
```

kostko:
Dumb linter, I'll fix this.

PR #196, opened 2017-01-19
Increase size parameter in elastic viewset

### LINKED ISSUE #195: Elasticserach returns paginated results when querying/mapping features using RESDK
In resolwe-bio tools/goea.py `org_features = res.feature.filter(source=args.source_db, query=genes)` should return all genes, not just the first 10.

dblenkus:
Closes #195

### DIFF
--- a/resolwe/elastic/viewsets.py
+++ b/resolwe/elastic/viewsets.py
@@ -15,6 +15,7 @@ from elasticsearch_dsl.query import Q
 from django.conf import settings
 from django.contrib.auth import get_user_model
 
+from rest_framework.exceptions import APIException
 from rest_framework.response import Response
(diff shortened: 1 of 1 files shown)


resolwe/elastic/viewsets.py
```diff
@@ -155,6 +155,8 @@ def list_with_post(self, request):
         search = self.order_search(search)
         search = self.filter_permissions(search)
 
+        search = search.extra(size=10000)  # return 10000 results (default maximum limit by ES)
```
mstajdohar: @jkokosar is this ok for your use case?

dblenkus:
Added exception if query returns too many results.

PR #197, opened 2017-01-20
Add pagination support for POST requests

dblenkus:


resolwe/elastic/pagination.py
```diff
@@ -0,0 +1,58 @@
...
+
+"""
+
+
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from rest_framework.pagination import LimitOffsetPagination, _positive_int
+
+
+def get_query_param(request, key):
```
kostko: Is this the same method as in the base viewset? Can we have this only in one place?

dblenkus: This function raises error if key is not found and other one returns `None` (or default value). 

If error is not raised, I would have to handle it in a different way than DRF and there is much more space for errors.

kostko: Ok.

PR #200, opened 2017-01-24
Update versions

mstajdohar:
Do not merge until we write a PR for resolwe-bio that works with this PR.


resolwe/flow/models.py
```diff
@@ -101,7 +100,7 @@ class Meta:
     slug = AutoSlugField(populate_from='name', unique_with='version', editable=True, max_length=100)
 
     #: process version
-    version = VersionField(number_bits=VERSION_NUMBER_BITS, default=0)
+    version = VersionField(number_bits=VERSION_NUMBER_BITS, default="0.0.0")
```
kostko: Single quotes.

resolwe/flow/tests/processes/spawned.yml
```diff
@@ -46,5 +46,5 @@
     program: |
       # TODO: use re-import when added to resolwe-runtime-utils
       # re-import {{ input_file.file_temp }} {{ input_file.file }} "bar" "bar" 0.9
-      mv {{ input_file.file_temp }} {{ input_file.file }}
-      re-save-file saved_file {{ input_file.file }}
+      mv "{{ input_file.file_temp }}" "{{ input_file.file }}"
```
kostko: What does this change have to do with Django version bump? This is probably something that should be moved into a separate commit?

resolwe/flow/tests/test_api.py
```diff
@@ -36,14 +36,14 @@ def setUp(self):
...
-    @mock.patch('resolwe.flow.models.Process.objects.all')
-    def test_prefetch(self, process_mock):
+    def test_prefetch(self):
         request = factory.get('/', '', format='json')
         force_authenticate(request, self.user)
-        self.data_viewset(request)
 
         # check that only one request is made to get all processes' types
-        self.assertEqual(process_mock.call_count, 1)
+        # with prefetch 16 queries, without prefetch 17 queries
```
kostko: I think you should explain in the comment that the number of queries without prefetch depends on the number of Data objects. Just so we know when we add another Data object in the future.

resolwe/flow/utils/test.py
```diff
@@ -224,6 +224,8 @@ def tearDown(self):
                 shutil.rmtree(data_dir, ignore_errors=True)
                 shutil.rmtree(export_dir, ignore_errors=True)
 
+        Process.objects.all().delete()
```
kostko: Why is this needed? Aren't all unit tests executed in transactions?

resolwe/flow/tests/test_api.py
kostko: BTW, in theory it could be less queries, it just shouldn't be more. Is it possible to ensure that no more than 16 queries are executed?

resolwe/flow/utils/test.py
mstajdohar: This is interesting. All but the first test in `resolwe.flow.tests.test_executors.ManagerRunProcessTest` have failed after the update to Django 1.10 (see stacktrace below). I thought maybe processes were not removed because we register them by calling a Django command. 

```
ERROR: test_workflow (resolwe.flow.tests.test_executors.ManagerRunProcessTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/miha/.virtualenvs/resolwe/lib/python2.7/site-packages/django/test/testcases.py", line 216, in __call__
    self._post_teardown()
  File "/Users/miha/.virtualenvs/resolwe/lib/python2.7/site-packages/django/test/testcases.py", line 908, in _post_teardown
    self._fixture_teardown()
  File "/Users/miha/.virtualenvs/resolwe/lib/python2.7/site-packages/django/test/testcases.py", line 1064, in _fixture_teardown
    connections[db_name].check_constraints()
  File "/Users/miha/.virtualenvs/resolwe/lib/python2.7/site-packages/django/db/backends/postgresql/base.py", line 224, in check_constraints
    self.cursor().execute('SET CONSTRAINTS ALL IMMEDIATE')
  File "/Users/miha/.virtualenvs/resolwe/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/Users/miha/.virtualenvs/resolwe/lib/python2.7/site-packages/django/db/utils.py", line 94, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/Users/miha/.virtualenvs/resolwe/lib/python2.7/site-packages/django/db/backends/utils.py", line 62, in execute
    return self.cursor.execute(sql)
IntegrityError: insert or update on table "flow_process" violates foreign key constraint "flow_process_contributor_id_ae49ec6b_fk_auth_user_id"
DETAIL:  Key (contributor_id)=(30) is not present in table "auth_user".
```

resolwe/flow/tests/processes/spawned.yml
mstajdohar: Moved to a separate commit.

resolwe/flow/utils/test.py
kostko: This is indeed strange. Where did this contributor with id `30` come from? Is this the value of `self.admin.pk` in ProcessTestCase?

mstajdohar: What I still do not know is why it worked before the Django update.

mstajdohar: For each test we created 2 new users in the database:
- https://github.com/genialis/resolwe/blob/master/resolwe/flow/utils/test.py#L197
- https://github.com/genialis/resolwe/blob/master/resolwe/flow/tests/test_executors.py#L57

I guess the auto ID field was not reset after each test.

resolwe/flow/tests/processes/spawned.yml
dblenkus: Add quotes to the command in the comment.

resolwe/flow/utils/test.py
dblenkus: Problem is that `_register_schemas` caches processes (with contributor relation), but admin id is increased in each test run, so it cannot link them. Maybe this is new in Django 1.10.

I've also noted that super-user is created in `setUp` and in `_register_schemas` (only if it doesn't exist), so you can probably remove the latest.

dblenkus: Look like that manual delete works, but Django follows a bit different procedure in `tearDown` which crashes because of this anomaly.

mstajdohar: I suggest you improve the test logic in a separate PR.

mstajdohar: Fixed by updating a contributor in cached processes and schemas.

docs/CHANGELOG.rst
```diff
@@ -47,6 +49,8 @@ Fixed
 - Improve handling of validation errors
 - Ignore file size in assertFields
 - Order data objects in ``CollectionViewSet``
+- Fix tests for Django 1.10
+- Add quotes to a test paths
```
dblenkus: Make it a bit more descriptive, maybe: Add quotes to file paths in test processes

resolwe/flow/tests/test_api.py
```diff
@@ -31,19 +33,21 @@ def setUp(self):
...
         force_authenticate(request, self.user)
-        self.data_viewset(request)
 
-        # check that only one request is made to get all processes' types
-        self.assertEqual(process_mock.call_count, 1)
+        for _ in range(10):
+            Data.objects.create(contributor=self.user, process=self.proc)
+
+        # Check prefetch. The number of queries without prefetch depends on the
+        # number of Data objects. With prefetch 56 queries, without prefetch 73 queries
```
dblenkus: `56` -> `65`

resolwe/flow/utils/test.py
```diff
@@ -157,6 +157,18 @@ class ProcessTestCase(TestCase):
 
     """
 
+    def _update_schema_relations(self, schemas):
+        """Update foreign keys on process and descriptor schema.
+
+        The field contributor is updated.
+
+        """
+        if len(schemas) == 0:
```
dblenkus: Do you need this check?

PR #204, opened 2017-01-26
Prepare release 1.4.0

tjanez:


docs/conf.py
```diff
@@ -67,7 +67,7 @@
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = ['_build', 'CHANGELOG.rst']
```
mstajdohar: :)

tjanez: I have a few more grey hairs because of this :)

PR #205, opened 2017-01-26
Prepare release 1.4.1

tjanez:
Fix installation instructions for installing Resolwe from PyPI.

NOTE: Resolwe Bioinformatics was also updated for this in genialis/resolwe-bio#200 PR.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -6,9 +6,9 @@ All notable changes to this project are documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_.
 
 
-==========
-Unreleased
-==========
+==================
+1.4.1 - 2017-01-27
+==================
 
 Fixed
 -----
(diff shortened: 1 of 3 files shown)


tjanez:
Another alternative to the `--no-binary resolwe` "dance" would be to only make sdist releases (i.e. no wheels) while Resolwe has dependency links.

That would mean we could radically simplify [this commit in Resolwe Bioinformatics](https://github.com/genialis/resolwe-bio/pull/200/commits/d20987cfe31765ad0cad1aa1bf030c46ee8c27f3).

@mstajdohar, @dblenkus, what do you think?

tjanez:
> Another alternative to the `--no-binary resolwe` "dance" would be to only make sdist releases (i.e. no wheels) while Resolwe has dependency links.

As discussed off-line, I've changed the approach to only make sdist releases while Resolwe has dependency links.

@mstajdohar, @dblenkus, please review!

PR #207, opened 2017-01-30
Split flow models and enable isort

dblenkus:


kostko:
Can you summarize what the split is about?

dblenkus:
To split the `models` module to more readable pieces. And as a side-effect I've moved some utility functions that don't belong to models.

PR #209, opened 2017-01-31
Fix permissions in ES tests

mstajdohar:


docs/contributing.rst
```diff
@@ -93,7 +93,7 @@ Running tests
 To run the tests, use::
 
     cd tests
-    ./manage.py test resolwe --parallel
+    ./manage.py test resolwe --parallel=2
```
mstajdohar: I have decided to include `=2` as an example. Also 2 will work best on most operating systems.

PR #211, opened 2017-02-02
Add `descriptor_dirty` flag to Data, Collection and Entity

dblenkus:


resolwe/flow/models/utils.py
```diff
@@ -137,11 +141,14 @@ def validate_data(data_pk, type_):
...
+    is_dirty = False
+    dirty_field = ''
     for _schema, _fields, _ in iterate_schema(instance, schema):
         name = _schema['name']
 
-        if test_required and _schema.get('required', True) and name not in _fields:
-            raise ValidationError("Required field \"{}\" not given.".format(name))
+        if not is_dirty and test_required and _schema.get('required', True) and name not in _fields:
+            is_dirty = True
+            dirty_field = name
```
kostko: If multiple fields are dirty, this will just show the last one. Should we use a list instead?

resolwe/flow/tests/test_validation.py
```diff
@@ -7,51 +7,44 @@
...
 from resolwe.flow.models.utils import validate_schema
 from resolwe.test import TestCase
 
 
 class ValidationTest(TestCase):
 
+    def setUp(self):
+        super(ValidationTest, self).setUp()
+
+        self.user = get_user_model().objects.create(username="test_user")
```
kostko: Single quotes.

resolwe/flow/models/utils.py
dblenkus: Good idea. I've tried to keep it as it was and just show the first one, but it would be even better to list all fields.

PR #214, opened 2017-02-07
Add `descriptor_completed` field to Entity

dblenkus:


resolwe/elastic/apps.py
```diff
@@ -12,7 +12,8 @@ class ElasticConfig(AppConfig):
     def ready(self):
         """Perform application initialization."""
         is_testing = sys.argv[1:2] == ['test']
-        if is_testing:
+        is_migrating = sys.argv[1:2] == ['migrate']
+        if is_testing or is_migrating:
             # Do not register signals and ES indices when testing
```
kostko: Please update the comment to explain why this must be disabled in migrations (e.g. because the model instances used during migrations do not contain the full functionality of models and things like content types don't work correctly).

Issue #213, opened 2017-02-07
Test framework deletes test directories when `keep_all` is set


dblenkus:
`self.keep_all()` keeps test directories (i.e. `1`, `2`,...), but test framework deletes main test directory (i.e. `test_1`) at the end.

dblenkus:
Also `purge` shouldn't be called when `keep_all` is used.

mstajdohar:
Fixed

PR #217, opened 2017-02-09
Introduce keep_data instead of keep_all and keep_failed

mstajdohar:


mstajdohar:
@jkokosar, @JenkoB, note the change.

mstajdohar:
Closes #213.

tjanez:
Looks good so far

resolwe/test/testcases.py
```diff
@@ -146,29 +147,31 @@ def setUp(self):
...
             shutil.rmtree(settings.FLOW_EXECUTOR['DATA_DIR'], ignore_errors=True)
 
         self.settings.disable()
 
         super(TransactionTestCase, self).tearDown()
 
-    def keep_all(self):
-        """Do not delete output files after test for all data."""
-        self._keep_all = True
+    def keep_data(self, mock_purge=True):
```
dblenkus: I don't like `mock_purge` name. User who will use this doesn't need to know what mock is. It should describe what it does, not how it does.

PR #219, opened 2017-02-13
Add special ES fields: Name, ProcessType

kostko:


resolwe/elastic/tests/test_app/elastic_indexes.py
```diff
@@ -14,13 +14,18 @@ class TestSearchDocument(BaseDocument):
...
+
     class Meta:
         index = 'test_search'
 
 
 class TestSearchIndex(BaseIndex):
     mapping = {
         'num': 'number',
+        'field_name': 'name',
+        'field_process_type': 'name',
```
dblenkus: Will be much more clear if you add `type` or `process_type` field instead of just using `name`.

PR #222, opened 2017-02-16
Add relations between entities

dblenkus:


resolwe/flow/models/entity.py
```diff
@@ -26,3 +27,39 @@ class Meta(BaseCollection.Meta):
...
+
+
+class PositionInRelation(models.Model):
+    """Intermediate modell to connect Entities with Relations."""
+
+    entity = models.ForeignKey(Entity, on_delete=models.CASCADE)
+
+    relation = models.ForeignKey(Relation, on_delete=models.CASCADE)
+
+    position = models.CharField(max_length=100, null=True, blank=True)
```
kostko: I assume the datatype of `position` is based on specific use cases? Perhaps it would be useful to have this as a JSON field in order to be more general (the position can then be an integer or something else).

resolwe/flow/models/entity.py
```diff
@@ -26,3 +27,39 @@ class Meta(BaseCollection.Meta):
 
     #: indicate whether `descriptor` is completed (set by user)
     descriptor_completed = models.BooleanField(default=False)
+
+
+class RelationType(models.Model):
+    name = models.CharField(max_length=100, unique=True)
+
+    ordered = models.BooleanField(default=False)
```
kostko: If it is ordered, what is the order key? Is it "position"? How is it ordered (e.g. what is the comparator)?

dblenkus: Idea is that client (or package with defined relation types) takes care of ordering. 
`position` can be something like `0`, `1`, `beginning`, `intermediate `, `end`, `case`, `control`,... So there is no general rule for ordering.

kostko: > Idea is that client (or package with defined relation types) takes care of ordering.

But in this case, the backend cannot return entities in proper order and all sorting must be done on the client? What if there are a lot of entities and the client only wants a subset? Or is this not a supported use case?

dblenkus: Ordering for specific relation types can be implemented in packages where those types are defined (i.e. resolwe-bio). Normally there won't be lots of entities in single relation (<10), so ordering them on the client shouldn't be a problem.

kostko: Ok.

resolwe/flow/tests/fixtures/relationtypes.yaml
```diff
@@ -0,0 +1,11 @@
+- model: flow.relationtype
+  pk: 1
+  fields:
+    name: replicates
+    ordered: False
+
+- model: flow.relationtype
+  pk: 2
+  fields:
+    name: timeseries
```
mstajdohar: Maybe use a dash: `time-series`.

mstajdohar:
Fix linters

mstajdohar:
@dblenkus Rebase and merge.

dblenkus:
This is ready for another review pass. @mstajdohar, @kostko

resolwe/flow/models/entity.py
```diff
@@ -26,3 +28,47 @@ class Meta(BaseCollection.Meta):
...
+        permissions = (
+            ("view_relation", "Can view relation"),
+            ("edit_relation", "Can edit relation"),
+            ("share_relation", "Can share relation"),
+            ("owner_relation", "Is owner of the relation"),
+        )
+
+    type = models.ForeignKey('RelationType', on_delete=models.PROTECT)
+
+    collection = models.ForeignKey('Collection', on_delete=models.CASCADE)
```
mstajdohar: Do we want the cascade rule?

PR #225, opened 2017-02-20
Use refresh=True in ES writes/deletes

dblenkus:


mstajdohar:
We will have to monitor how this impacts the response time. @dblenkus how long does the index update take on the QA server?

dblenkus:
It is really quick. Main part (generation of the document) is blocking anyway. This will have more impact, when there will be multiple writes (of same kind of documents) in same second.

PR #227, opened 2017-02-21
Set appropriate SELinux labels when mounting tools in Docker containers

tjanez:
This is needed on hosts where SELinux is enabled.

@dblenkus, @mstajdohar, can you please run (one) [process test case in Resolwe Bioinformatics](https://github.com/genialis/resolwe-bio/tree/master/resolwe_bio/tests/processes) with this PR since I'm not sure how this new flag behaves on systems without SELinux.


tjanez:
Note that this is needed by genialis/resolwe-bio#221.

PR #230, opened 2017-02-26
Remove write and exec permissions after data object is done

dblenkus:


mstajdohar:
@dblenkus Rebase and address Tadej's comments.

dblenkus:
@tjanez I've addressed your comments regarding current implementation. For improvement proposal I think that it makes sense, but is out of scope for this pull request.

dblenkus:
@tjanez Tests probably fail because of the permissions issue in new containers?

PR #231, opened 2017-02-27
Use sha256 in data checksum and make sure that input keys are ordered

dblenkus:


dblenkus:
I can add test, but the only scenario I can imagine is different order of dict keys. The problem is that the old code also took care for this, but looks like that py2 is not reliable. The following code returns `True` in py3 and `False` in py2. But in general dict keys are ordered before serializing also in py2.

    str({u'tss': 0, u'genome': u'HG19'}) == str({u'genome': u'HG19', u'tss': 0})

resolwe/flow/utils/__init__.py
```diff
@@ -24,7 +25,7 @@
 def get_data_checksum(proc_input, proc_slug, proc_version):
     """Compute checksum of processor inputs, name and version."""
     checksum = hashlib.sha1()
```
kostko: BTW, any reason for SHA1?

kostko:
> I can add test, but the only scenario I can imagine is different order of dict keys.

Yes, can you look into how to test this? Possibly by generating a dict and performing some updates in order to make it force a different ordering, which fails on the old code and passes on the new one.

resolwe/flow/utils/__init__.py
dblenkus: Probably not, I will change it to SHA256 as we already have to update all checksums

dblenkus:
Updated

resolwe/flow/utils/__init__.py
tjanez: This is even more relevant since [Google uncovered the first practical attack on SHA1](https://security.googleblog.com/2017/02/announcing-first-sha1-collision.html).

PR #232, opened 2017-02-28
Use last descriptor schema when creating entity

dblenkus:


resolwe/flow/models/data.py
```diff
@@ -217,7 +217,7 @@ def create_entity(self):
                 entity = entity_query.first()
             else:
 
-                descriptor_schema = DescriptorSchema.objects.get(slug=ds_slug)
+                descriptor_schema = DescriptorSchema.objects.filter(slug=ds_slug).order_by('version').last()
```
kostko: Can we instead use `.latest()` and define `get_latest_by` meta to use the `version` field by default?

dblenkus:
Cool, I didn't know that `latest` exists. I've added another commit with changes.

kostko:
Nice, looks good!

PR #236, opened 2017-03-09
Improve permissions handling

### LINKED ISSUE #234: Permission inheritance for newly registered processes
**Problem:** When new process is registered, non-admin user cannot run it (or any older process with same slug), because he doesn't have permission on that process and API takes the newest.

**Solution:** New process should probably inherit permissions from the last process with same slug. On the other hand API should take the newest process on which user has permission, not the newest in general.

dblenkus:
Closes #234

### DIFF
--- a/resolwe/flow/views.py
+++ b/resolwe/flow/views.py
@@ -118,12 +118,14 @@ class ResolweCreateModelMixin(mixins.CreateModelMixin):
         ds_slug = request.data.get('descriptor_schema', None)
         if ds_slug:
             ds_query = DescriptorSchema.objects.filter(slug=ds_slug)
-            if not ds_query.exists():
+            ds_query = get_objects_for_user(request.user, 'view_descriptorschema', ds_query)
+            try:
+                request.data['descriptor_schema'] = ds_query.latest().pk
+            except DescriptorSchema.DoesNotExist:
                 return Response(
                     {'descriptor_schema': [
                         'Invalid descriptor_schema slug "{}" - object does not exist.'.format(ds_slug)]},
                     status=status.HTTP_400_BAD_REQUEST)
-            request.data['descriptor_schema'] = ds_query.latest().pk
 
         request.data['contributor'] = user.pk
         try:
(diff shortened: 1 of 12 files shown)


resolwe/flow/views.py
```diff
@@ -200,20 +202,13 @@ def create(self, request, *args, **kwargs):
         # translate processe's slug to id
         process_slug = request.data.get('process', None)
         process_query = Process.objects.filter(slug=process_slug)
+        process_query = get_objects_for_user(request.user, 'view_process', process_query)
         if not process_query.exists():
-            # XXX: security - is it ok to reveal which processes (don't) exist?
```
kostko: What was the original reasoning behind this security-related comment?

resolwe/flow/views.py
```diff
@@ -118,6 +118,7 @@ def create(self, request, *args, **kwargs):
         ds_slug = request.data.get('descriptor_schema', None)
         if ds_slug:
             ds_query = DescriptorSchema.objects.filter(slug=ds_slug)
+            ds_query = get_objects_for_user(request.user, 'view_descriptorschema', ds_query)
             if not ds_query.exists():
```
kostko: Why is this being handled with `.exists()` (also below)? I think it would be better to catch the `DoesNotExist` exception that `.latest()` will throw.

resolwe/flow/views.py
```diff
@@ -158,6 +159,7 @@ def update(self, request, *args, **kwargs):
...
```
kostko: As above.

resolwe/flow/views.py
```diff
@@ -200,20 +202,13 @@ def create(self, request, *args, **kwargs):
         # translate processe's slug to id
         process_slug = request.data.get('process', None)
         process_query = Process.objects.filter(slug=process_slug)
+        process_query = get_objects_for_user(request.user, 'view_process', process_query)
         if not process_query.exists():
```
kostko: As above.

resolwe/flow/views.py
dblenkus: Before there were different status codes (401, 403, 404) indicating if process exists or not. Now you get 404 if you don't have the permission or if the process doesn't exist.

resolwe/flow/views.py
dblenkus: Agree

PR #238, opened 2017-03-14
File archiver

mstajdohar:


mstajdohar:
@hadalin Here is the file archiver process you can use to download a batch of outputs.

resolwe/flow/processes/arciver.yml
```diff
@@ -0,0 +1,26 @@
...
+      type: list:basic:string
+  output:
+    - name: archive
+      label: Archive
+      type: basic:file
+  run:
+    language: bash
+    program: |
+      zip -0 results.zip {% for d in data %}{% for f in fields%}"{{d[f].file}}" {% endfor %}{% endfor %}
+      echo "{\"archive\":{\"file\":\"results.zip\"}}"
```
kostko: Why don't you use `re-save-file` instead?

resolwe/flow/processes/arciver.yml
```diff
@@ -0,0 +1,26 @@
+- slug: arciver
```
tjanez: "h" is missing in the slug and the YAML file name.

resolwe/flow/processes/arciver.yml
tjanez: Probably because he wants it to work with every Docker image, not only ours?

resolwe/flow/processes/arciver.yml
```diff
@@ -0,0 +1,26 @@
...
+      label: Output file fields
+      type: list:basic:string
+  output:
+    - name: archive
+      label: Archive
+      type: basic:file
+  run:
+    language: bash
+    program: |
+      zip -0 results.zip {% for d in data %}{% for f in fields%}"{{d[f].file}}" {% endfor %}{% endfor %}
```
tjanez: I would add a check if `zip` command is available and return a descriptive error if not.

tjanez: Add a space between `fields` and `%`. Also, add spaces around `d[f].file`.

resolwe/flow/processes/arciver.yml
kostko: Oh yes, we still have this as a separate package. We should really implement all these functions as predefined macros available as part of the `bash` language, as otherwise we will have these echos everywhere.

mstajdohar:
Thanks for the comments. I have addressed the typos. 

> Why don't you use re-save-file instead?

Because I wanted to make it independent of Resolwe Runtime Utils.

resolwe/flow/processes/archiver.yml
```diff
@@ -0,0 +1,26 @@
...
+      label: Output file fields
+      type: list:basic:string
+  output:
+    - name: archive
+      label: Archive
+      type: basic:file
+  run:
+    language: bash
+    program: |
+      zip -0 results.zip {% for d in data %}{% for f in fields %}{% if d[f].file %}"{{ d[f].file }}" {% for r in d[f].refs %}"{{r}}" {% endfor %}{% endif %}{% endfor %}{% endfor %}
```
dblenkus: Does this works for `basic:dir:` fields?

mstajdohar:
> Does this work for basic:dir: fields?

No. Also, it does not work for nested outputs (in groups). We do not need either yet.

docs/CHANGELOG.rst
```diff
@@ -47,6 +48,7 @@ Fixed
 - Data objects created by the workflow inherite it's permissions
 - If user doesn't have permissions on the latest versions of processes
   and descriptor schemas, use older ones except returning error
+- Support data: and list:data types
```
tjanez: Use ``` ``data:`` ``` and ``` ``list:data:`` ```.

resolwe/flow/processes/archiver.yml
```diff
@@ -0,0 +1,28 @@
...
+      label: Output file fields
+      type: list:basic:string
+  output:
+    - name: archive
+      label: Archive
+      type: basic:file
+  run:
+    language: bash
+    program: |
+      zippp -0 results.zip {% for d in data %}{% for f in fields %}{% if d[f].file %}"{{ d[f].file }}" {% for r in d[f].refs %}"{{r}}" {% endfor %}{% endif %}{% endfor %}{% endfor %}
```
tjanez: Since this is getting quite long, consider using `\` to break things into multiple lines. The code will get more readable and it will be easier to review changes in the future.

Also, surround `r` in `{{r}}` with spaces.

tjanez: Regarding my [previous comment about checking if the `zip` command is available](https://github.com/genialis/resolwe/pull/238#discussion_r105906010), maybe a  better solution would be specify a requirement on the Docker image and use some minimal Docker image with `zip` utility installed. Something like https://hub.docker.com/r/resolwe/test/.

I can prepare the image.

hadalin: `zippp`?

mstajdohar: Can you check if resolwe/test has zip?

docs/CHANGELOG.rst
mstajdohar: Usually we omit the last `:`, because our platform handles it automatically. Since `data` was uninformative I decided to make an exception here and write `data:`.

resolwe/test/testcases.py
```diff
@@ -281,7 +281,7 @@ def setUp(self):
         self._register_schemas()
 
         self.collection = Collection.objects.create(contributor=self.admin, name="Test collection")
-        self.files_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'files')
+        self.files_path = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', 'toolkit', 'tests', 'files'))
```
dblenkus: I don't like changing this here. Move this to `toolkit`'s tests to:

1. keep this test case general
2. ensure independency of apps

mstajdohar: Before it was even worse. The `self.files_path` pointed to a folder that did not exist at all. I suggest we keep my change until we have process tests in two places.

mstajdohar: I also do not like that this breaks the independency of apps. There is no good general way to support this though. I would have to override the `self.files_path` in every single test class.

dblenkus: You can create main test class in toolkit and override it there. It would be probably the best if error is raised if `self.files_path` is not overridden in child class.

mstajdohar: We need a test class we can use to test processes anywhere. It would be extremely inconvenient if we would always have to subclass the process test class.

dblenkus: But at the moment test files are not used anywhere (if directory doesn't even exist till now). I think that right solution, if needed, would be to include general test processes in `test` app and not use real processes from other apps.

Or maybe even better would be to find a way to find the path to where the code is executed and not where it lives.

mstajdohar: But what about tests of the real processes like in the resolwe/toolkit and resolwe-bio/processes? We should have automatic files discovery for this tests.

docs/CHANGELOG.rst
dblenkus: Add backticks

tests/settings.py
```diff
@@ -122,6 +123,21 @@
     'resolwe.flow.finders.AppDirectoriesFinder',
 )
 
+FLOW_DOCKER_MAPPINGS = [
+    {'src': os.path.join(FLOW_EXECUTOR['DATA_DIR'], '{data_id}'),
+     'dest': '/home/biolinux/data',
```
dblenkus: Change this to more general path.

tests/settings.py
```diff
@@ -122,6 +123,21 @@
...
+     'mode': 'rw,Z'},
+    {'src': FLOW_EXECUTOR['DATA_DIR'],
+     # NOTE: Destination directory must not be under /home/biolinux as the
+     # latter has the owner and group recursively changed to match the host
+     # user's and that cannot work with read-only volumes.
+     'dest': '/data_all',
+     'mode': 'ro,z'},
+    {'src': FLOW_EXECUTOR['UPLOAD_DIR'],
+     'dest': '/home/biolinux/upload',
```
dblenkus: This also.

tests/settings.py
```diff
@@ -122,6 +123,21 @@
     'resolwe.flow.finders.AppDirectoriesFinder',
 )
 
+FLOW_DOCKER_MAPPINGS = [
```
dblenkus: Should we also set the default docker image in settings?

tests/settings.py
mstajdohar: `/home/resolwe/data`?

mstajdohar: It does not work on another path. @tjanez please review and consult.

tests/settings.py
mstajdohar: Done

resolwe/permissions/tests/test_shortcuts.py
```diff
@@ -159,21 +159,27 @@ def test_user_permissions(self):
             {'permissions': ['edit', 'view'], 'type': 'group', 'id': self.group1.pk, 'name': 'Test group 1'},
         ]
         perms = get_object_perms(self.collection, self.user1)
-        six.assertCountEqual(self, expected_perms, perms)
+
+        def _sort_perms(perms):
```
dblenkus: Define this as a class method and also fix permissions assertions in above test.

mstajdohar: Done

tjanez:
@mstajdohar, new Docker images are ready.

PR #239, opened 2017-03-14
Set data status to error if worker crashes

dblenkus:


resolwe/flow/models/data.py
```diff
@@ -259,7 +259,8 @@ def save(self, render_name=False, *args, **kwargs):
 
         self.save_storage(self.output, self.process.output_schema)  # pylint: disable=no-member
 
-        hydrate_size(self)
+        if self.status != Data.STATUS_ERROR:
```
dblenkus: @mstajdohar, @kostko how should failed data objects be treated in save?

mstajdohar:
Improve tests. Make sure we test the error handling.

resolwe/flow/models/data.py
dblenkus: At the moment only output validation and file size hydration are skipped. If any other step will fail, also error handling will fail. One example is that invalid json is stored to json field. But do we want to create storage objects also for failed object? Or at least try and leave it it data object if it fails?

Anyway this will be quite challenging to render because of inconsistencies.

kostko: > One example is that invalid json is stored to json field. But do we want to create storage objects also for failed object? Or at least try and leave it it data object if it fails?

I think that all fields which could potentially be corrupted should be reset in case of an error. And for debugging we should have some explicit output/artifacts.

dblenkus:
I will merge this PR as is. It handles all currently known issues, so it is ok for now and we can handle any additional issues when they occur.

PR #240, opened 2017-03-15
Split resolwe.flow.testcases module into multiple modules

tjanez:
This will avoid `resolwe.flow.testcases` module becoming too long (i.e. more than 1000 lines).

Please, advise whether this deserves a CHANGELOG entry or not.


mstajdohar:
Fix tests and linter errors.

tjanez:
> I think that the changelog entry is not required, because this doesn't change the way it is used (all imports stays the same).

Agreed.

PR #243, opened 2017-03-16
Add Resolwe toolkit Docker images

tjanez:
This is needed by #238.


tjanez:
I've also configured appropriate [automated builds on Docker Hub](https://hub.docker.com/u/resolwe/).

PR #245, opened 2017-03-17
Test file upload processes and fix tabular upload

mstajdohar:


mstajdohar:
@tjanez The `re-import` command in resolwe/base docker image does not export the NAME environment variable.

tjanez:
@mstajdohar, I've changed `re-import` to the old implementation as a Bash function in `~/.bash_profile` in #246.

resolwe/toolkit/tests/test_files.py
```diff
@@ -0,0 +1,39 @@
...
+        upload_file = self.run_process('upload-file', {'src': 'file binary'})
+        self.assertFile(upload_file, 'file', 'file binary')
+
+    @with_docker_executor
+    def test_upload_image(self):
+        upload_image = self.run_process('upload-image-file', {'src': 'file image.png'})
+        self.assertFile(upload_image, 'file', 'file image.png')
+
+    @with_docker_executor
+    def test_upload_tabular(self):
```
tjanez: I would break the 4 tests into separate test methods so developers/testers can more easily see progress when calling tests with `-v 2`.

PR #220, opened 2017-02-13
Add 'with_custom_executor' and 'with_resolwe_host' test decorators

tjanez:
- Add tests for accessing Resolwe API from a process
- Add `TransactionProcessTestCase` based on `TransactionTestCase`

~~**WIP: If tests are run together with some `resolwe.permissions.test_shortcuts.GetObjectsForUser` tests, they fail.**~~

~~For example:~~
```
(resolwe) [tadej@tlinux64 tests (add-resolwe-api-host-tests)]$ ./manage.py test -v 2 resolwe.flow.tests.test_access_api resolwe.permissions.tests.test_shortcuts.GetObjectsForUser.test_has_any_group_permissions
Creating test database for alias 'default' ('resolwe_test')...
Operations to perform:
  Synchronize unmigrated apps: elastic, mathfilters, permissions, resolwe, rest_framework, staticfiles, versionfield
  Apply all migrations: auth, contenttypes, flow, guardian, sessions
Synchronizing apps without migrations:
  Creating tables...
    Running deferred SQL...
Running migrations:
  Applying contenttypes.0001_initial... OK
  Applying contenttypes.0002_remove_content_type_name... OK
  Applying auth.0001_initial... OK
  Applying auth.0002_alter_permission_name_max_length... OK
  Applying auth.0003_alter_user_email_max_length... OK
  Applying auth.0004_alter_user_username_opts... OK
  Applying auth.0005_alter_user_last_login_null... OK
  Applying auth.0006_require_contenttypes_0002... OK
  Applying auth.0007_alter_validators_add_error_messages... OK
  Applying auth.0008_alter_user_username_max_length... OK
  Applying flow.0001_initial... OK
  Applying flow.0002_project_to_collection... OK
  Applying flow.0003_support_sample... OK
  Applying flow.0004_autoslug_field... OK
  Applying flow.0005_process_data_name... OK
  Applying flow.0006_data_named_by_user... OK
  Applying flow.0007_add_owner... OK
  Applying flow.0008_fix_jsonfields... OK
  Applying flow.0009_data_parents... OK
  Applying flow.0010_fix_jsonfields... OK
  Applying flow.0011_calculate_checksum... OK
  Applying flow.0012_require_checksum... OK
  Applying flow.0013_add_requirements... OK
  Applying flow.0014_add_entity... OK
  Applying flow.0015_make_data_indexes... OK
  Applying flow.0016_update_versionfield... OK
  Applying flow.0017_update_checksum... OK
[...]

PR #247, opened 2017-03-17
Set environment variables in executor

mstajdohar:


mstajdohar:
New configuration environment variables: `RESOLWE_API_USERNAME` and `RESOLWE_API_PASSWORD`

docs/CHANGELOG.rst
```diff
@@ -28,6 +28,7 @@ Added
 - Archive file process
 - File upload processes
 - Resolwe process tests
+- Add RESOLWE_API_USERNAME and RESOLWE_API_PASSWORD configuration options
```
dblenkus: Describe where and why settings were added.

kostko:
> How do you think this would fit into the resources constrains you are working on?

Hm, so these are both security tokens, which are VERY sensitive. AFAIK, each process that needs them should be explicitly granted permissions to have these tokens (e.g., via a special `api` resource, which should also specify which permissions should be granted).

IMO the best approach would be to create a temporary security token before running the process, which is then revoked immediately after the process exits.

mstajdohar:
Let's not over-engineer here. The goal should be that we do not have to pass `resdk` credentials to the nodes at all. When we will introduce an upgrade of the workflow language (to support the demultiplexing use case) we should remove this _temporary hack_.

kostko:
If this is a temporary hack, please add `XXX: Temporary hack, remove when we support demultiplexing.` in appropriate places.

mstajdohar:
> please add XXX: Temporary hack, remove when we support demultiplexing

Will do. 

> But note that this hack introduces a potential security risk.

This hack does not introduce a security risk if not used. We should not and will not use this on public servers.

mstajdohar:
With this change even the `RESOLWE_HOST_URL` is not needed anymore. We should remove it when we update the install scripts.

docs/CHANGELOG.rst
```diff
@@ -28,6 +28,7 @@ Added
 - Archive file process
 - File upload processes
 - Resolwe process tests
+- Add SET_ENV setting to set environment variables in executor
```
tjanez: I would change this to something like:
```
``FLOW_EXECUTOR['SET_ENV']`` setting that allows setting environment variables in executor
```

resolwe/flow/tests/test_env_vars.py
```diff
@@ -12,33 +15,42 @@ class EnvVarsTest(TestCase):
...
-            contributor=self.contributor,
-            type='test:data:envvars:',
-            input_schema=[],
-            output_schema=[
-                {'name': 'resolweapihost', 'type': 'basic:string:'},
-            ],
-            run={
-                'language': 'bash',
-                'program': """
+        flow_executor = copy.copy(getattr(settings, 'FLOW_EXECUTOR', {}))
```
tjanez: After #220 is merged, this will be much nicer to write with the `@with_custom_executor` decorator.

PR #249, opened 2017-03-18
Add support for ES index dependencies

kostko:


resolwe/elastic/builder.py
```diff
@@ -40,13 +47,70 @@ def __init__(self, index, method_name):
         """Initialize signal."""
         self.index = index
         self.method_name = method_name
+        self.connections = []
+
+    def connect(self, signal_type, **kwargs):
```
dblenkus: To me it would make more sense if you rename `signal_type` to `signal`.

PR #252, opened 2017-03-23
Support file lists in archiver

mstajdohar:
This should fix the bug with archiver. Please review ASAP.


docs/CHANGELOG.rst
```diff
@@ -29,6 +29,7 @@ Added
 - File upload processes
 - Resolwe process tests
 - Add SET_ENV setting to set environment variables in executor
+- Support file lists in archiver
```
tjanez: I think this is unnecessary since the change log for this release already has `Archive file process` under _Added_

PR #253, opened 2017-03-23
Enable filtering relations by type and label

dblenkus:


resolwe/flow/views.py
```diff
@@ -625,10 +625,18 @@ def _filter_queryset(self, queryset):
         # query_params must be casted to dict, otherwise list values cannot be retrieved
         query_params = dict(self.request.query_params)
 
+        rel_type = query_params.get('type', None)
```
mstajdohar: Why not just `type` or `relation_type`?

dblenkus: `type` is reserved word. I can change it to `relation_type`.

PR #255, opened 2017-03-27
Add tags

kostko:


resolwe/flow/models/entity.py
```diff
@@ -29,6 +29,9 @@ class Meta(BaseCollection.Meta):
     #: indicate whether `descriptor` is completed (set by user)
     descriptor_completed = models.BooleanField(default=False)
 
+    #: tags for categorizing objects
+    tags = ArrayField(models.CharField(max_length=255), default=list)
```
mstajdohar: Can you pass a `set` for the default callable or does Django have something like a SetField? Tag values are unique.

kostko: Passing a set will not change anything. We can add a special field which transforms this on save (but you can't guarantee this at the database level).

PR #260, opened 2017-03-30
Add serializer arguments parameter to paginate response utility function

hadalin:


resolwe/elastic/viewsets.py
```diff
@@ -128,7 +128,7 @@ def filter_permissions(self, search):
 class PaginationMixin(object):
     """Mixin for making paginated response in case pagination parameters are provided."""
 
-    def paginate_response(self, queryset):
+    def paginate_response(self, queryset, **serializers_kwargs):
```
dblenkus: Use `serializers_kwargs` instead of `**serializers_kwargs`, otherwise it will be hard to pass attributes for other functions if needed.

PR #261, opened 2017-04-02
Run manager when new data object is created

dblenkus:


kostko:
Looks good, but it would be great to have a test that checks this. Note that since the signal defers `communicate` until the end of transaction, you need to use a `TransactionTestCase` to test this properly.

dblenkus:
@kostko I've added tests for processes. Can you please add relevant tests for workflows?

@tjanez I've cherry-picked one commit from #220 that I needed for writing tests.

resolwe/flow/executors/__init__.py
```diff
@@ -251,49 +251,53 @@ def run(self, data_id, script, verbosity=1):
...
-
-        try:
-            # Cleanup after processor
-            data_purge(data_ids=[data_id], delete=True, verbosity=verbosity)
-        except:  # pylint: disable=bare-except
-            logger.error(__("Purge error:\n\n{}", traceback.format_exc()))
+        # This transaction is needed to make sure that processing of
+        # current data object is finished before manager for spawned
+        # processes is triggered.
+        with transaction.atomic():
```
dblenkus: @kostko, @mstajdohar What do you think of this solution? It only causes problems in tests, because manager is synchronous, but I think that it is good to handle it properly anyway.

mstajdohar: I agree

tjanez:
> @tjanez I've cherry-picked one commit from #220 that I needed for writing tests.

Sure, I though #220 will be merged already, but I still haven't debugged the Travis test failures.

resolwe/flow/executors/__init__.py
kostko: Looks good.

PR #262, opened 2017-04-03
Add Null executor

dblenkus:


resolwe/test/utils.py
```diff
@@ -88,3 +88,26 @@ def wrapper(*args, **kwargs):
             manager.discover_engines()
 
     return wrapper
+
+
+def with_null_executor(method):
```
dblenkus: @tjanez You will have to update this in #220

mstajdohar:
Django calls such code _dummy_, for instance `DummyCache`.

PR #264, opened 2017-04-06
Retry slug generation on data races

kostko:


resolwe/flow/views.py
```diff
@@ -39,6 +39,9 @@
     ProcessSerializer, RelationSerializer, StorageSerializer,
 )
 
+# Maximum number of slug-generation retries.
+MAX_SLUG_RETRIES = 10
```
mstajdohar: Should move this to Django settings?

kostko:
I've updated the fix, so it is now more general (works for all `BaseModel` subclasses).

resolwe/flow/views.py
kostko: I don't think this should be a setting as it is mainly there to prevent infinite loops. If 10 retries are not enough, there must be serious data insert contention, which should be examined.

PR #265, opened 2017-04-06
Filter hydrated data by user permissions

dblenkus:


resolwe/flow/serializers.py
```diff
@@ -171,18 +173,36 @@ class Meta:
         fields = ('slug', 'name', 'description', 'settings', 'descriptor_schema', 'descriptor',
                   'data') + update_protected_fields + read_only_fields
 
+    def _serialize_data(self, data):
+        """Return serialized data or list of ids, depending on `hydrate_data` qquery param."""
```
kostko: One extra `q`.

dblenkus:
@kostko I've changed the query to `collection.data.all()`, so now it should work with prefetch and use values cached in `collection` object.

PR #271, opened 2017-04-18
Add support for filter in ES M2M dependency

kostko:


mstajdohar:
The test coverage has decreased slightly.

PR #273, opened 2017-04-22
Fix problem with shared ES connection between workers

dblenkus:


dblenkus:
I've added a warning to the `prepare_connection` function.

PR #285, opened 2017-05-18
Fix public permissions in elastic app

dblenkus:
Fixes https://github.com/genialis/bug-reports/issues/24


dblenkus:
@kostko, good idea! I've added it.

PR #291, opened 2017-05-23
Treat None values as if the field is missing

kostko:


resolwe/flow/models/utils.py
```diff
@@ -154,6 +154,10 @@ def validate_data(data_pk, type_):
             field = _fields[name]
             type_ = _schema.get('type', "")
 
+            # Treat None as if the field is missing.
+            if not _schema.get('required', True) and field is None:
```
dblenkus: Save `_schema.get('required', True)` to variable and also use it 10 lines above.

PR #293, opened 2017-05-26
Add resource requirements

kostko:
* [x] Support and tests for network resource.

### DIFF
--- a/resolwe/flow/executors/docker/__init__.py
+++ b/resolwe/flow/executors/docker/__init__.py
@@ -1,13 +1,16 @@
 """Docker workflow executor."""
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+import json
 import os
 import shlex
 import subprocess
+import tempfile
 
 from django.conf import settings
 
-from .local import FlowExecutor as LocalFlowExecutor
+from ..local import FlowExecutor as LocalFlowExecutor
+from .seccomp import SECCOMP_POLICY
 
 
 class FlowExecutor(LocalFlowExecutor):
@@ -20,6 +23,7 @@ class FlowExecutor(LocalFlowExecutor):
         super(FlowExecutor, self).__init__(*args, **kwargs)
 
         self.mappings_tools = None
+        self.policy_file = None
         self.command = getattr(settings, 'FLOW_DOCKER_COMMAND', 'docker')
 
     def start(self):
@@ -30,15 +34,66 @@ class FlowExecutor(LocalFlowExecutor):
             'container_image': self.requirements.get('image', settings.FLOW_EXECUTOR['CONTAINER_IMAGE']),
         }
 
+        # Get limit defaults and overrides.
+        limit_defaults = getattr(settings, 'FLOW_DOCKER_LIMIT_DEFAULTS', {})
+        limit_overrides = getattr(settings, 'FLOW_DOCKER_LIMIT_OVERRIDES', {})
+
+        # Set resource limits.
+        limits = []
+        if 'cores' in self.resources:
+            # Each core is equivalent to 1024 CPU shares. The default for Docker containers
+            # is 1024 shares (we don't need to explicitly set that).
+            limits.append('--cpu-shares={}'.format(int(self.resources['cores']) * 1024))
+
+        memory = limit_overrides.get('memory', {}).get(self.process.slug, None)
+        if memory is None:
+            memory = int(self.resources.get(
+                'memory',
+                # If no memory resource is configured, check settings.
+                limit_defaults.get('memory', 4096)
+            ))
+
+        # Set both memory and swap limits as we want to enforce the total amount of memory
+        # used (otherwise swap would not count against the limit).
+        limits.append('--memory={0}m --memory-swap={0}m'.format(memory))
+
+        command_args['limits'] = ' '.join(limits)
+
         # set container name
         container_name_prefix = getattr(settings, 'FLOW_EXECUTOR', {}).get('CONTAINER_NAME_PREFIX', 'resolwe')
         command_args['container_name'] = '--name={}_{}'.format(container_name_prefix, self.data_id)
 
-        # Configure Docker network mode for the container (if specified).
-        # By default, current Docker versions use the 'bridge' mode which
-        # creates a network stack on the default Docker bridge.
-        network = getattr(settings, 'FLOW_EXECUTOR', {}).get('NETWORK', '')
-        command_args['network'] = '--net={}'.format(network) if network else ''
+        if 'network' in self.resources:
+            # Configure Docker network mode for the container (if specified).
+            # By default, current Docker versions use the 'bridge' mode which
+            # creates a network stack on the default Docker bridge.
+            network = getattr(settings, 'FLOW_EXECUTOR', {}).get('NETWORK', '')
+            command_args['network'] = '--net={}'.format(network) if network else ''
+        else:
+            # No network if not specified.
+            command_args['network'] = '--net=none'
+
+        # Security options.
+        security = []
+
+        # Generate and set seccomp policy to limit syscalls.
+        self.policy_file = tempfile.NamedTemporaryFile(mode='w')
+        json.dump(SECCOMP_POLICY, self.policy_file)
+        self.policy_file.file.flush()
+        if not getattr(settings, 'FLOW_DOCKER_DISABLE_SECCOMP', False):
+            security.append('--security-opt seccomp={}'.format(self.policy_file.name))
+
+        # Drop all capabilities and only add ones that are needed.
+        security.append('--cap-drop=all')
+        # TODO: We should not required dac_override, but we need to fix permissions first (e.g., use --user).
+        security.append('--cap-add=dac_override')
+        # TODO: We should not require setuid/setgid, but we need to get rid of gosu first (e.g., use --user).
+        security.append('--cap-add=setuid')
+        security.append('--cap-add=setgid')
+        # TODO: We should not required chown, but we need to fix permissions first (e.g., use --user).
+        security.append('--cap-add=chown')
+
+        command_args['security'] = ' '.join(security)
 
         # render Docker mappings in FLOW_DOCKER_MAPPINGS setting
         mappings_template = getattr(settings, 'FLOW_DOCKER_MAPPINGS', [])
@@ -76,8 +131,8 @@ class FlowExecutor(LocalFlowExecutor):
 
         self.proc = subprocess.Popen(
             shlex.split(
-                '{command} run --rm --interactive {container_name} {network} {volumes} {envs} '
-                '{workdir} {container_image} {shell}'.format(**command_args)),
+                '{command} run --rm --interactive {container_name} {network} {volumes} {envs} {limits} '
+                '{security} {workdir} {container_image} {shell}'.format(**command_args)),
             stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
             universal_newlines=True)
 
@@ -97,6 +152,9 @@ class FlowExecutor(LocalFlowExecutor):
     def end(self):
         """End process execution."""
         self.proc.wait()
+        if self.policy_file is not None:
+            self.policy_file.close()
+            self.policy_file = None
 
         return self.proc.returncode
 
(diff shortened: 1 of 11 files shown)


kostko:
I have now added network resources and further security hardening using both SECCOMP and dropping of ALL capabilities. Processes which require network access will require changes.

mstajdohar:
cc @jkokosar Review the existing processes and write a list that will require updates.

kostko:
Damn, it seems seccomp is not working on Travis:
```
seccomp profiles are not supported on this daemon, you cannot specify a custom seccomp profile.
```

This is an issue as we want to have seccomp enabled in all cases. Perhaps we should just disable it on Travis via an environmental variable, e.g. `RESOLWE_DOCKER_DISABLE_SECCOMP=1`? @tjanez?

resolwe/flow/executors/docker/__init__.py
```diff
@@ -30,15 +34,45 @@ def start(self):
...
         }
 
+        # Set resource limits.
+        limits = []
+        if 'cores' in self.resources:
+            # Each core is equivalent to 1024 CPU shares. The default for Docker containers
+            # is 1024 shares.
+            limits.append('--cpu-shares={}'.format(int(self.resources['cores']) * 1024))
+
+        if 'memory' in self.resources:
```
dblenkus: Should we set default limit if it is not present?

kostko: We could. What should it be?

kostko:
I've now change it so that SECCOMP is disabled on Travis. One problem with this is that `resolwe-bio` processes will also not be tested under SECCOMP, which may cause some of them to fail later (most likely not as they don't use any strange syscalls).

Just be sure to test everything on a SECCOMP-enabled system before deploy.

kostko:
Seems some processes are now failing with permission-related errors. This is due to the fact that we now drop all capabilities and it seems that we are not handling file permissions correctly and were relying on the root user having permissions to modify everything?

resolwe/flow/executors/docker/__init__.py
dblenkus: @jkokosar, @mstajdohar do you have any suggestion? Maybe 1 core and 4-8gb of memory?

kostko: One core is already the default (for all Docker containers).

kostko:
I've re-added some capabilities (`dac_override, setuid, setgid`), which are needed due to the way permissions are currently handled. But I've added TODOs that these should be changed and these capabilities should be dropped as well.

resolwe/flow/executors/docker/__init__.py
```diff
@@ -30,15 +35,51 @@ def start(self):
...
+            limits.append('--memory={0}m --memory-swap={0}m'.format(int(self.resources['memory'])))
```
tjanez: Could you add a comment why also setting `--memory-swap` is needed?

resolwe/flow/executors/docker/__init__.py
```diff
@@ -30,15 +35,51 @@ def start(self):
...
+            command_args['network'] = '--net=none'
+
+        # Security options.
+        security = []
+
+        # Generate and set seccomp policy to limit syscalls.
+        self.policy_file = tempfile.NamedTemporaryFile(mode='w')
+        json.dump(SECCOMP_POLICY, self.policy_file)
+        self.policy_file.file.flush()
+        if getattr(settings, 'RESOLWE_DOCKER_DISABLE_SECCOMP', None) != '1':
```
tjanez: Please, use [`distutils.util.strtobool`](https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool) here to make parsing of the setting more robust and user-friendly.

resolwe/flow/static/flow/processorSchema.json
```diff
@@ -114,6 +114,32 @@
...
+                    "properties": {
+                        "cores": {
+                            "description": "Number of CPU cores",
+                            "type": "integer"
+                        },
+                        "memory": {
+                            "description": "Amount of memory in megabytes",
+                            "type": "integer"
+                        },
+                        "temporary_storage": {
```
tjanez: Do you plan to tackle these two (`temporary_storage` and `output_storage`) in this PR or later?

resolwe/flow/executors/docker/seccomp.py
```diff
@@ -0,0 +1,1681 @@
+"""SECCOMP policy."""
```
tjanez: Could you expand this docstring?

It would be nice to mention how did you obtain this policy and an overview of what it blocks/allows.

resolwe/flow/executors/docker/__init__.py
```diff
@@ -30,15 +35,51 @@ def start(self):
...
+            security.append('--security-opt seccomp={}'.format(self.policy_file.name))
+
+        # Drop all capabilities and only add ones that are needed.
+        security.append('--cap-drop=all')
+        # TODO: We should not required dac_override, but we need to fix permissions first (e.g., use --user).
+        security.append('--cap-add=dac_override')
+        # TODO: We should not require setuid/setgid, but we need to get rid of gosu first (e.g., use --user).
+        security.append('--cap-add=setuid')
+        security.append('--cap-add=setgid')
```
tjanez: Don't forget to temporarily include `chown` capability which is needed by [docker-bio-linux8-resolwe's entrypoint script](https://github.com/genialis/docker-bio-linux8-resolwe/blob/master/docker-entrypoint.sh).

resolwe/flow/static/flow/processorSchema.json
kostko: I've added them as they are also present in CWL and make sense to have at some point. But I think this is something we can add later as it is not clear how exactly we will handle them.

resolwe/flow/executors/docker/__init__.py
kostko: Done.

resolwe/flow/executors/docker/__init__.py
```diff
@@ -30,15 +36,56 @@ def start(self):
             'container_image': self.requirements.get('image', settings.FLOW_EXECUTOR['CONTAINER_IMAGE']),
         }
 
+        # Set resource limits.
+        limits = []
+        if 'cores' in self.resources:
+            # Each core is equivalent to 1024 CPU shares. The default for Docker containers
+            # is 1024 shares (we don't need to explicitly set that).
+            limits.append('--cpu-shares={}'.format(int(self.resources['cores']) * 1024))
```
dblenkus: There are some processes that supports setting number of cores as input parameter (i.e. [cuffquant](https://github.com/genialis/resolwe-bio/blob/ed1c342c8517bb8625e381b2e299ded790105706/resolwe_bio/processes/expression/cuffquant.yml#L56). ). How will this be handled?

kostko: Resource requirements are currently static.

Note that these CPU limits are not hard limits (e.g., this does not mean that a container with a limit of 1024 shares will only be given one fixed dedicated core), they are only in effect when all the running containers are using all of the CPUs. So it is more like a relative weight.

mstajdohar: @jkokosar Do we still need this? What is the use case?

dblenkus: It would be great if executor injects limits in process so tools can be run depending on resources that they have. I.e. instead of:

    cuffquant --num-threads '{{ threads }}'

you can run:
 
    cuffquant --num-threads '{{ limits.cores }}'

mstajdohar:
We could enforce in the register function that values of the string choice fields are text.

resolwe/flow/executors/docker/__init__.py
kostko: In this case I think it would be better to expose all requirements as `requirements` in the context passed to the expression engine?

kostko: The problem with this is that currently all inputs are directly exposed as top-level elements in the context. We can't just add `requirements` in there as there could be an input called `requirements`?

resolwe/flow/static/flow/processorSchema.json
tjanez: Ok, agreed.

resolwe/flow/executors/docker/__init__.py
tjanez: I agree that it would be useful to expose all requirements to processes so they can, for example, bound their memory usage to the given resource limit.

Regarding the implementation, I'm more in favor of just making `requirements` an invalid input name 🙂 . I think it will be easier for process writers to use `requirements.resources.memory` or `requirements.resources.cores` than `_requirements.resources.memory` or `_requirements.resources.cores`.

kostko: Done, we now expose `requirements`.

kostko:
I'll run the `resolwe-bio` test suite to see what needs updating there before we merge this.

kostko:
Some minor changes were needed (see genialis/resolwe-bio#285) to make the tests pass. But we will probably need to do further testing and update resources based on that. cc @jkokosar

mstajdohar:
Please also describe this in docs. This should probably go here: http://resolwe.readthedocs.io/en/latest/proc.html

kostko:
> Please also describe this in docs.

Done. I've also described all the other requirements that we currently support.

resolwe/flow/executors/docker/__init__.py
```diff
@@ -30,15 +35,56 @@ def start(self):
...
+            command_args['network'] = '--net=none'
+
+        # Security options.
+        security = []
+
+        # Generate and set seccomp policy to limit syscalls.
+        self.policy_file = tempfile.NamedTemporaryFile(mode='w')
+        json.dump(SECCOMP_POLICY, self.policy_file)
+        self.policy_file.file.flush()
+        if not strtobool(getattr(settings, 'RESOLWE_DOCKER_DISABLE_SECCOMP', '0')):
```
dblenkus: I think it would be better to move `strtobool` to settings and also set default there.

kostko:
I've now changed resource defaults and overrides for the memory resource, which can now be configured in settings:
```python
# Global default for processes without explicit requirements.
RESOLWE_DOCKER_LIMIT_DEFAULTS = {
    'memory': 4096,
}

# Per-process resource limit overrides.
RESOLWE_DOCKER_LIMIT_OVERRIDES = {
    'memory': {
        'alignment-star': 8192,  # Process identified by its slug.
    }
}
```

@tjanez was this what you had in mind?

tjanez:
> I've now changed resource defaults and overrides for the memory resource, which can now be configured in settings:
> ```python
> # Global default for processes without explicit requirements.
> RESOLWE_DOCKER_LIMIT_DEFAULTS = {
>     'memory': 4096,
> }
> 
> # Per-process resource limit overrides.
> RESOLWE_DOCKER_LIMIT_OVERRIDES = {
>     'memory': {
>         'alignment-star': 8192,  # Process identified by its slug.
>     }
> }
> ```
> 
> @tjanez was this what you had in mind?

@kostko, yes, even better than what I had in mind 🙂 !

tjanez:
@kostko, one more thing. I know we currently don't have all Docker executor setting in one place (e.g. we have `FLOW_DOCKER_COMMAND`, `FLOW_DOCKER_MAPPINGS` and `FLOW_EXECUTOR['CONTAINER_IMAGE'`], which is really the Docker image etc.), but it would be more consistent (for the time being) if you renamed all new settings you introduced from `RESOLWE_DOCKER_*` to `FLOW_DOCKER_*`.

kostko:
Hm, that is true, but I find the naming strange as the whole application is called "Resolwe" not "Flow" :-) The `resolwe.flow` is just one part so to me either `RESOLWE_*` or `RESOLWE_FLOW_*` would be more correct than `FLOW_*`.

But I agree that we should be consistent for now, so I'll change the new settings' names.

tjanez:
> Hm, that is true, but I find the naming strange as the whole application is called "Resolwe" not "Flow" :-) The `resolwe.flow` is just one part so to me either `RESOLWE_*` or `RESOLWE_FLOW_*` would be more correct than `FLOW_*`.

Yes, agreed.

> But I agree that we should be consistent for now, so I'll change the new settings' names.

Thanks!

We should definitely consolidate, rename and document our settings in the near future.

kostko:
Renamed.

kostko:
Are we good to merge?

PR #297, opened 2017-06-07
Refactor flow's views

dblenkus:


resolwe/flow/tests/test_api.py
```diff
@@ -225,6 +243,58 @@ def test_add_remove_data(self):
...
+        collection.entity_set.add(entity_1, entity_2)
+
+        assign_perm("view_collection", self.user, collection)
+        assign_perm("edit_collection", self.user, collection)
+
+        request = factory.delete('{}?delete_content=1'.format(self.detail_url(collection.pk)))
+        force_authenticate(request, self.user)
+        self.collection_detail_viewset(request, pk=collection.pk)
+
+        # Only objects with `edit` permission can be deleted.
```
kostko: But in this case, some objects are deleted while others are silently kept and (for the caller) there is no way to know that this happened? This is also inconsistent with how filesystem permissions behave.

resolwe/flow/views/collection.py
```diff
@@ -0,0 +1,111 @@
...
+        for data in obj.data.all():
+            if user.has_perm('share_data', data):
+                update_permission(data, payload)
+
+    def destroy(self, request, *args, **kwargs):
+        """Destroy a model instance.
+
+        If ``delete_content`` flag is set in query parameters, also all
+        Data objects and Entities contained in collection will be
+        deleted.
```
kostko: This comment is not true, only objects for which the user has the edit permission will be deleted.

resolwe/flow/views/collection.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        obj = self.get_object()
+        user = request.user
+
+        if strtobool(request.query_params.get('delete_content', 'false')):
+            for entity in obj.entity_set.all():
+                if user.has_perm('edit_entity', entity):
+                    entity.delete()
```
kostko: What happens with data objects, which are part of the entity? What if they are not part of the same collection? What if they have different permissions that the entity object?

resolwe/flow/views/entity.py
```diff
@@ -0,0 +1,139 @@
...
+        return resp
+
+    @detail_route(methods=[u'post'])
+    def remove_data(self, request, pk=None):
+        """Remove Data from Entity and delete it if it is empty."""
+        resp = super(EntityViewSet, self).remove_data(request, pk)
+
+        entity = self.get_object()
+        if entity.data.count() == 0:
+            entity.delete()
```
kostko: Is this needed? There is a signal handler in `signals.py`, which does this automatically?

dblenkus: Good catch 👍🏼

PR #298, opened 2017-06-08
Add process scheduling class

kostko:


resolwe/flow/static/flow/processorSchema.json
```diff
@@ -39,6 +39,11 @@
             "type": "string",
             "pattern": "^RAW|CACHED|TEMP$"
         },
+        "scheduling_class": {
+            "description": "Scheduling class",
+            "type": "string",
+            "enum": ["interactive", "batch"]
```
dblenkus: Options in `persistence` are enforced with `"pattern": "^RAW|CACHED|TEMP$"`. Make this consistent.

resolwe/flow/managers/base.py
```diff
@@ -148,6 +148,9 @@ def communicate(self, run_sync=False, verbosity=1):
                     if program is not None:
                         priority = 'normal'
                         if data.process.persistence == Process.PERSISTENCE_TEMP:
+                            # TODO: This should probably be removed.
```
dblenkus: Yes this is only for backward compatibility.

resolwe/flow/static/flow/processorSchema.json
mstajdohar: I prefer enum. Still, follow Domen's advice and make it consistent.

kostko: I've fixed the `persistence` schema to use `enum` instead.

PR #305, opened 2017-06-27
Replace django-autofield with custom slug field

### LINKED ISSUE #206: Remove dependency link on django-autoslug
As it is, `pip install --process-dependecy-links resolwe` fails when it is run the second time since pip uses the cached wheel it built the first time Resolwe was installed. Hence, it doesn't install Resolwe from the sdist release, which supports dependency links, but from binary/wheel release which doesn't support dependency links.

To resolve this, remove dependency link on django-autoslug and release a forked version on PyPI if the [issue](https://github.com/neithere/django-autoslug/pull/18) is not resolved by the next release.

dblenkus:
Closes #206.


resolwe/flow/models/fields.py
```diff
@@ -0,0 +1,91 @@
...
+            query_params = {
+                'slug': slug,
+                'slug_column': self.column,
+                'table_name': model._meta.db_table,  # pylint: disable=protected-access
+            }
+
+            with connection.cursor() as cursor:
+                cursor.execute(
+                    """
+                    SELECT MAX(slug_index) + 1 FROM (
```
dblenkus: @kostko any operation (i.e. `MAX`, `ORDER BY`, ...) on `slug_index` is slow due to lack of index. Is it possible to create index only on index of slug (`slug`field is composed by slug and index, i.e. `my-slug-30`) and then sort by it?

resolwe/flow/models/fields.py
```diff
@@ -0,0 +1,91 @@
...
+                                RIGHT({slug_column}, -length('{slug}')-1),
+                                ''
+                            ),
+                            '0'
+                        )::text::integer AS slug_index
+                        FROM {table_name} WHERE (
+                            {slug_column} ~ '^{slug}(-[0-9]*)?$'
+                        )
+                    ) AS tmp
+                    """.format(**query_params)
```
kostko: Yikes! Never do this as you just introduced an SQL injection vulnerability. Also, why don't you use Django's raw query instead of accessing the underlying backend (psycopg2) directly?

kostko: Ok, it may not be that bad due to `slugify`, but I still don't like it.

resolwe/flow/models/fields.py
kostko: Yes, you can [index by an arbitrary expression in PostgreSQL](https://www.postgresql.org/docs/9.6/static/indexes-expressional.html), so you could then sort by that same expression and an index should be used.

resolwe/flow/models/fields.py
```diff
@@ -0,0 +1,158 @@
...
+            slug = None
+
+            if not self.blank:
+                slug = instance._meta.model_name  # pylint: disable=protected-access
+            elif not self.null:
+                slug = ''
+
+        if slug:
+            # Make sure that potential sequence added won't excede maximal length.
+            slug = slug[:(self.max_length - 10)]
```
kostko: Is the assumption here that the sequential number will be at most 9 digits long? This is probably ok, but perhaps it should be stored in a constant like `MAX_SLUG_SEQUENCE_DIGITS`.

resolwe/flow/models/fields.py
```diff
@@ -0,0 +1,158 @@
...
+            instance_pk_name = instance._meta.pk.name  # pylint: disable=protected-access
+
+            query_params = {
+                'constraints_placeholder': constraints_placeholder,
+                'slug_column': self.column,
+                'table_name': self.model._meta.db_table,  # pylint: disable=protected-access
+                'pk_neq_placeholder': 'AND {} != %s'.format(instance_pk_name) if instance.pk else ''
+            }
+
+            query_escape_params = [slug]
```
kostko: You could instead use named parameters so this definition would be less fragile.

resolwe/flow/models/fields.py
```diff
@@ -0,0 +1,158 @@
...
+                                        {constraints_placeholder}
+                                    )
+                                )
+                            ) THEN MAX(slug_sequence) + 1
+                            ELSE NULL
+                        END
+                    FROM (
+                        SELECT COALESCE(
+                            NULLIF(
+                                RIGHT({slug_column}, -length(%s)-1),
```
kostko: Isn't this length constant? Any reason why you don't compute it in Python and just provide the actual value here?

resolwe/flow/models/fields.py
```diff
@@ -0,0 +1,158 @@
...
+            # Make sure that potential sequence added won't excede maximal length.
+            slug = slug[:(self.max_length - 10)]
+
+            constraints_placeholder, constraints_values = self._get_unique_constraints(instance)
+
+            instance_pk_name = instance._meta.pk.name  # pylint: disable=protected-access
+
+            query_params = {
+                'constraints_placeholder': constraints_placeholder,
+                'slug_column': self.column,
```
kostko: You should probably use `django.db.connection.ops.quote_name` for column and table names.

resolwe/flow/models/fields.py
kostko: And perhaps it should be validated that the generated sequence number does not exceed this amount of digits (raise an error otherwise).

dblenkus:
@kostko thanks for comments! I've added another commit with fixes (and some extra features), so it is now ready for the final review.

resolwe/flow/models/fields.py
```diff
@@ -0,0 +1,185 @@
...
+                    ) AS tmp
+                    """.format(**query_params),
+                    params=query_escape_params
+                )
+                result = cursor.fetchone()[0]
+
+            if result is not None:
+                if len(str(result)) > MAX_SLUG_SEQUENCE_DIGITS:
+                    raise DatabaseError(
+                        'Auto-generated slug sequence too long - please chose a different slug.'
```
kostko: chose -> choose

PR #306, opened 2017-06-29
Add input string variable sanitization

kostko:
This is a **breaking change** that will require fixing of all processes using the `bash` execution engine.

TODO
* [x] Add documentation for new behavior under the Jinja expression engine (note that escape syntax is expression engine dependent).
* [x] Add sanitization tests.
* [x] Go through resolwe-bio and fix all processes.
* [ ] Go through pipelines-* and fix all processes.

### DIFF
--- a/docs/proc.rst
+++ b/docs/proc.rst
@@ -265,8 +265,10 @@ fieldSchema.json.
 Run
 ---
 
-The algorithm that transforms inputs into outputs. Bash runtime is supported,
-but we envision more runtimes in the future (`e.g.,` a Python or R runtime). Commands should be written to a ``bash`` subfield.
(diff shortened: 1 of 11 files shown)


docs/proc.rst
```diff
@@ -474,11 +481,29 @@ referenced data objects:
...
+      volcanoplot.py {{ reads.fastq.0.file }}
+
+    will actually be transformed into something like (depending on the value):
+
+    .. code-block:: bash
+
+      volcanoplot.py '/path/to/reads with spaces.gz'
+
+    If you do not want this behaviour for a certain variable and you are sure
+    that it is safe to do so, you can use the ``safe`` filter.
```
dblenkus: It would be nice to add example with `safe` filter.

PR #307, opened 2017-07-03
Prepare release 1.5.0

dblenkus:
@tjanez please merge #303 before the release.
@kostko Do we wait for #306? When will it be ready?

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -6,9 +6,9 @@ All notable changes to this project are documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_.
 
 
-==========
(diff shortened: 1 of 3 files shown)


mstajdohar:
I suggest to include https://github.com/genialis/resolwe/pull/306 in the next release, because it is a breaking change.

kostko:
> When will it be ready?

The major thing is fixing all issues in resolwe-bio. I would like to do that before we merge that one as otherwise everything will break once merged.

tjanez:
> @tjanez please merge #303 before the release.

Done

tjanez:
> I suggest to include #306 in the next release, because it is a breaking change.

I agree.

PR #311, opened 2017-07-11
Add UNICODE characters to test files and remove redundant code

jvrakor:
@mstajdohar @jkokosar @dblenkus


mstajdohar:
We should test and validate the existing processes that depend on this process.

resolwe/toolkit/tools/parse_tabular_file.py
```diff
@@ -23,7 +21,7 @@ def main():
     args = parse_arguments()
 
     ext = os.path.splitext(args.input_file)[-1].lower()
-    with gzip.open(args.output_file, mode='wt') as outfile:
+    with gzip.open(args.output_file, mode='wt', encoding='utf-8') as outfile:
```
tjanez: The proper fix for using UTF-8 encoding by default is to [coerce the default `C` locale in the Docker container to `C.UTF-8`](https://www.python.org/dev/peps/pep-0538/).

Fedora will [properly coerce the `C` locale to `C.UTF-8` in version 26](https://fedoraproject.org/wiki/Changes/python3_c.utf-8_locale) which was released just today.

I'll prepare a PR that will update the `resolwe/base` image to Fedora 26.

resolwe/toolkit/tools/parse_tabular_file.py
```diff
@@ -38,11 +36,10 @@ def main():
                         outline = line.strip().split(',')
                         csvwriter.writerow(outline)
             elif ext in ('.xls', '.xlsx'):
-                with open(args.input_file) as infile:
-                    workbook = xlrd.open_workbook(args.input_file)
-                    worksheet = workbook.sheets()[0]
-                    for rownum in range(worksheet.nrows):
-                        csvwriter.writerow(worksheet.row_values(rownum))
+                workbook = xlrd.open_workbook(args.input_file)
```
tjanez: You want to use `xlrd.open_workbook()` with a `with` statement to release resources after finishing the `with` code block.

jvrakor: I believe ```xlrd.open``` does not need closing as it isn't the same as ```open```
https://groups.google.com/forum/#!msg/python-excel/U2mKQkfLkGQ/juEAQsxlI-YJ

mstajdohar:
We have updated the Fedora version in our docker images. This feature should now work out-of-the-box. Please test and close this PR if everything works as expected.

resolwe/toolkit/tools/parse_tabular_file.py
tjanez: @jvrakor, you are right, `xlrd.open_workbook()` doesn't hold the file open.

Here is a simple script that proves this:
```python
import psutil
import xlrd

proc = psutil.Process()

print(proc.open_files())

wb = xlrd.open_workbook('file tab.1.xlsx')

print(proc.open_files())

f = open('file tab.1.xlsx')

print(proc.open_files())
```

Its output is:
```
[]
[]
[popenfile(path='/home/tadej/Genialis/resolwe/resolwe/toolkit/tests/files/file tab.1.xlsx', fd=3, position=0, mode='r', flags=557056)]
```
which means that only the builin `open()` actually holds the file descriptor open.

tjanez:
@jvrakor, @mstajdohar, the other two changes in the PR are OK and improve the quality of the `parse_tabular_file.py ` tool.

mstajdohar:
Rebase to master and add a note to the change log.

PR #319, opened 2017-07-18
Remove remaining references to using dependency links

tjanez:
Update README to no longer use lines starting with "From"


mstajdohar:
Maybe also update the `--process-dependency-links` in `contributing.rst`.

tjanez:
> Maybe also update the `--process-dependency-links` in `contributing.rst`.

I already have 🙂

PR #322, opened 2017-07-20
Improve permissions

dblenkus:


resolwe/permissions/tests/test_permissions.py
```diff
@@ -148,18 +154,50 @@ def test_protect_owner(self):
 
         assign_perm("owner_collection", self.user1, self.collection)
 
-        # User with share permission can grant ``owner`` permission
+        # ``owner`` permission cannot be assigned to a group
+        data = {'groups': {'add': {self.group.pk: ['owner']}}}
+        resp = self._detail_permissions(self.collection.pk, data, self.user1)
+        self.assertEqual(resp.status_code, status.HTTP_403_FORBIDDEN)
```
kostko: I think bad request would be the more correct error code?

dblenkus: Agree, fixed.

PR #334, opened 2017-08-01
Add Entity to the same Collection as Data when the latest is created

dblenkus:


resolwe/flow/executors/__init__.py
```diff
@@ -302,10 +303,16 @@ def run(self, data_id, script, verbosity=1):
                         # Copy permissions.
                         copy_permissions(parent_data, d)
 
+                        entities = Entity.objects.filter(data=d).annotate(num_data=Count('data')).filter(num_data=1)
```
mstajdohar: Include a note about this "hack".

resolwe/flow/views/data.py
```diff
@@ -115,8 +116,14 @@ def perform_create(self, serializer):
             for permission in list(zip(*instance._meta.permissions))[0]:  # pylint: disable=protected-access
                 assign_perm(permission, instance.contributor, instance)
 
+            entities = Entity.objects.filter(data=instance).annotate(num_data=Count('data')).filter(num_data=1)
```
mstajdohar: Add a note about this "hack".

PR #336, opened 2017-08-01
Add command to list Docker images used by processes

abukosek:
The `list_docker_images` management command prepares a list of Docker images used by processes (in YAML format, suitable for use in Ansible).

### DIFF
--- a/resolwe/flow/management/commands/list_docker_images.py
+++ b/resolwe/flow/management/commands/list_docker_images.py
@@ -1,0 +1,64 @@
+""".. Ignore pydocstyle D400.
+
+==================
+List Docker images
+==================
+
+"""
+
+import functools
+import operator
+
+import yaml
+
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+
+from resolwe.flow.models import Process
+
+
+class Command(BaseCommand):
+    """List Docker images used by processes."""
+
+    help = "List Docker images used by processes in either plain text or YAML (for Ansible)"
+
+    def add_arguments(self, parser):
+        """Add an argument to specify output format."""
+        parser.add_argument('--format', dest='format', default='plain',
+                            help="Set output format ('plain' [default] or 'yaml')")
+
+    def handle(self, *args, **options):
+        """Handle command list_docker_images."""
+        # Check that the specified output format is valid
+        if options['format'] != 'plain' and options['format'] != 'yaml':
+            raise CommandError("Unknown output format: %s" % options['format'])
+
+        # Gather a list of all custom Docker requirements the processes are using
+        all_docker_requirements = [
+            p.requirements['executor']['docker']
+            for p in Process.objects.filter(requirements__icontains='docker')
+        ]
+
+        # Filter the list to include only unique images (also, the 'images' field is optional)
+        unique_docker_images = set(i['image'] for i in all_docker_requirements if 'image' in i)
+
+        # Add the default image if it exists
+        if 'CONTAINER_IMAGE' in settings.FLOW_EXECUTOR:
+            # This Docker image is used if a process doesn't specify its own
+            default_docker_image = settings.FLOW_EXECUTOR['CONTAINER_IMAGE']
+
+            unique_docker_images.add(default_docker_image)
+
+        # Convert the set of unique Docker images into a list of dicts
+        imgs = [
+            dict(name=s[0], tag=s[1] if len(s) == 2 else 'latest')
+            for s in (img.split(':') for img in unique_docker_images)
+        ]
+
+        # Output list in specified format
+        if options['format'] == 'yaml':
+            out = yaml.safe_dump(imgs, default_flow_style=True, default_style="'")
+        else:
+            out = functools.reduce(operator.add, ('{name}:{tag}\n'.format(**i) for i in imgs), '')
+
+        self.stdout.write(out, ending='')
--- a/resolwe/flow/tests/test_commands.py
+++ b/resolwe/flow/tests/test_commands.py
@@ -3,7 +3,10 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import os
 
+import yaml
+
 from django.core.management import call_command
+from django.core.management.base import CommandError
 from django.test import TestCase as DjangoTestCase
 from django.utils.six import StringIO
 
@@ -11,7 +14,7 @@ from guardian.models import UserObjectPermission
 from guardian.shortcuts import assign_perm, get_perms
 
 from resolwe.flow.models import Process
-from resolwe.test import TestCase
+from resolwe.test import ProcessTestCase, TestCase
 
 PROCESSES_DIR = os.path.join(os.path.dirname(__file__), 'processes')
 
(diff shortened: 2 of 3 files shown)


resolwe/flow/management/commands/list_docker_images.py
```diff
@@ -0,0 +1,34 @@
+from django.core.management.base import BaseCommand, CommandError
+from django.conf import settings
+from resolwe.flow.models import Process
+
+
+class Command(BaseCommand):
+	help = "List Docker images used by processes in a format suitable for use in Ansible"
```
kostko: Indentaion seems wrong ;-)

resolwe/flow/management/commands/list_docker_images.py
```diff
@@ -0,0 +1,34 @@
...
+		for i in unique_docker_images:
+			# Split Docker image specifier into image name and tag
+			s = i.split(':')
+			img = s[0]
+			tag = s[1] if len(s) == 2 else 'latest'
+
+			if out:
+				out += ', '
+
+			out += "{ 'name': '" + img + "', 'tag': '" + tag + "' }"
```
kostko: I would add an argument to select the output format (`--format`), with default to `plain` (output image names with tag, one per line) and `yaml`.

kostko:
Also add tests for the management command.

resolwe/flow/management/commands/list_docker_images.py
abukosek: Oops, fixed :D

resolwe/flow/management/commands/list_docker_images.py
abukosek: Added!

resolwe/flow/management/commands/list_docker_images.py
```diff
@@ -0,0 +1,64 @@
...
+        if options['format'] != 'plain' and options['format'] != 'yaml':
+            raise CommandError("Unknown output format: %s" % options['format'])
+
+        # This Docker image is used if a process doesn't specify its own
+        default_docker_image = settings.FLOW_EXECUTOR['CONTAINER_IMAGE']
+
+        # Gather a list of all custom Docker images the processes are using
+        unique_docker_images = set(
+            p.requirements['executor']['docker']['image']
+            for p in Process.objects.filter(requirements__icontains='docker'))
```
dblenkus: Move closing parenthesis to a new line.

resolwe/flow/management/commands/list_docker_images.py
```diff
@@ -0,0 +1,64 @@
...
+
+        unique_docker_images.add(default_docker_image)
+
+        # Figure out whether we should output in YAML
+        want_yaml = options['format'] == 'yaml'
+
+        # Assemble a list of Docker images
+        out = ''
+
+        for i in unique_docker_images:
```
dblenkus: Use [PyYAML](http://pyyaml.org/wiki/PyYAMLDocumentation) instead of manually generating yaml.

mstajdohar:
Merge commits as they all the same feature. Add a note to change log. Do not forget about tests.

resolwe/flow/tests/test_commands.py
```diff
@@ -88,3 +91,53 @@ def test_process_register_no_admin(self):
         err = StringIO()
         self.assertRaises(SystemExit, call_command, 'register', path=[PROCESSES_DIR], stderr=err)
         self.assertEqual('Admin does not exist: create a superuser\n', err.getvalue())
+
+
+class ListDockerImagesTest(ProcessTestCase):
```
kostko: Why does this require `ProcessTestCase`? The above test cases only call `register` via `call_command`, I think this should be enough here as well?

The process test case is usually used for test cases where processes need to be executed, but here I think it should be enough for them to just be registered.

abukosek: Domen told me to use `_register_schemas` instead of calling the `register` command, since it's faster.

kostko: Should we then also update all of the above tests (in a separate commit) in order to have this unified?

dblenkus: No, above tests are testing `register` function, so it has to be used. 

But in general reading and parsing yaml files is very slow (especially if you do it in each test), so `_register_schemas` only do it once and then use cached results for the rest of the test case.

kostko: Ok, sounds good.

kostko:
Will we have a separate `pull_docker_images` or can we have `docker_images` with a `--pull` argument? Because the image discovery logic will be duplicated otherwise (we could also solve this with a common utility function or inheritance).

resolwe/flow/management/commands/list_docker_images.py
```diff
@@ -0,0 +1,61 @@
...
+        imgs = [
+            dict(name=s[0], tag=s[1] if len(s) == 2 else 'latest')
+            for s in (img.split(':') for img in unique_docker_images)
+        ]
+
+        # Output list in specified format
+        if options['format'] == 'yaml':
+            out = yaml.safe_dump(imgs, default_flow_style=True, default_style="'")
+        else:
+            out = functools.reduce(operator.add, (i['name'] + ':' + i['tag'] + '\n' for i in imgs), '')
```
dblenkus: Change `i['name'] + ':' + i['tag'] + '\n'` to `'{name}:{tag}\n'.format(i)`

dblenkus:
@kostko for now the idea is to generate the list of images and them pull them separately in the deployment process (i.e. with Ansible). But I like your idea of built-in pull command, so this can be the next step.

resolwe/flow/management/commands/list_docker_images.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def handle(self, *args, **options):
+        """Handle command list_docker_images."""
+        # Check that the specified output format is valid
+        if options['format'] != 'plain' and options['format'] != 'yaml':
+            raise CommandError("Unknown output format: %s" % options['format'])
+
+        # Gather a list of all custom Docker images the processes are using
+        unique_docker_images = set(
+            p.requirements['executor']['docker']['image']
```
dblenkus: `image` is not the required field in `docker` dict, so this has to be checked separately.

abukosek:
Nits addressed, ready for re-review :)

I also like the idea of a pull command, but that is out of scope for this PR -- it should be implemented in a separate PR once this one is merged into master.

resolwe/flow/management/commands/list_docker_images.py
```diff
@@ -0,0 +1,64 @@
...
+                            help="Set output format ('plain' [default] or 'yaml')")
+
+    def handle(self, *args, **options):
+        """Handle command list_docker_images."""
+        # Check that the specified output format is valid
+        if options['format'] != 'plain' and options['format'] != 'yaml':
+            raise CommandError("Unknown output format: %s" % options['format'])
+
+        # Gather a list of all custom Docker images the processes are using
+        all_docker_images = [
```
dblenkus: Rename this to `all_docker_requirements`.

abukosek: Done :)

mstajdohar:
Rebase to master

PR #341, opened 2017-08-07
Add better error handling

dblenkus:


resolwe/flow/executors/__init__.py
```diff
@@ -67,6 +67,8 @@ def __init__(self, *args, **kwargs):
         super(BaseFlowExecutor, self).__init__(*args, **kwargs)
         self.data_id = None
         self.process = None
+        # Flag to mark current object as failed.
+        self.failed = False
```
mstajdohar: Maybe use a more verbose name like `process_failed`.

resolwe/flow/executors/__init__.py
```diff
@@ -123,28 +125,41 @@ def update_data_status(self, **kwargs):
...
             # Ensure that we only update the fields that were changed.
             data.save(update_fields=update_fields)
-        except ValidationError as ex:
-            data.process_error.append(str(ex))
+
+            if kwargs.get('status', None) == Data.STATUS_ERROR:
+                self.failed = True
+                logger.error(
+                    __("Error occured while running a '{}' process.", self.process.name),
+                    extra={'data_id': self.data_id}
```
mstajdohar: I like that you report the data ID. Should we also include a link to the data object on the REST endpoint? It will save us some time if we'll be able to navigate to the data object with one click.

resolwe/flow/executors/__init__.py
```diff
@@ -315,7 +330,7 @@ def run(self, data_id, script, verbosity=1):
                             for entity in entities:
                                 entity.collections.add(collection)
 
-            if process_rc == 0:
+            if process_rc == 0 and not self.failed:
```
mstajdohar: Couldn't you simply check if the data does not have status ERROR instead of `self.failed`?

dblenkus: No, executor doesn't have the read access to the database, so this flag has to be used to remember what status has been set.

dblenkus:
@mstajdohar fixed.

PR #343, opened 2017-08-10
Ubuntu 14.04-based base Docker image

jkokosar:


tjanez:
@jkokosar, please also update the [`README`](https://github.com/genialis/resolwe/blob/master/resolwe/toolkit/docker_images/base/README.md).

PR #344, opened 2017-08-10
Catch error occurred while processing Elasticsearch indices

dblenkus:


resolwe/elastic/indices.py
```diff
@@ -198,40 +202,51 @@ def process_object(self, obj, push=True):
...
+                    pass
 
-            # get value from the object
-            try:
-                object_value = dict_dot(obj, field)
-                setattr(document, field, object_value)
-                continue
-            except KeyError:
-                pass
+                raise AttributeError('Cannot determine mapping for field {}'.format(field))
```
kostko: Double quotes?

resolwe/elastic/indices.py
```diff
@@ -198,40 +202,51 @@ def process_object(self, obj, push=True):
...
 
-            raise AttributeError('Cannot determine mapping for field {}'.format(field))
+            except:  # pylint: disable=bare-except
+                logger.error(
```
kostko: There is also `logger.exception`, any reason you didn't use that one instead?

resolwe/elastic/indices.py
```diff
@@ -270,7 +285,24 @@ def build(self, obj=None, push=True):
         for obj in queryset:
             if self.filter(obj) is False:
                 continue
-            self.process_object(self.preprocess_object(obj), push)
+
+            try:
+                obj = self.preprocess_object(obj)
+            except:  # pylint: disable=bare-except
+                logger.error(
```
kostko: Same as above.

resolwe/elastic/indices.py
```diff
@@ -270,7 +285,24 @@ def build(self, obj=None, push=True):
...
+                    __("Error occurred while preprocessing '{}' Elasticsearch index.", self.__class__.__name__),
+                    exc_info=True,
+                    extra={'object_type': self.object_type, 'obj_id': obj.pk}
+                )
+
+            try:
+                self.process_object(obj, push)
+            except:  # pylint: disable=bare-except
+                logger.error(
```
kostko: Same as above.

resolwe/elastic/indices.py
dblenkus: Actually I did't knew it exists.

dblenkus:
@kostko I've addressed your comments.

PR #345, opened 2017-08-11
Specify tag of the required base Docker image

mstajdohar:
Set the fedora-26 tag of the resolwe/base Docker image where it was previously used without a tag.


mstajdohar:
Thanks, fixed.

PR #346, opened 2017-08-11
Reduce base docker image size

kostko:


mstajdohar:
@jkokosar Review the tricks to reduce the image size and use them in your Docker images.

PR #349, opened 2017-08-19
Use /etc/profile.d instead of .bash_profile

kostko:
Using the root user's `.bash_profile` is very ugly and should be avoided. It may also become unsupported (e.g. will not be evaluated), based on how we implement running as the correct user.

In genialis/resolwe-bio#335 the relevant section will also change to use `/etc/profile.d/resolwebio-legacy.sh` instead. Other Docker images should follow the same naming convention, each creating its own file in `/etc/profile.d`.


kostko:
That part is already completely changed (with the description already updated to mention profile.d) with the new correct-user execution. I'll make a PR soon.

PR #350, opened 2017-08-21
Execute Docker container as non-root user

kostko:
This breaks execution of any `bio-linux8` based Docker image (and possibly any other image, which requires to be run as the root user).

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -10,6 +10,10 @@ This project adheres to `Semantic Versioning <http://semver.org/>`_.
 Unreleased
 ==========
 
+Added
+-----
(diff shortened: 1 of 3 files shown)


docs/CHANGELOG.rst
```diff
@@ -10,6 +10,10 @@ This project adheres to `Semantic Versioning <http://semver.org/>`_.
 Unreleased
 ==========
 
+Changed
+-------
+- **BACKWARD INCOMPATIBLE:** Run Docker images as non-root user
```
tjanez: Just a nitpick, maybe "Run Docker containers ..." sounds better?

kostko: Fixed.

PR #352, opened 2017-08-23
Prepare release 2.0.0 and bump master to 2.1.0a1

tjanez:


tjanez:
@dblenkus, @kostko, @mstajdohar, I propose we do releases like this from now on:
- Create a release commit.
- Create a following commit that bumps version from `X.Y.Z` to `X.(Y+1).0a1`. If there are breaking changes, we bump `X` before the release.

If there are security/bug fixes needed for a released version `X.Y.0`, we create a `stable/X.Y.x` branch where we create the necessary `X.Y.{1,2,3,...}` releases.

mstajdohar:
Do we want to merge #341 before the release. @dblenkus ?

PR #353, opened 2017-08-24
Fix version handling in 'list_docker_images'

abukosek:
Previously, we listed all images, including those required for older
versions of processes, which caused problems in deployment (e.g. if
an older image was deleted from dockerhub in the meantime).

This commit changes the `list_docker_images` command to list only
the images required by the latest versions of processes.

### DIFF
--- a/resolwe/flow/management/commands/list_docker_images.py
+++ b/resolwe/flow/management/commands/list_docker_images.py
@@ -33,14 +33,21 @@ class Command(BaseCommand):
         if options['format'] != 'plain' and options['format'] != 'yaml':
             raise CommandError("Unknown output format: %s" % options['format'])
 
-        # Gather a list of all custom Docker requirements the processes are using
-        all_docker_requirements = [
-            p.requirements['executor']['docker']
-            for p in Process.objects.filter(requirements__icontains='docker')
-        ]
-
-        # Filter the list to include only unique images (also, the 'images' field is optional)
-        unique_docker_images = set(i['image'] for i in all_docker_requirements if 'image' in i)
+        # Gather only unique latest custom Docker requirements that the processes are using
+        # The 'image' field is optional, so be careful about that as well
+        unique_docker_images = set(
+            p.requirements['executor']['docker']['image']
+            for p in Process.objects.order_by(
+                'slug', '-version'
+            ).distinct(
+                'slug'
+            ).only(
+                'requirements'
+            ).filter(
+                requirements__icontains='docker'
+            )
+            if 'image' in p.requirements.get('executor', {}).get('docker', {})
+        )
 
         # Add the default image if it exists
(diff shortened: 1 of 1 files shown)


resolwe/flow/management/commands/list_docker_images.py
```diff
@@ -33,14 +33,21 @@ def handle(self, *args, **options):
...
+            for p in Process.objects.order_by(
+                'slug', '-version'
+            ).distinct(
+                'slug'
+            ).only(
+                'requirements'
+            ).filter(
+                requirements__icontains='docker'
+            )
+            if 'image' in p.requirements['executor']['docker']
```
mstajdohar: Don't know if `'executor'` and `'docker'` are required fields, but this seams safer:
```
if 'image' in p.requirements.get('executor', {}).get('docker', {})
```

resolwe/flow/management/commands/list_docker_images.py
```diff
@@ -33,14 +33,21 @@ def handle(self, *args, **options):
...
-        all_docker_requirements = [
-            p.requirements['executor']['docker']
-            for p in Process.objects.filter(requirements__icontains='docker')
-        ]
-
-        # Filter the list to include only unique images (also, the 'images' field is optional)
-        unique_docker_images = set(i['image'] for i in all_docker_requirements if 'image' in i)
+        # Gather only unique latest custom Docker requirements that the processes are using
+        # The 'image' field is optional, so be careful about that as well
+        unique_docker_images = set(
```
mstajdohar: For the sake of nicer print-outs, I would order images by name: `...sorted(set(...`.

resolwe/flow/management/commands/list_docker_images.py
kostko: There is an `requirements__icontains` check above, which filters out those without docker. But I don't think it is specific enough as it should explicitly only match `docker` inside an `executor`.

tjanez:
Just a minor comment, please don't use `` ` ``s in commit messages. While it is used in Markdown to denote inline code, it is unpleasant in other forms looking at the commit message.

You can omit the `` ` ``s completely, or use `'`s like in other commit messages:

```bash
$ git log | grep "'" | wc -l
139
$ git log | grep '`' | wc -l
0
```

resolwe/flow/management/commands/list_docker_images.py
abukosek: Added as a separate step below, since we also need to add the default image if it's present.

resolwe/flow/management/commands/list_docker_images.py
abukosek: Fixed, thanks!

abukosek:
Comments addressed, commit message fixed, ready for re-review :)

abukosek:
OK, I will write a test and put it in a separate PR.

PR #354, opened 2017-08-28
Add additional tests for 'list_docker_images'

abukosek:
This commit adds two additional tests for the `list_docker_images`
management command.  It tests sorting of the output and proper handling
of process versions.

### DIFF
--- a/resolwe/flow/tests/test_commands.py
+++ b/resolwe/flow/tests/test_commands.py
@@ -141,3 +141,21 @@ class ListDockerImagesTest(ProcessTestCase):
         self.assertTrue(len(imgs) != 0)
         self.assertTrue(isinstance(imgs[0], dict))
         self.assertTrue(dict(name='resolwe/test', tag='base') in imgs)
+
+    def test_list_is_sorted(self):
+        # The returned list must be sorted alphabetically
+        out, err = StringIO(), StringIO()
+        call_command('list_docker_images', stdout=out, stderr=err)
+        self.assertEqual('', err.getvalue())
+        self.assertNotEqual('', out.getvalue())
+        lines = out.getvalue().split('\n')[:-1]
+        self.assertEqual(lines, sorted(lines))
+
+    def test_list_versions(self):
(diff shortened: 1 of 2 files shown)


resolwe/flow/tests/test_commands.py
```diff
@@ -141,3 +141,21 @@ def test_list_has_test_image_yaml(self):
...
         self.assertTrue(dict(name='resolwe/test', tag='base') in imgs)
+
+    def test_list_is_sorted(self):
+        # The returned list must be sorted alphabetically
+        out, err = StringIO(), StringIO()
+        call_command('list_docker_images', stdout=out, stderr=err)
+        self.assertEqual('', err.getvalue())
+        self.assertNotEqual('', out.getvalue())
+        lines = out.getvalue().split('\n')[:-1]
+        self.assertEqual(lines, sorted(lines))
```
dblenkus: How do you make sure that output is not sorted by accident?

abukosek: I don't.  It's highly unlikely that this would happen by accident, even in the test environment with its small number of processes with Docker images, let alone in production.  Sorting of the output is a cosmetic feature only, anyways.

resolwe/flow/tests/test_commands.py
```diff
@@ -141,3 +141,21 @@ def test_list_has_test_image_yaml(self):
...
+
+    def test_list_versions(self):
+        # The returned list must contain only the Docker image from
+        # the latest version of a given process
+        out, err = StringIO(), StringIO()
+        call_command('list_docker_images', stdout=out, stderr=err)
+        self.assertEqual('', err.getvalue())
+        self.assertTrue('resolwe/test:versioning-2' in out.getvalue())
```
tjanez: You can use [`assertIn()` and `assertNotIn()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIn) here.

abukosek: Fixed, thanks!

PR #358, opened 2017-09-11
Add assertAlmostEqualGeneric test case helper

kostko:


resolwe/test/testcases/__init__.py
```diff
@@ -41,16 +41,22 @@
 class TestCaseHelpers(object):
     """Mixin for test case helpers."""
 
-    def assertAlmostEqualList(self, actual, expected):  # pylint: disable=invalid-name
-        """Assert almost equality of two lists."""
+    def assertAlmostEqualGeneric(self, actual, expected):  # pylint: disable=invalid-name
+        """Assert almost equality for common types of objects."""
```
tjanez: Please, write a more verbose docstring explaining that *almost equal* part applies to floats inside (nested) common container types (dicts, lists and tuples).

PR #363, opened 2017-09-14
Add missing re-import dependencies to resolwe/base

kostko:


mstajdohar:
Should you a more thorough cleanup, e.g.:
```
    apt-get autoremove --purge -y && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/* && \
    rm -rf /root/.cache/pip && \
    rm -rf /tmp/*
```

kostko:
Yes, I didn't look at cleanup but we can add it so it is similar as in `resolwebio/base`.

kostko:
Added. The `autoremove` is not needed here as we don't actually remove any build-only packages.

kostko:
Hm, actually this image should also install all the decompression utilities referenced in `re-import` as otherwise unpacking may still fail?

mstajdohar:
Yes, thanks for spotting this.

PR #364, opened 2017-09-15
Relabel temporary user/group files

kostko:
For some reason this wasn't caught by tests, but later caused problems with iClip tests on Jenkins with SELinux enabled.

This should probably be backported to 3.x.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -10,6 +10,10 @@ This project adheres to `Semantic Versioning <http://semver.org/>`_.
 Unreleased
 ==========
 
+Fixed
+-----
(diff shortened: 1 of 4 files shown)


kostko:
> For some reason this wasn't caught by tests

Probably because we bumped matplotlib version to 2.0.2 in `resolwebio/legacy`, which does not seem to trigger this issue.

kostko:
I've added a test, but this needs to (also) be tested on a SELinux-enabled system to be useful (e.g on our Jenkins server).

kostko:
@dblenkus can you backport this and release a new version with this fix, e.g. 3.0.1?

dblenkus:
Done.

PR #368, opened 2017-09-20
Add current_user_permissions to serializer's fields

dblenkus:
Broken because of https://github.com/encode/django-rest-framework/commit/b905197f24a74066e79888c56ddcf676c994cefc#diff-acb883ee0c042314f361a94aa00e5424

### DIFF
--- a/resolwe/permissions/mixins.py
+++ b/resolwe/permissions/mixins.py
@@ -7,13 +7,24 @@ from django.contrib.contenttypes.models import ContentType
 from django.db import transaction
 
 from guardian.models import UserObjectPermission
-from rest_framework import exceptions, status
+from rest_framework import exceptions, serializers, status
 from rest_framework.decorators import detail_route, list_route
 from rest_framework.response import Response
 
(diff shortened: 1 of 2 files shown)


resolwe/permissions/mixins.py
```diff
@@ -16,6 +16,25 @@
...
+    permissions = serializers.ListField(
+        child=serializers.CharField(max_length=30)
+    )
+
+    def get_fields(self):
+        """Return serializer's fields."""
+        fields = super(CurrentUserPermissionsSerializer, self).get_fields()
+
+        # Manually add `type` as it is a reserved keyword in Python.
+        fields['type'] = serializers.CharField(max_length=50)
```
kostko: This is not true, `type` is not a reserved keyword and you can easily use it in attribute access. That it looks reserved is probably an issue with syntax highlighting.

dblenkus: Correct, thanks for noticing that.

PR #369, opened 2017-09-22
Add data dependency kind

kostko:


resolwe/flow/migrations/0004_data_dependency_2.py
```diff
@@ -0,0 +1,36 @@
...
+
+    dependencies = [
+        ('flow', '0003_data_dependency_1'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='datadependency',
+            name='kind',
+            field=models.CharField(choices=[('io', 'Input/output dependency'), ('subprocess', 'Subprocess')], default='io', max_length=16),
```
kostko: We currently set the dependency kind for all existing dependencies to `io`, even though this is not entirely correct (we previously treated both dependencies as IO, but abused them in workflows and spawned processes).

We could add a data migration to correct this, but it would be quite complex in order to handle IO dependencies, workflows and spawned processes.

dblenkus: You can check inputs of the child process and if parent is there it is IO dependency and subprocess dependency otherwise.

resolwe/flow/executors/__init__.py
```diff
@@ -326,7 +326,11 @@ def run(self, data_id, script, verbosity=1):
 
                     with transaction.atomic():
                         d = Data.objects.create(**d)
-                        d.parents.add(parent_data)
+                        DataDependency.objects.create(
```
dblenkus: Same here.

resolwe/flow/migrations/0004_data_dependency_2.py
kostko: Sure, the question is do we even need to care about old dependencies?

resolwe/flow/models/data.py
```diff
@@ -196,7 +198,11 @@ def save_dependencies(self, instance, schema):
         def add_dependency(value):
             """Add parent Data dependency."""
             try:
-                self.parents.add(Data.objects.get(pk=value))  # pylint: disable=no-member
+                DataDependency.objects.update_or_create(
+                    parent=Data.objects.get(pk=value),
+                    child=self,
+                    defaults={'kind': DataDependency.KIND_IO},
```
dblenkus: Is there a reason why you don't set a default in `DataDependency.kind` field?

resolwe/flow/execution_engines/workflow/__init__.py
```diff
@@ -105,7 +105,11 @@ def evaluate(self, data):
                 contributor=data.contributor,
                 input=data_input,
             )
-            data_object.parents.add(data)
+            DataDependency.objects.create(
```
dblenkus: I don't like this logic outside of model. Can you move it to the model method, e.g. `Data.add_subprocess_dependency(<parent>)`.

resolwe/flow/migrations/0004_data_dependency_2.py
dblenkus: I think that we should keep the state consistent, no matter when the object were created. Otherwise it can cause some strange behaviour in the future.

resolwe/flow/execution_engines/workflow/__init__.py
kostko: This is the Django-documented standard way to do this, why create another method? I don't like your approach as it requires a new method for each type.

resolwe/flow/models/data.py
kostko: Yes, to force the caller to explicitly set a kind and not assume anything by default.

resolwe/flow/migrations/0004_data_dependency_2.py
kostko: > You can check inputs of the child process and if parent is there

This of course also requires duplicating the input processing code in the migration, which is IMO what makes it a bit more complex as we either need to depend on some utility functions in resolwe modules (less desirable as those functions may change later) or copy everything into the migration (so it is a snapshot in time).

resolwe/flow/execution_engines/workflow/__init__.py
dblenkus: You can pass the type to the method, it is OK for me. The main thing I don't like ist that intermediate model, which should be transparent (in Django's many to many relations), is not transparent anymore.

kostko: It is transparent in queries, just not in `add`, `create` or `set` as mentioned in [Django docs](https://docs.djangoproject.com/en/1.10/topics/db/models/#extra-fields-on-many-to-many-relationships). I can add this easily, but I think it is inconsistent about how this is handled elsewhere, e.g. `PositionInRelation` is one such intermediate model, which is always referenced directly and there are no similar helpers.

Note that we will most probably want to expose the dependency between data instances via our API as well (e.g., to show computation graphs in the frontend), which means that `DataDependency` may get its own endpoint.

resolwe/flow/migrations/0004_data_dependency_2.py
dblenkus: But I think that it is worth keeping relations properly labeled. You only have to copy `iterate_fields` function and it will get deleted once migrations will be squashed.

resolwe/flow/execution_engines/workflow/__init__.py
dblenkus: Ok, then leave it this way. I didn't know there is a plan to publicly expose this.

resolwe/flow/migrations/0004_data_dependency_2.py
kostko: Ok, I'll copy it over.

kostko: Data migration added and tested on a QA2 database dump.

Issue #189, opened 2017-01-11
Add documentation reference for Resolwe Django settings


tjanez:
We would probably want something similar to [Django's Settings Reference](https://docs.djangoproject.com/en/1.10/ref/settings/).

PR #372, opened 2017-10-02
Pull Docker images after process registration is complete

abukosek:
This PR adds the `--pull` option to the `list_docker_images` management command, as well as a `post_register_hook` to the `BaseFlowExecutor` class, which is called by the `register` command after process registration finishes.  This hook is implemented by the Docker executor to pull Docker images.

The WIP part: in its current state, it should work in production, but the tests use the local executor and override it for Docker tests, so the hook is never run.  Working on fixing this at the moment.

### DIFF
--- a/resolwe/flow/executors/docker/__init__.py
+++ b/resolwe/flow/executors/docker/__init__.py
@@ -8,6 +8,7 @@ import subprocess
 import tempfile
 
 from django.conf import settings
+from django.core.management import call_command
 
 from resolwe.flow.models import Process
 
@@ -183,3 +184,8 @@ class FlowExecutor(LocalFlowExecutor):
     def terminate(self, data_id):
         """Terminate a running script."""
         subprocess.call(shlex.split('{} rm -f {}'.format(self.command, data_id)))
+
+    def post_register_hook(self):
+        """Pull Docker images required by processes after the 'register' command is done."""
+        if not getattr(settings, 'FLOW_DOCKER_DONT_PULL', False):
+            call_command('list_docker_images', pull=True)
--- a/resolwe/flow/management/commands/list_docker_images.py
+++ b/resolwe/flow/management/commands/list_docker_images.py
@@ -8,6 +8,8 @@ List Docker images
 
 import functools
 import operator
+import shlex
+import subprocess
 
 import yaml
 
@@ -18,7 +20,7 @@ from resolwe.flow.models import Process
 
 
 class Command(BaseCommand):
-    """List Docker images used by processes."""
+    """List Docker images used by processes.  Optionally also pull them."""
 
     help = "List Docker images used by processes in either plain text or YAML (for Ansible)"
 
@@ -26,6 +28,12 @@ class Command(BaseCommand):
         """Add an argument to specify output format."""
         parser.add_argument('--format', dest='format', default='plain',
                             help="Set output format ('plain' [default] or 'yaml')")
+        parser.add_argument('--pull', dest='pull', default=False, action='store_true',
+                            help="Pull all images with Docker")
+        parser.add_argument('--ignore-pull-errors', dest='ignore_pull_errors',
+                            default=getattr(settings, 'FLOW_DOCKER_IGNORE_PULL_ERRORS', False),
+                            action='store_true',
+                            help="Don't fail whenever a Docker image can't be pulled")
 
     def handle(self, *args, **options):
         """Handle command list_docker_images."""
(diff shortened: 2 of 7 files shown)


resolwe/flow/management/commands/register.py
```diff
@@ -319,3 +319,5 @@ def handle(self, *args, **options):
                 self.find_schemas(desc_path, filters=schemas, schema_type='descriptor', verbosity=verbosity))
 
         self.register_descriptors(descriptor_schemas, user_admin, force, verbosity=verbosity)
+
+        manager.get_executor().post_register_hook()
```
kostko: Output some text to note what is happening (e.g., "Running executor post-registration hook...").

kostko:
Also, some tests would be nice, but I see that you are already working on that.

resolwe/flow/management/commands/list_docker_images.py
```diff
@@ -18,21 +20,30 @@
...
 
     help = "List Docker images used by processes in either plain text or YAML (for Ansible)"
 
     def add_arguments(self, parser):
         """Add an argument to specify output format."""
         parser.add_argument('--format', dest='format', default='plain',
                             help="Set output format ('plain' [default] or 'yaml')")
+        parser.add_argument('--pull', dest='pull', default=False, action='store_true',
+                            help="Pull all images with Docker")
+        parser.add_argument('--stop-if-pull-fails', dest='stop_if_pull_fails',
```
dblenkus: In my opinion the default behaviour should be to stop on error. Then you can also use shorter name for this argument, e.g. `--ignore-errors`.

resolwe/flow/management/commands/register.py
abukosek: Done! :)

resolwe/flow/management/commands/list_docker_images.py
abukosek: I've implemented this now -- it fails by default, unless you give it the `--ignore-pull-errors` option or set the `FLOW_DOCKER_IGNORE_PULL_ERRORS` settings variable to `True` (or the `RESOLWE_DOCKER_IGNORE_PULL_ERRORS` environment variable to `1`).

By default, errors are ignored in the test environment, but not in production.  No changes are required to the `settings.py` file in production, since the code simply defaults to fail-on-error if those variables aren't set.

abukosek:
Ready for re-review :)

The `post_register_hook` is now re-run in the existing `with_custom_executor` wrapper that's used only in tests.  It slows down testing somewhat, so it's not an ideal solution, but I think it's the least-kludgy solution and should be good enough for now.

resolwe/flow/executors/local.py
```diff
@@ -57,3 +57,7 @@ def terminate(self, data_id):
         time.sleep(self.kill_delay)
         if proc.poll() is None:
             proc.kill()
+
+    def post_register_hook(self):
```
tjanez: Hmm... isn't this redundant since it is the same as in `BaseFlowExecutor`?

tests/settings.py
```diff
@@ -102,6 +102,9 @@
 if 'RESOLWE_DOCKER_COMMAND' in os.environ:
     FLOW_DOCKER_COMMAND = os.environ['RESOLWE_DOCKER_COMMAND']
 
+# Ignore errors in tests when pulling Docker images from 'list_docker_images --pull' command
+FLOW_DOCKER_IGNORE_PULL_ERRORS = strtobool(os.environ.get('RESOLWE_DOCKER_IGNORE_PULL_ERRORS', '1'))
```
tjanez: Why do you default to `'1'` here and also explicitly set this to `1` in `.travis.yml`?

resolwe/flow/executors/docker/__init__.py
```diff
@@ -183,3 +184,7 @@ def end(self):
     def terminate(self, data_id):
         """Terminate a running script."""
         subprocess.call(shlex.split('{} rm -f {}'.format(self.command, data_id)))
+
+    def post_register_hook(self):
+        """Pull Docker images required by processes after the 'register' command is done."""
+        call_command('list_docker_images', pull=True)
```
tjanez: @abukosek, what would a rough estimate of the total size of Docker images downloaded on a deploy? And how long does it typically take to download them?

I'm thinking whether it would be wise to make the pulling configurable when we are testing our deployment scripts in Vagrant VMs.

resolwe/flow/executors/local.py
abukosek: Good catch!  I'll remove it.

tests/settings.py
abukosek: The setting in the Travis config is redundant, I'll remove it.

resolwe/flow/executors/docker/__init__.py
abukosek: The images used for tests are rather small (less than 1 GB, from memory), so pulling them takes a few minutes at most, but I agree that this should be configurable.  I'll add a settings option for this.

abukosek:
Comments addressed, ready for re-review :)

tests/settings.py
```diff
@@ -102,6 +102,12 @@
...
+
+# Don't pull Docker images (default is to pull, but we can disable this on Vagrant boxes)
```
tjanez: I would change this to something like:
```
# Don't pull Docker images if set via environment variable.
```
These are actually test project's settings, so there is no need to document the function in the Vagrant context. Unfortunately, we still [lack documentation of all available Resolwe's settings](https://github.com/genialis/resolwe/issues/189).

tests/settings.py
```diff
@@ -102,6 +102,12 @@
 if 'RESOLWE_DOCKER_COMMAND' in os.environ:
     FLOW_DOCKER_COMMAND = os.environ['RESOLWE_DOCKER_COMMAND']
 
+# Ignore errors in tests when pulling Docker images from 'list_docker_images --pull' command
```
tjanez: I would change this to:
```
# Ignore errors when pulling Docker images with 'list_docker_images --pull' command.
```

abukosek:
Done!

tjanez:
Great!

It seems we all forgot about the CHANGELOG entry. Please, add it and I will merge the PR.

tjanez:
Also, rebase the PR.

abukosek:
Rebased and added changelog entry, ready for merging!

tjanez:
@abukosek, you missed the major part that should be added to the CHANGELOG:
```
Pull Docker images after process registration is complete
```
Add this under `Changed`.

abukosek:
Oops, done!

PR #373, opened 2017-10-03
 Generalize jinja filters to accept lists of data objects

mzganec:


resolwe/flow/expression_engines/jinja/filters.py
```diff
@@ -23,16 +23,22 @@ def _get_data_attr(data, attr):
 
 def name(data):
     """Return `name` of `Data`."""
+    if isinstance(data, (list, tuple)):
```
dblenkus: Move the whole logic to a helper function, so you don't need to repeat it in each function.

kostko: I agree with @dblenkus as it is repeating quite a lot here. Perhaps have something like `apply_filter_list(value, lambda item: _get_data_attr(item, 'name'))` and then you can handle both list and non-list cases in a generic fashion.

resolwe/flow/expression_engines/jinja/filters.py
```diff
@@ -91,6 +91,23 @@ def descriptor(obj, path=''):
     return resp
 
 
+def apply_filter_list(func, objs):
```
dblenkus: Define this function before it is used.

mstajdohar:
Looks good, thanks.

PR #380, opened 2017-10-12
Fix entity collections filter

kostko:
~Note that this could be an issue for other filters as well and we don't seem to have any tests for filters.~


mstajdohar:
Review tests.

PR #382, opened 2017-10-13
 Fix elastic indices to run in parallel tests

dblenkus:


resolwe/elastic/builder.py
```diff
@@ -212,7 +214,13 @@ def discover_indexes(self):
                 for attr_name in dir(indexes_module):
                     attr = getattr(indexes_module, attr_name)
                     if inspect.isclass(attr) and issubclass(attr, BaseIndex) and attr is not BaseIndex:
-                        self.indexes.append(attr())
+                        kwargs = {}
+                        if getattr(settings, 'ELASTICSEARCH_TESTING', False):
+                            kwargs['_index'] = "{}_test_{}".format(
+                                attr.document_class._doc_type.index, os.getpid()  # pylint: disable=protected-access
```
kostko: Since you already access `attr.document_class._doc_type.index` directly here, what is the reason that you also don't update it directly here but in the constructor a few lines below instead?

PR #390, opened 2017-10-20
Make ProcessTestCase non-transactional

tjanez:
Remove `ProcessTestCase` that enclosed tests' code in a transaction and rename the non-transactional `TrasactionProcessTestCase` to `ProcessTestCase`.
Remove `ProcessTestCase.run_process`' `run_manager` argument since it no longer applies.
Update `test_executors.ManagerRunProcessTest.test_workflow` test to make it work with non-transactional `ProcessTestCase`.

### DIFF
--- a/resolwe/test/testcases/process.py
+++ b/resolwe/test/testcases/process.py
@@ -21,7 +21,6 @@ import zipfile
 from six.moves import filterfalse
 
 from django.conf import settings
-from django.contrib.contenttypes.models import ContentType
 from django.core import management
 from django.db import transaction
 from django.test import override_settings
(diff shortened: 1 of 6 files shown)


resolwe/test/testcases/process.py
```diff
@@ -312,6 +298,13 @@ def run_process(self, process_slug, input_={}, assert_status=Data.STATUS_DONE,
         :class:`~resolwe.flow.models.Data` object's status matches
         it after the process has finished.
 
+        .. note::
+
+            If you need to delay calling the manager, you must put the
+            desired code in a ``with transaction.atomic()`` block and
+            use ``assertStatus=Data.STATUS_RESOLVING`` since the process
```
kostko: Can we do this automatically - e.g. check if we are in a transaction or not?

tjanez: Good point. I've added code to check this automatically.

PR #391, opened 2017-10-23
Refactor test runner to make it more generic

kostko:


resolwe/test_helpers/test_runner.py
```diff
@@ -0,0 +1,205 @@
...
+                # Process re-spawn instructions to discover dependencies.
+                matches = SPAWN_PROCESS_REGEX.findall(program)
+                if matches:
+                    for match in matches:
+                        dependencies.setdefault(match, set()).add(slug)
+
+        return dependencies
+
+    def resolve_process_tags(self, files):
+        """Handle command."""
```
tjanez: This is probably not the correct docstring?

resolwe/test_helpers/test_runner.py
```diff
@@ -0,0 +1,205 @@
...
+            help="Only test changes against given Git commit reference"
+        )
+
+        parser.add_argument(
+            '--changes-file-types', dest='changes_file_types',
+            help="File which describes what kind of changes are available"
+        )
+
+    def run_tests(self, test_labels, **kwargs):
+        """Run tests."""
```
tjanez: Please, also document all parameters with Sphinx's reST fields notation.

Applies to all public functions.

PR #392, opened 2017-10-23
 Copy permissions from Collection and Entity when Data object is created

dblenkus:


resolwe/flow/tests/test_api.py
```diff
@@ -126,25 +132,61 @@ def test_public_create(self):
...
 
-    def test_create_entity(self):
-        collection = Collection.objects.create(name='Test collection', contributor=self.contributor)
-        process = Process.objects.create(
-            name='Entity process', contributor=self.contributor, flow_collection='test-schema'
-        )
-        assign_perm('view_collection', self.user, collection)
-        assign_perm('add_collection', self.user, collection)
-        assign_perm('view_process', self.user, process)
+    def test_inherite_permissions(self):
```
kostko: Extra `e`.

resolwe/permissions/tests/test_utils.py
```diff
@@ -31,6 +44,11 @@ def test_copy_permissions(self):
...
-    def test_copy_perms_wrong_ctype(self):
-        with self.assertRaises(AssertionError):
-            copy_permissions(self.src_process, self.collection)
+    def test_copy_different_ctype(self):
+        assign_perm('view_collection', self.contributor, self.collection)
+        assign_perm('add_collection', self.contributor, self.collection)
+
+        copy_permissions(self.collection, self.dst_process)
+
+        # Only 'view' is copied as process ha no 'add' permission.
```
kostko: Missing `s`.

resolwe/permissions/utils.py
```diff
@@ -21,18 +21,61 @@
...
+    """Keep permission action and change content type to ``dest_obj``."""
+    action = get_perm_action(perm)
+    return get_full_perm(action, dest_obj)
+
+
 def copy_permissions(src_obj, dest_obj):
     """Copy permissions form ``src_obj`` to ``dest_obj``."""
+    def _process_permission(codename, user_or_group, dest_obj):
+        """Process single permission."""
+        if relabel:
```
kostko: Does this implicitly come from the parent scope? If so, this should be changed, so that either `relabel` is defined above this function or (better) that relabel is a parameter.

PR #393, opened 2017-10-23
Move common test environment preparation to TestCaseHelpers mixin

tjanez:


resolwe/test/testcases/__init__.py
```diff
@@ -34,17 +34,132 @@
...
+        self._override_executor_settings()
+        self.settings.enable()
+
+        self._keep_data = False
+
+        self.addCleanup(self._clean_up)
+
+    def keep_data(self, mock_purge=True):
+        """Do not delete output files after test."""
+        self._keep_data = True
```
kostko: So this will be moved to the test runner later?

tjanez: Yes, that's the plan.

PR #394, opened 2017-10-25
Prepare release 4.0.0 and bump version to 4.1.0a1

tjanez:


docs/CHANGELOG.rst
```diff
@@ -44,17 +50,10 @@ Fixed
   process is finished
 - Remove unrelated permissions occasionally listed among group
   permissions on ``permissions`` endpoint
-- Fix ``ResolwePermissionsMixin`` to work correctly with multi-words 
-  model names, i.e. DescriptorSchema
-
-
-==================
-3.0.1 - 2017-09-15
```
dblenkus: Why was this deleted? You should also include change logs for all `3.x` versions.

docs/CHANGELOG.rst
```diff
@@ -44,17 +50,10 @@ Fixed
...
-==================
-
-Fixed
------
-- Correctly relabel SELinux contexts on user/group files
+- Fix ``ResolwePermissionsMixin`` to work correctly with multi-words
+  model names, i.e. ``DescriptorSchema``
+- Fix Elasticsearch indices to work when running tests in parallel
```
dblenkus: This is actually not true, as old version worked with Django 1.10, which was required by Resolwe. It was changed to keep the functionality when requirement was bumped to Django 1.11. But from user"s point of view nothing has changed.

docs/CHANGELOG.rst
```diff
@@ -44,17 +50,10 @@ Fixed
...
+- Fix Elasticsearch pagination error
```
dblenkus: Add a more descriptive entry.

tjanez: Done.

docs/CHANGELOG.rst
tjanez: Ok, I see. I've removed the CHANGELOG.

I would ask you to put this info to the corresponding commit message next time, as it helps understand why the change was made.

docs/CHANGELOG.rst
tjanez: I've reverted the change and added information about all `3.x` releases that happened in the mean time.

dblenkus: Thanks!

tjanez:
The Travis builds timed out due to [an issue they have with pulling Docker images from Docker Hub](https://twitter.com/traviscistatus/status/923123856948129792).

PR #395, opened 2017-11-03
Drop Python 2 support, require Python 3.4+

tjanez:
Use `python_requires` `setup()` argument to specify the required Python version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,6 +75,7 @@ setup(
         # have to pin it
         'django-filter~=1.0.0',
     ],
+    python_requires='>=3.4, <3.6',
     extras_require={
         'docs':  [
             'sphinx_rtd_theme',
@@ -99,10 +100,6 @@ setup(
             'tblib>=1.3.0',
             'isort',
         ],
-        ':python_version == "2.7"': [
-            # Backport of shutil.which function to Python 2
-            'shutilwhich',
-        ]
     },
 
     classifiers=[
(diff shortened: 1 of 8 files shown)


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,34,35},docs,linters,packaging,migrations
```
dblenkus: Will Tox use Python 3 for packaging, docs and linters envs? Or do you have to set `basepython`?

setup.py
```diff
@@ -75,6 +75,7 @@
         # have to pin it
         'django-filter~=1.0.0',
     ],
+    python_requires='~=3.4',
```
dblenkus: You should also allow Python 3.5

tjanez: According to the [official Python Packaging User Guide](https://packaging.python.org/tutorials/distributing-packages/#python-requires), this would allow Python 3.4, 3.5, 3.6, ... anything below 4.

As we've discussed online, I'll change this to `python_requires='>=3.4, <3.6'` since we only officially support Python 3.4 and 3.5 right now.

tox.ini
tjanez: Nice catch!

Since the introduction of `python_requires`, Tox actually fails to run in a *vanilla* setting, e.g. running `tox -e packaging` on a system with `/usr/bin/python` pointing to Python 2.

I'll fix this by adding:
```
# ensure we run the tests with the latest supported Python version
basepython = python3.5
```
to all these Tox environments.

PR #396, opened 2017-11-08
Data usage dashboard

JureZmrzlikar:


resolwe/flow/models/utils.py
```diff
@@ -339,45 +339,74 @@ def hydrate_input_uploads(input_, input_schema, hydrate_values=True):
...
+                total_size += get_dir_size(ref_path)
+            elif os.path.exists(ref_path):
+                raise ValidationError("Reference is not a file or directory ({})".format(ref_path))
+            else:
+                raise ValidationError("Reference does not exist ({})".format(ref_path))
+
+            # It is a common case that ``obj['file']`` is also contained in
+            # one of obj['ref']. In that case, we need to make sure that it's
+            # size is not counted twice:
+            if ref_path in obj_path:
```
dblenkus: Move this just after `for` and change it to: 
```
 if ref_path in obj_path:
    continue
```

resolwe/flow/models/utils.py
```diff
@@ -339,45 +339,74 @@ def hydrate_input_uploads(input_, input_schema, hydrate_values=True):
...
+    def get_refs_size(obj, obj_path):
+        """Caculate size of all references."""
+        total_size = 0
+        for ref in obj.get('refs', []):
+            ref_path = os.path.join(settings.FLOW_EXECUTOR['DATA_DIR'], str(data.pk), ref)
+            if os.path.isfile(ref_path):
+                total_size += os.path.getsize(ref_path)
+            elif os.path.isdir(ref_path):
+                total_size += get_dir_size(ref_path)
+            elif os.path.exists(ref_path):
```
dblenkus: Move this and the `else` to the validation function: https://github.com/genialis/resolwe/blob/master/resolwe/flow/models/utils.py#L105-L129

resolwe/flow/models/utils.py
```diff
@@ -339,45 +339,74 @@ def hydrate_input_uploads(input_, input_schema, hydrate_values=True):
...
-        if data.status in [Data.STATUS_DONE, Data.STATUS_ERROR] and 'size' in obj:
+        if data.status in [Data.STATUS_DONE, Data.STATUS_ERROR] and 'size' in obj and not force:
             return
 
         path = os.path.join(settings.FLOW_EXECUTOR['DATA_DIR'], str(data.pk), obj['dir'])
         if not os.path.isdir(path):
             raise ValidationError("Referenced dir does not exist ({})".format(path))
 
         obj['size'] = get_dir_size(path)
+        obj['total_size'] = obj['size'] + get_refs_size(obj, path)
```
dblenkus: Other option would be to omit `obj['size']` here and calculate size of whole reference.

JureZmrzlikar: Yes, i was thinking about that. I assumed that `total_size` will be used more often than "ref_size", so it makes more sense to compute it directly, instead of having to sum up `size` and "ref_size" to get it.

resolwe/flow/migrations/0008_compute_size.py
```diff
@@ -0,0 +1,30 @@
...
+def calculate_size(apps, schema_editor):
+    """Compute ``size`` attribute in Data model."""
+    Data = apps.get_model('flow', 'Data')
+    for data in Data.objects.all():
+        hydrate_size(data, force=True)
+        data.save()
+
+
+def reverse_migration(apps, schema_editor):
+    pass
```
dblenkus: Add a comment that this can be skipped as the field will be deleted in the next step, i.e. previous migration.

resolwe/flow/models/utils.py
```diff
@@ -112,8 +113,10 @@ def validate_file(field, regex):
 
         if path_prefix:
             path = os.path.join(path_prefix, filename)
-            if not os.path.isfile(path):
+            if not os.path.exists(path):
                 raise ValidationError("Referenced file ({}) does not exist".format(path))
+            if not os.path.isfile(path):
+                raise ValidationError("Referenced file ({}) is not a normal file".format(path))
```
dblenkus: Do we want to enforce that? I think it should be ok to reference a link, even though i don't see a use-case for this.

resolwe/flow/models/utils.py
```diff
@@ -123,8 +126,10 @@ def validate_dir(field):
 
         if path_prefix:
             path = os.path.join(path_prefix, dirname)
-            if not os.path.isdir(path):
+            if not os.path.exists(path):
                 raise ValidationError("Referenced dir ({}) does not exist".format(path))
+            if not os.path.isdir(path):
+                raise ValidationError("Referenced dir ({}) is not a directory".format(path))
```
dblenkus: `Referenced dir` -> `Referenced path`

resolwe/flow/models/utils.py
```diff
@@ -339,45 +344,72 @@ def hydrate_input_uploads(input_, input_schema, hydrate_values=True):
...
     """
     from .data import Data  # prevent circular import
 
+    def get_dir_size(path):
+        """Get directory size."""
+        total_size = 0
+        for dirpath, _, filenames in os.walk(path):
+            for file_name in filenames:
+                file_path = os.path.join(dirpath, file_name)
+                if not os.path.isfile(file_path):  # Skip (broken) links
```
dblenkus: It would be clearer to use [islink](https://docs.python.org/3/library/os.path.html#os.path.islink).

JureZmrzlikar: Actually, i understood we wish to calculate sizes only for normal files. Any other structure (link is just an example that i bumped into) is potentially problematic. For example, links will point to certain location where the actual file is stored (and file size should be counted there, not here). Should i juts fix comment to make it more clear why file-like-ness is checked?

mstajdohar: > Should i juts fix comment to make it more clear why file-like-ness is checked?

Yes, thanks.

JureZmrzlikar:
@dblenkus , I think i've adressed all comments, do we have bandwidth to move this forward?

mstajdohar:
This should wait until after the next stable release.

mstajdohar:
We are ready to merge this if @dblenkus approves.

mstajdohar:
@JureZmrzlikar Rebase

JureZmrzlikar:
Rebased.

PR #398, opened 2017-11-15
Correctly handle removed processes

kostko:


resolwe/test_helpers/test_runner.py
```diff
@@ -365,18 +366,22 @@ def run_tests(self, test_labels, **kwargs):
...
-
-                top_level_path = subprocess.check_output(['git', 'rev-parse', '--show-toplevel'])
-                top_level_path = top_level_path.decode('utf8').strip()
-            except subprocess.CalledProcessError:
-                print("ERROR: Git command failed.")
+            # Check if there are changed files. We need to be able to switch between branches
+            # so we can correctly detect changes.
+            repo_status = self._git('status', '--porcelain')
+            if repo_status:
+                print("ERROR: Git repository is not clean. Running tests with --only-changes-to")
```
tjanez: As we've discussed in person, print errors to stderr (everywhere).

resolwe/test_helpers/test_runner.py
```diff
@@ -393,8 +398,31 @@ def run_tests(self, test_labels, **kwargs):
                 self.tags = tags
                 test_labels = tests
 
+                print("Running with following partial tags: {}".format(', '.join(tags)))
+                print("Running with following partial tests: {}".format(', '.join(tests)))
+
         return super().run_tests(test_labels, **kwargs)
 
+    def _git(self, *args):
```
tjanez: Did you consider using any of the Python git libraries?

I'm not saying you should, just interested in your thoughts 🙂 .

resolwe/test_helpers/test_runner.py
```diff
@@ -515,17 +543,41 @@ def resolve_process_tags(self, files):
         for proc_path in processes_paths:
             process_schemas.extend(self.find_schemas(proc_path))
 
+        # Switch to source branch and get all the schemas from there as well, since
+        # some schemas might have been removed.
+        with self.git_branch(self.only_changes_to):
+            for proc_path in processes_paths:
+                process_schemas.extend(self.find_schemas(proc_path))
```
tjanez: Would it make sense to check if the process schema is not already included in `process_schemas`?
It would probably often halve the size of the list.

resolwe/test_helpers/test_runner.py
```diff
@@ -365,18 +366,22 @@ def run_tests(self, test_labels, **kwargs):
...
-            try:
-                changed_files = subprocess.check_output(['git', 'diff', '--name-only', self.only_changes_to, 'HEAD'])
-                changed_files = changed_files.decode('utf8').strip().split('\n')
-                changed_files = [file for file in changed_files if file.strip()]
-
-                top_level_path = subprocess.check_output(['git', 'rev-parse', '--show-toplevel'])
-                top_level_path = top_level_path.decode('utf8').strip()
-            except subprocess.CalledProcessError:
-                print("ERROR: Git command failed.")
+            # Check if there are changed files. We need to be able to switch between branches
```
tjanez: Can you wrap the comments at the same wrap limit as the surrounding code?
Here it would probably at 100 characters.

resolwe/test_helpers/test_runner.py
```diff
@@ -393,8 +398,31 @@ def run_tests(self, test_labels, **kwargs):
                 self.tags = tags
                 test_labels = tests
 
+                print("Running with following partial tags: {}".format(', '.join(tags)))
```
tjanez: Nice addition!

resolwe/test_helpers/test_runner.py
```diff
@@ -515,17 +543,41 @@ def resolve_process_tags(self, files):
         for proc_path in processes_paths:
             process_schemas.extend(self.find_schemas(proc_path))
 
+        # Switch to source branch and get all the schemas from there as well, since
```
tjanez: Same as above, can you wrap the comments at the same wrap limit as the surrounding code?

resolwe/test_helpers/test_runner.py
```diff
@@ -515,17 +543,41 @@ def resolve_process_tags(self, files):
...
 
                 for process in data:
                     # Add all process slugs.
                     processes.add(process['slug'])
 
+        for filename in files:
+            try:
+                load_process_slugs(filename)
+            except FileNotFoundError:
+                # File was removed, so we will handle it below when we check the
```
tjanez: Same as above, can you wrap the comments at the same wrap limit as the surrounding code?

resolwe/test_helpers/test_runner.py
kostko: No, because the schema might exist in both cases but be different. So it needs to be indexed twice to cover all possible dependencies. We could check if it is the same schema that was already indexed before parsing it, but this would just complicate the logic for little gain (schema parsing is not really that expensive).

tjanez: Ok, let's leave it as is then.

PR #399, opened 2017-11-16
Add process schema migration operations

kostko:
TODO:
* [x] Check all the information that default generator classes need (at least access to `from_state`).
* [x] Add migration history.
* [x] Add "purge temporary data objects" operation.
* [x] Write an example migration for `species` field in Resolwe Bio.

The idea is for these operations to be used as follows in migrations:
```python
# -*- coding: utf-8 -*-
# Generated by Django 1.11.6 on 2017-11-16 05:51
from __future__ import unicode_literals

from django.db import migrations

from resolwe.flow.migration_ops import ResolweProcessAddField, DataDefaultOperation
from resolwe.flow.utils import dict_dot


class SpeciesDataDefault(DataDefaultOperation):
    """Default generator for species."""

    def __init__(self, source, feature_storage, feature_ids):
        """
        :param source: Field path for field containing feature source
        :param feature_storage: Field path for storage field
        :param feature_ids: Field path for field inside storage containing
            feature ids
        """
        self.source = source
        self.feature_storage = feature_storage
        self.feature_ids = feature_ids

    def prepare(self, data, from_state):
        """Prepare data object migration.

        Called after all relevant process schemas have already
        been migrated.

        :param data: Queryset containing all data objects that need
            to be migrated
        :param from_state: Database model state
        """
        pass

    def get_default_for(self, data, from_state):
        """Return default for given data object.

        :param data: Data object instance
        :param from_state: Database model state
        """
        Storage = from_state.apps.get_model('flow', 'Storage')
        Feature = from_state.apps.get_model('resolwe_bio_kb', 'Feature')

        source = dict_dot(data, self.source)
        feature_storage = Storage.objects.get(pk=dict_dot(data, self.feature_storage))
        feature_ids = dict_dot(feature_storage.json, self.feature_ids).keys()
        try:
            return Feature.objects.filter(source=source, feature_id__in=feature_ids)[0].species
        except IndexError:
            raise ValueError("Unable to determine species for data object {}".format(data.pk))


class Migration(migrations.Migration):

    dependencies = [
        ('resolwe_bio', '0010_add_relation_types'),
    ]

    operations = [
        ResolweProcessAddField(
            process='upload-expression',
            field='input.species',
            schema={
                'name': 'species',
                'label': 'Species',
                'type': 'basic:string:',
            },
            default=SpeciesDataDefault(
                source='input.source',
                feature_storage='output.exp_json',
                feature_ids='genes',
            ),
        )
    ]
```


resolwe/flow/migration_ops.py
```diff
@@ -0,0 +1,351 @@
...
+        :param schema_type: Type of schema (must be either ``input`` or
+            ``output``)
+        """
+
+        self.process = process
+        self.schema_type = schema_type
+        if self.schema_type not in ('input', 'output'):
+            raise ValueError("Schema type must be either input or output")
+
+        # Unfortunately, we can't easily get the identifier of the current
```
kostko: I don't like this very much, but I don't see another way as there is no method that gets called with the current migration instance and Django's migration log is only updated after the migration has been applied.

kostko:
Migration log has been added.

resolwe/flow/migration_ops.py
```diff
@@ -0,0 +1,352 @@
...
+                'schema': self.schema,
+                'default': self.default,
+            }
+        )
+
+    def migrate_process_schema(self, process, schema, from_state):
+        """Migrate process schema.
+
+        :param process: Process instance
+        :param schema: Process schema to migrate
```
JureZmrzlikar: Do you think it would make sense to emphasize the difference between `schema`(process schema) parameter and `self.schema`(added field schema)? Rename one variable or maybe add a comment explaining the difference. Just an idea, you decide if it makes sense.

kostko:
@JureZmrzlikar as discussed, I have now also added `ResolweDataRemoveTemporary` migration operation, which removes all data objects from processes which use `TMP` persistence (so you don't need to write migrations for those).

resolwe/flow/migration_ops.py
```diff
@@ -0,0 +1,373 @@
...
+    def deconstruct(self):
+        """Deconstruct operation."""
+        return (
+            self.__class__.__name__,
+            [],
+            {
+                'process': self.process,
+                'field': self._raw_field,
+                'schema': self.schema,
+                'default': self.default,
```
dblenkus: `schema_type` is used in `ResolweProcessOperation.__init__`. Probably you have to add it here?

resolwe/flow/migration_ops.py
```diff
@@ -0,0 +1,373 @@
...
+        :param from_state: Database model state
+        :return: True if the process was migrated, False otherwise
+        """
+        container = dict_dot(schema, '.'.join(self.field[:-1]))
+        assert container
+
+        # Ignore processes, which already contain the target field.
+        for field in container:
+            if field['name'] == self.field[-1]:
+                return False
```
dblenkus: If field already exists, this will cause an inconsistent state. Maybe it would be better to raise an error?

resolwe/flow/migration_ops.py
kostko: No, because in this case it is computed by the constructor from `field`.

resolwe/flow/migration_ops.py
kostko: :+1:

kostko: Oh, now I remember what the issue is here. There can be multiple process versions and the question is which ones to migrate? The latest version may already contain the given field, so it should probably be skipped? This is why it is currently made to skip by default.

How should we handle this? Error on all except latest if they already have the field and error on latest if it doesn't have the field?

dblenkus: Can you raise error only if schemas don't match?

kostko:
I've added process type consistency checks after each migration. Note that these checks currently fail even for master (@jkokosar is working on a fix for those).

mstajdohar:
> I've added process type consistency checks after each migration.
👍

mstajdohar:
This is ready for review, @dblenkus, @tjanez.

mstajdohar:
@dblenkus Please review

resolwe/flow/migration_ops.py
```diff
@@ -0,0 +1,656 @@
...
+    def migrate_process_schema(self, process, schema, from_state):
+        """Migrate process schema.
+
+        :param process: Process instance
+        :param schema: Process schema to migrate
+        :param from_state: Database model state
+        :return: True if the process was migrated, False otherwise
+        """
+        container = dict_dot(schema, '.'.join(self.field[:-1]))
+        assert container
```
JureZmrzlikar: There actually is a process (`bam-split`) that does not have outputs defined (yet). Consequently, this assert fails. Could we skip this?

mstajdohar: @kostko What is the point of this assert? Should we test that container is not None, `assert container is None`? The problem is output fields are not required in processes.

kostko: I see. So if the container is not defined, we'll need to create the container in this case. I'll push a fix.

kostko: Should be fixed now, @JureZmrzlikar please check.

resolwe/flow/models/migrations.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __str__(self):
+        """Format model name."""
+        return self.migration
+
+
+class ProcessMigrationHistory(MigrationHistoryBase):
+    """Model for storing process migration history."""
+
+    process = models.ForeignKey(Process, related_name='migration_history')
```
dblenkus: Add `on_delete=models.CASCADE`, as it is required in Django 2.0

resolwe/flow/models/migrations.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class DataMigrationHistory(MigrationHistoryBase):
+    """Model for storing data migration history."""
+
+    data = models.ForeignKey(Data, related_name='migration_history')
```
dblenkus: And here.

resolwe/flow/models/migrations.py
tjanez: @dblenkus, @kostko, is this comment still relevant?

dblenkus: Yes, this will have to be fixed at some point.

PR #400, opened 2017-11-23
Reorder Data spawning and finalizing in the listener

jberci:
Some events are fired when `Data` objects get their `DONE` status, and these events already need to see spawned children. The changeset in this PR reorders operations in the Resolwe listener so that spawning doesn't happen after finalising anymore.


tjanez:
@jberci, one more thing. Can you include the explanation you put in this PR's description in the commit message's body?

PR #401, opened 2017-11-27
Add database-side JSON projections for Storage models

kostko:
TODO
* [x] Tests

### DIFF
--- a/resolwe/flow/models/storage.py
+++ b/resolwe/flow/models/storage.py
@@ -5,6 +5,35 @@ from django.contrib.postgres.fields import JSONField
 from django.db import models
 
 from .base import BaseModel
+from .functions import JsonGetPath
+from .utils import json_path_components
+
+
+class StorageManager(models.Manager):
+    """Manager for Storage objects."""
+
+    def with_json_path(self, path, field=None):
+        """Annotate Storage objects with a specific JSON path.
+
+        :param path: Path to get inside the stored object, which can be
+            either a list of path components or a comma-separated
+            string
+        :param field: Optional output field name
+        """
+        if field is None:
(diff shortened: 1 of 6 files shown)


resolwe/flow/models/storage.py
```diff
@@ -5,6 +5,31 @@
 from django.db import models
 
 from .base import BaseModel
+from .functions import JsonGetPath
+
+
+class StorageManager(models.Manager):
+    """Manager for Storage objects."""
+
+    def with_json_path(self, path, field='result'):
```
dblenkus: It would be more useful if `field` is generated from `path`, so you can call method multiple times (without manually defining `field`).

resolwe/flow/models/storage.py
```diff
@@ -5,6 +5,31 @@
...
+        """Annotate Storage objects with a specific JSON path.
+
+        :param path: Path to get inside the stored object, which can be
+            either a list of path components or a comma-separated
```
dblenkus: Dot-separated string should be also supported, to make it similar to `dict_dot`.

resolwe/flow/models/storage.py
kostko: I thought about it, but the question is how to make this the most intuitive. Perhaps we could name the resulting field something like `'_'.join(['json'] + path)`? The only problem with this is that currently the path parsing logic is inside the projection function, while this would require path processing here.

resolwe/flow/models/storage.py
kostko: We could also just drop everything and support just lists/tuples of path components. I don't actually like these different formats here and it would also make reasoning about paths easier (e.g. as mentioned in the comment above).

kostko:
Ready for re-review.

PR #403, opened 2017-12-04
Multiple test runner fixes

kostko:


resolwe/test/testcases/process.py
```diff
@@ -240,8 +240,23 @@ def tearDown(self):
...
+        # Adapted from: https://stackoverflow.com/a/39606065
+        def list2reason(exc_list):
+            """Error reason conversion helper."""
+            if exc_list and exc_list[-1][0] is self:
+                return exc_list[-1][1]
+
+        if hasattr(self, '_outcome'):  # Python 3.4+
+            result = self.defaultTestResult()  # these 2 methods have no side effects
+            self._feedErrorsToResult(result, self._outcome.errors)
+        else:  # Python 3.2 - 3.3 or 3.0 - 3.1 and 2.7
```
kostko: We could easily drop this part as we only support Python 3.4+?

resolwe/test_helpers/test_runner.py
```diff
@@ -329,6 +329,15 @@ def add_arguments(cls, parser):
             help="Do not mock purging functions (implies --keep-data)"
         )
 
+    def build_suite(self, *args, **kwargs):
+        """Build test suite."""
+        suite = super().build_suite(*args, **kwargs)
+        # Build suite first constructs the parallel suite and then may reduce self.parallel,
```
tjanez: Can you report this Django test runner's bug upstream and extend this note to explain that this is a fix for this bug (also include upstream's bug number)?

Also add a note to the CHANGELOG since this impacts Resolwe's users (e.g. Resolwe Bio developers).

resolwe/test_helpers/test_runner.py
```diff
@@ -416,14 +416,18 @@ def run_tests(self, test_labels, **kwargs):
     def _git(self, *args):
         """Helper to run Git command."""
         try:
-            return subprocess.check_output(['git'] + list(args)).strip()
+            return subprocess.check_output(['git'] + list(args)).decode('utf8').strip()
```
tjanez: I would move the `decode('utf-8')` handling change into a separate commit.

resolwe/test_helpers/test_runner.py
```diff
@@ -416,14 +416,18 @@ def run_tests(self, test_labels, **kwargs):
...
         except subprocess.CalledProcessError:
             raise CommandError("Git command failed.")
 
     @contextlib.contextmanager
     def git_branch(self, branch):
         """Temporarily switch to a different Git branch."""
         current_branch = self._git('rev-parse', '--abbrev-ref', 'HEAD')
+        if current_branch == 'HEAD':
+            # Detached HEAD state, we need to get the actual commit.
+            current_branch = self._git('rev-parse', 'HEAD')
```
tjanez: Looks good!

Also mention this in the CHANGELOG since this impacts Resolwe users that run the test suite where HEAD is in a detached state.

resolwe/test/testcases/process.py
mstajdohar: 👍

resolwe/test_helpers/test_runner.py
```diff
@@ -405,12 +405,24 @@ def run_tests(self, test_labels, **kwargs):
...
+                    self.tags = tags
+                    super().run_tests(test_labels, **kwargs)
+
+                # Then run with specific test labels.
+                if tests:
+                    self.parallel = parallel
+                    self.tags = set()
+                    super().run_tests(tests, **kwargs)
+
+                return
```
tjanez: What do `super().run_tests()` calls return? Do we need to propagate this here?

resolwe/test_helpers/test_runner.py
```diff
@@ -405,12 +405,24 @@ def run_tests(self, test_labels, **kwargs):
                 print("Non-test code or unknown files have been modified, running full test suite.")
             else:
                 # Run specific tests/tags.
-                self.tags = tags
-                test_labels = tests
-
                 print("Running with following partial tags: {}".format(', '.join(tags)))
                 print("Running with following partial tests: {}".format(', '.join(tests)))
 
+                # First run with specific tags. Since run_tests may modify self.parallel, we need to store
```
tjanez: Also note this change in the CHANGELOG for similar reasons as I wrote before.

resolwe/test/testcases/process.py
```diff
@@ -240,8 +240,23 @@ def tearDown(self):
 
         super().tearDown()
 
+        # Check test outcome to prevent failing the test twice.
```
tjanez: 🙂 , I was thinking of the same thing today when I watched the output of process failures.

Please, add a CHANGELOG entry. Perhaps a more succinct explanation would be something like:
Do not report additional failure when testing a tagged process errors or fails.

resolwe/test/testcases/process.py
tjanez: Yes, drop this part.

PR #404, opened 2017-12-04
Address missed review comments

kostko:
See #403


mstajdohar:
Thanks @tjanez for a thorough review.

PR #406, opened 2017-12-05
Add a way to pass secrets to processes in a controlled fashion

kostko:
**DEPLOYMENT CONSIDERATIONS:**
* Secrets are encrypted using `SECRET_KEY`, keep it secure and backed up as it should already be ;-)
* Setting `FLOW_DOCKER_MAPPINGS` has been removed in favor of new `FLOW_DOCKER_VOLUME_EXTRA_OPTIONS` and `FLOW_DOCKER_EXTRA_VOLUMES`.


resolwe/flow/models/secret.py
```diff
@@ -0,0 +1,57 @@
...
+            given contributor will be looked up).
+        """
+        queryset = self.all()
+        if contributor is not None:
+            queryset = queryset.filter(contributor=contributor)
+        secret = queryset.get(handle=handle)
+        return secret.value
+
+
+class Secret(models.Model):
```
dblenkus: I think we will need an extra to describe where the secret comes from. E.g. if you get users API token first time, you don't have to ask for it in following requests.

resolwe/flow/tests/processes/secrets.yml
```diff
@@ -0,0 +1,33 @@
...
+      label: Secret token
+      type: basic:secret
+  output:
+    - name: secret
+      label: Secret value
+      type: basic:string
+  run:
+    program: |
+      # Echo secret to output. NEVER do this in an actual process!
+      re-save secret "$(cat /secrets/{{token.handle}})"
```
dblenkus: It would be useful to add Jinja filter for this, i.e. `{{ token | get_secret }}`

resolwe/flow/managers/base.py
```diff
@@ -353,15 +356,29 @@ def _prepare_context(self, data_id, data_dir, runtime_dir, **kwargs):
         }
 
         # extend the settings with whatever the executor wants
-        self.executor.extend_settings(data_id, files)
+        self.executor.extend_settings(data_id, files, extra)
```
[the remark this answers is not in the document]
jberci: On the other hand, the docstring for `extended_settings` specifies this plainly as a field for general raw files (as opposed to JSON serializations in the `files` parameter). Secrets are just one kind of opaque blob, so usage agrees with the docstring's specification.

resolwe/flow/models/secret.py
```diff
@@ -0,0 +1,57 @@
...
+        """Create a new secret, returning its handle.
+
+        :param value: Secret value to store
+        :param contributor: User owning the secret
+        :return: Secret handle
+        """
+        secret = self.create(value=value, contributor=contributor)
+        return str(secret.handle)
+
+    def get_secret(self, handle, contributor=None):
```
dblenkus: Is there a use-case where contributor is not defined? If not i would prefer to make it mandatory.

resolwe/flow/managers/base.py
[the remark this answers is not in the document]
dblenkus: But we also prevent listing of such dirs and restrict masks of files. So it may have unwanted consequences if you put such file directly into executor dir.
I would prefer to narrow use of this, make it more securely (it already is) and clearly state that.

resolwe/flow/executors/docker/run.py
```diff
@@ -118,6 +118,13 @@ def start(self):
         mappings.append({'src': passwd_file.name, 'dest': '/etc/passwd', 'mode': 'ro,Z'})
         mappings.append({'src': group_file.name, 'dest': '/etc/group', 'mode': 'ro,Z'})
 
+        # Mount secrets.
+        mappings.append({
+            'src': os.path.join(SETTINGS['FLOW_EXECUTOR']['RUNTIME_DIR'], str(self.data_id), 'secrets'),
+            'dest': '/secrets',
+            'mode': 'ro,Z'
```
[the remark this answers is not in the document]
kostko: Hm, but we are using this also above for mounting passwd/group?

resolwe/flow/executors/prepare.py
```diff
@@ -26,17 +26,22 @@ def extend_settings(self, data_id, files):
             filenames, values are the objects that will be serialized
             into those files. Standard filenames are listed in
             :class:`resolwe.flow.managers.protocol.ExecutorFiles`.
+        :param extra: Extra raw files dictionary describing additional raw
+            file content that should be created. Keys are filenames, values
+            are the strings that should be written into those files.
         """
+        data = Data.objects.get(pk=data_id)
```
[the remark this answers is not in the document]
kostko: As we are getting just a single object this doesn't make any difference in performance or the number of queries as even prefetching requires an additional query. We could use `select_related` here instead of prefetch to combine everything into a single join.

resolwe/flow/models/secret.py
kostko: We can add an arbitrary JSON metadata column for that?

resolwe/flow/executors/docker/run.py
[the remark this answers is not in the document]
dblenkus: Yes, those files are normally on local disk so it is ok. The difference here is that runtime dir can be on shared storage.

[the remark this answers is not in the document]
kostko: Ok, I will move this into configuration as we have for data/upload dirs.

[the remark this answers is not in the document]
dblenkus: But this means that you have to support different mounting points in the container.

What if we add a setting to enable/disable SELinux config, e.g. `FLOW_ENABLE_SELINUX_CONF`, and the you can construct mount here?

cc @tjanez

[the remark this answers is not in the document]
kostko: But then we must change this everywhere as this is very strange to have some modes in settings and others are derived from this?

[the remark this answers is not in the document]
kostko: > But this means that you have to support different mounting points in the container.

I don't think we should support this even with existing data/upload etc. endpoints. They should be fixed.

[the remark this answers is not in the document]
kostko: I've pushed a version which moves this into settings (I don't like this though). We can move this back if this makes sense.

resolwe/flow/models/secret.py
kostko: Done.

resolwe/flow/models/secret.py
kostko: :+1:

resolwe/flow/static/flow/typeSchema.json
```diff
@@ -113,6 +114,27 @@
...
+        "secret": {
+            "type": "object",
+            "required": ["type", "value"],
+            "additionalProperties": false,
+            "properties": {
+                "type": {
+                    "type": "string",
+                    "pattern": "^basic:secret:$"
+                },
+                "value": {
```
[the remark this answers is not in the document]
kostko: Done.

resolwe/flow/executors/docker/run.py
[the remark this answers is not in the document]
dblenkus: >But then we must change this everywhere 

Yes.

Actually we support this for data/upload dir:
https://github.com/genialis/resolwe/blob/ada2c06c826fad06a9a8181f4b4c3527db72dc19/resolwe/flow/executors/docker/run.py#L158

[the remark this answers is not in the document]
kostko: I know, but this (doing search and replace in arbitrary scripts) is a very ugly hack that should be removed.

kostko:
Now also pushed removal of `FLOW_DOCKER_MAPPINGS`.

resolwe/flow/tests/processes/secrets.yml
kostko: For which part? The whole command is specific for bash. Or did you mean just the `{{token.handle}}`?

resolwe/flow/executors/docker/run.py
```diff
@@ -153,9 +184,9 @@ def start(self):
 
     def run_script(self, script):
         """Execute the script and save results."""
-        mappings = SETTINGS.get('FLOW_DOCKER_MAPPINGS', {})
-        for map_ in mappings:
-            script = script.replace(map_['src'], map_['dest'])
+        # XXX: This is a huge hack and should be removed.
```
dblenkus: Can we mount everything on the same path as it is outside?

resolwe/flow/tests/processes/secrets.yml
dblenkus: I mean that you write `{{ token | get_secret }}` in the process and it is replaced with `$(cat /secrets/{{token.handle}})` by Jinja filter.

That would make it much more user friendly.

resolwe/flow/models/secret.py
```diff
@@ -0,0 +1,74 @@
...
+class SecretManager(models.Manager):
+    """Manager for Secret objects."""
+
+    def create_secret(self, value, contributor, metadata=None, expires=None):
+        """Create a new secret, returning its handle.
+
+        :param value: Secret value to store
+        :param contributor: User owning the secret
+        :param metadata: Optional metadata dictionary (must be JSON serializable)
+        :param expires: Optional expiry time
```
hadalin: Maybe add units for `expires`?

hadalin: Also maybe state it defaults to `None`.

resolwe/flow/executors/docker/run.py
kostko: I don't think this is the correct way to go as it is much cleaner to have fixed volumes inside the container, regardless of what is outside. In this way you can always rely on specific paths when you run inside Docker.

It would be much better to always defer path translation to the executor (actually to the prepare part of the executor) as it is executor specific (e.g., the local executor does nothing, the Docker executor uses fixed volumes, etc.). Not just blindly using `*_DIR` from settings.

resolwe/flow/models/secret.py
kostko: :+1:

kostko:
@jberci please review my last commit, which fixes some issues with incorrect singleton use in Channels workers, which got exposed when removing hacky path translation.

resolwe/flow/executors/docker/run.py
kostko: This hack has been removed in the latest commit.

resolwe/flow/managers/base.py
```diff
@@ -211,7 +218,18 @@ def __init__(self, *args, **kwargs):
         self._drop_ctypes = getattr(settings, 'FLOW_MANAGER_DISABLE_CTYPE_CACHE', False)
 
         if 'static' not in kwargs or not kwargs['static']:
-            super().__init__(*args, **kwargs)
+            # Update global singleton to point to the current instance. Otherwise,
+            # the global instance will be completely incorrect in Channels workers.
+            from resolwe.flow.managers import override_manager
+
+            with override_manager(self):
```
kostko: @jberci can this be called in parallel in case of multi-threaded Channels workers? In this case, the global manager singleton is not thread-safe and this new override I introduced is problematic.

kostko: Oh, it seems that Channels workers are single-threaded.

jberci:
LGTM so far. The global manager instance patching is still a bit hacky on the whole, but the Channel handler proxy will probably take care of that, since the manager will then be constructed once, as it was pre-channelization.

kostko:
@jberci I've now pushed an updated version of the "Remove hacky path translation in Docker executor" commit, which separates the Channels consumer from the manager and uses the global manager instance instead.

resolwe/flow/models/secret.py
```diff
@@ -0,0 +1,75 @@
...
+
+from django.conf import settings
+from django.contrib.postgres.fields import JSONField
+from django.db import models
+
+
+class SecretManager(models.Manager):
+    """Manager for Secret objects."""
+
+    def create_secret(self, value, contributor, metadata=None, expires=None):
```
hadalin: Should this method always be used when creating secrets? If yes, then update docs and maybe also raise error when using the usual `.create(`. In this case it would also be nicer to return `handle, secret` to remove the need to query the secret by its handle if you intend to use the secret directly.

Alternatively if `.create` is allowed, then override it and call it in `create_secret`.

hadalin: Both ways are ok I guess, but prevent confusion.

resolwe/flow/models/secret.py
```diff
@@ -0,0 +1,75 @@
...
+    contributor = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.PROTECT)
+
+    #: secret handle
+    handle = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    #: actual secret
+    value = EncryptedTextField()
+
+    #: secret metadata (not encrypted)
+    metadata = JSONField(default=dict)
```
hadalin: Since this is arbitrary data, should a naming convention be followed to prevent conflicts?

resolwe/flow/executors/docker/run.py
```diff
@@ -94,11 +97,39 @@ def start(self):
...
         command_args['security'] = ' '.join(security)
 
-        # render Docker mappings in FLOW_DOCKER_MAPPINGS setting
-        mappings_template = SETTINGS.get('FLOW_DOCKER_MAPPINGS', [])
-        context = {'data_id': self.data_id}
-        mappings = [{key.format(**context): value.format(**context) for key, value in template.items()}
-                    for template in mappings_template]
+        # Setup Docker volumes.
+        def add_volume(volumes, kind, base_dir_name, volume, path=None, read_only=True):
+            """Add a new volume to the volumes list."""
```
dblenkus: Add description of arguments.

resolwe/flow/executors/docker/run.py
```diff
@@ -115,26 +146,26 @@ def start(self):
         group_file.file.flush()
         self.temporary_files.append(group_file)
 
-        mappings.append({'src': passwd_file.name, 'dest': '/etc/passwd', 'mode': 'ro,Z'})
-        mappings.append({'src': group_file.name, 'dest': '/etc/group', 'mode': 'ro,Z'})
+        volumes.append({'src': passwd_file.name, 'dest': '/etc/passwd', 'options': 'ro,Z'})
```
dblenkus: It would be nice to make these options configurable.

resolwe/flow/executors/docker/run.py
```diff
@@ -115,26 +146,26 @@ def start(self):
...
+        volumes.append({'src': group_file.name, 'dest': '/etc/group', 'options': 'ro,Z'})
 
-        # create mappings for tools
+        # Create volumes for tools.
         # NOTE: To prevent processes tampering with tools, all tools are mounted read-only
         # NOTE: Since the tools are shared among all containers they must use the shared SELinux
         # label (z option)
-        self.mappings_tools = [{'src': tool, 'dest': '/usr/local/bin/resolwe/{}'.format(i), 'mode': 'ro,z'}
+        self.mappings_tools = [{'src': tool, 'dest': '/usr/local/bin/resolwe/{}'.format(i), 'options': 'ro,z'}
```
dblenkus: This SELinux label will also be a problem on a shared storage.

resolwe/flow/executors/docker/run.py
dblenkus: `self.volumes`? Or move `self.volumes = volumes.copy()` to the end of the function.

resolwe/test/testcases/setting_overrides.py
```diff
@@ -11,39 +11,6 @@
 TEST_SETTINGS_OVERRIDES = getattr(settings, 'FLOW_EXECUTOR', {}).get('TEST', {})
 FLOW_EXECUTOR_SETTINGS.update(TEST_SETTINGS_OVERRIDES)
 
-
-# update FLOW_DOCKER_MAPPINGS setting if necessary
-def _get_updated_docker_mappings(previous_settings=getattr(settings, 'FLOW_EXECUTOR', {})):
```
dblenkus: Nice :)

resolwe/flow/execution_engines/bash/__init__.py
```diff
@@ -35,7 +35,7 @@ def evaluate(self, data):
             # Include special 'proc' variable in the context.
             inputs['proc'] = {
                 'data_id': data.id,
-                'data_dir': settings.FLOW_EXECUTOR['DATA_DIR'],
+                'data_dir': self.manager.get_executor().resolve_data_path(data, '..'),
```
dblenkus: Maybe add dedicated function for this instead of passing `..` as filename?

resolwe/flow/managers/base.py
```diff
@@ -88,7 +87,7 @@ def dependency_status(data):
     return Data.STATUS_DONE
 
 
-class BaseManager(BaseConsumer):
+class BaseManager(object):
```
dblenkus: This should go into a separate commit.

resolwe/flow/executors/docker/run.py
kostko: `self.volumes` is no longer needed now that we got rid of hacky path translation, I forgot to remove it.

resolwe/flow/executors/docker/run.py
kostko: I left this as it was before? So this was already a problem before?

resolwe/flow/executors/docker/run.py
kostko: This is an internal helper, but ok, if you think that it should be added ;-)

resolwe/flow/executors/docker/run.py
dblenkus: Yes, it was already a problem before, I just pointed it out as the part for determining modes is being refactored.

resolwe/flow/executors/docker/run.py
dblenkus: You can add just a quick description of arguments. There are just a lot of them, so it takes some time to understand them :)

resolwe/flow/managers/base.py
```diff
@@ -571,6 +571,10 @@ def _communicate(self, run_sync=False, verbosity=1, executor='resolwe.flow.execu
             state.MANAGER_CONTROL_CHANNEL,
             executor
         ))
+
+        # Ensure we have the correct executor loaded.
+        self.executor = self.load_executor(executor)  # pylint: disable=attribute-defined-outside-init
```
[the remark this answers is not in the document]
kostko: This has already been changed (`discover_engines` is now called instead), but yes we need re-discovery as the executor may change with each message.

resolwe/flow/executors/docker/run.py
kostko: Good point, I'll add another extra option for `tools` and one for `passwd`/`group`.

resolwe/flow/managers/base.py
kostko: Done.

resolwe/flow/execution_engines/bash/__init__.py
dblenkus: @kostko make the `filename` argument optional and return root data dir if it is not defined.

kostko: Done.

resolwe/flow/managers/base.py
```diff
@@ -212,23 +217,13 @@ def __init__(self, *args, **kwargs):
...
-        Django settings are patched and the ordering (in terms of import
-        order) where they're needed clash and make things awkward.
-        """
-        # This is all a bit kludgy, but we want runtime settings
-        # patching after the module is already loaded, and Channels
-        # wants a class variable.
-        setattr(type(self), 'method_mapping', {
-            state.MANAGER_CONTROL_CHANNEL: 'handle_control_event',
-        })
+        # Ensure there is only one manager instance per process. This is required
```
tjanez: Please, wrap the comment at 72 characters like the comments above.

resolwe/flow/managers/consumer.py
```diff
@@ -0,0 +1,24 @@
+"""Manager Channels consumer."""
```
tjanez: Nice!

resolwe/flow/executors/docker/prepare.py
```diff
@@ -20,3 +23,31 @@ def post_register_hook(self):
         """Pull Docker images needed by processes after registering."""
         if not getattr(settings, 'FLOW_DOCKER_DONT_PULL', False):
             call_command('list_docker_images', pull=True)
+
+    def resolve_data_path(self, data=None, filename=None):
+        """Resolve data path for use with the executor.
+
+        :param data: Data object instance
+        :param filename: Filename to resolve
+        :return: Resolved filename, which can be used to access the given data
```
tjanez: Please, wrap the docstring at 72 characters.

resolwe/flow/executors/docker/prepare.py
```diff
@@ -20,3 +23,31 @@ def post_register_hook(self):
...
+        if filename is None:
+            return os.path.join(constants.DATA_ALL_VOLUME, str(data.id))
+
+        return os.path.join(constants.DATA_ALL_VOLUME, str(data.id), filename)
+
+    def resolve_upload_path(self, filename=None):
+        """Resolve upload path for use with the executor.
+
+        :param filename: Filename to resolve
+        :return: Resolved filename, which can be used to access the given uploaded
```
tjanez: Please, wrap the docstring at 72 characters.

resolwe/flow/executors/prepare.py
```diff
@@ -65,3 +65,31 @@ def post_register_hook(self):
         this point. By default, it does nothing.
         """
         pass
+
+    def resolve_data_path(self, data=None, filename=None):
+        """Resolve data path for use with the executor.
+
+        :param data: Data object instance
+        :param filename: Filename to resolve
+        :return: Resolved filename, which can be used to access the given data
```
tjanez: Please, wrap the docstring at 72 characters.

resolwe/flow/executors/prepare.py
```diff
@@ -65,3 +65,31 @@ def post_register_hook(self):
...
+        if filename is None:
+            return os.path.join(settings.FLOW_EXECUTOR['DATA_DIR'], str(data.id))
+
+        return os.path.join(settings.FLOW_EXECUTOR['DATA_DIR'], str(data.id), filename)
+
+    def resolve_upload_path(self, filename=None):
+        """Resolve upload path for use with the executor.
+
+        :param filename: Filename to resolve
+        :return: Resolved filename, which can be used to access the given uploaded
```
tjanez: Please, wrap the docstring at 72 characters.

resolwe/flow/executors/docker/prepare.py
kostko: Should we have linters for these things? I never look at where I wrap things :-)

resolwe/flow/executors/docker/run.py
```diff
@@ -94,11 +96,46 @@ def start(self):
...
+                options.add('rw')
+
+            volumes.append({
+                'src': os.path.join(SETTINGS['FLOW_EXECUTOR'].get(base_dir_name, ''), *path),
+                'dest': volume,
+                'options': ','.join(options),
+            })
+
+        volumes = []
+        add_volume(volumes, 'data', 'DATA_DIR', constants.DATA_VOLUME, [self.data_id], read_only=False)
```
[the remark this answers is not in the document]
kostko: Well, the thing is that we use it with different lists (`volumes` and `self.mappings_tools`). Also having it like you propose requires everything to be in the same place. But yeah we could use `volumes += [...]` instead? I don't really have a preference here.

resolwe/flow/executors/run.py
```diff
@@ -153,6 +153,24 @@ def update_data_status(self, **kwargs):
...
+            # Send error report.
+            self.update_data_status(process_error=[str(error)], status=DATA_META['STATUS_ERROR'])
+
+            finish_fields = {
+                ExecutorProtocol.FINISH_PROCESS_RC: 1,
+            }
+
+        if finish_fields is not None:
+            self._send_manager_command(ExecutorProtocol.FINISH, extra_fields=finish_fields)
+        # the feedback key deletes itself once the list is drained
```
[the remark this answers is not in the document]
kostko: I found it confusing as well. It seems to be talking about the Redis list, which is deleted when all elements are popped? @jberci can you confirm this is the case?

resolwe/flow/managers/base.py
```diff
@@ -353,15 +356,29 @@ def _prepare_context(self, data_id, data_dir, runtime_dir, **kwargs):
...
+        # Save the secrets in the runtime dir, with permissions to prevent listing the given directory.
+        secrets_dir = os.path.join(runtime_dir, ExecutorFiles.SECRETS_DIR)
+        os.makedirs(secrets_dir, mode=stat.S_IWUSR | stat.S_IXUSR)
+        for file_name, value in secrets.items():
+            file_path = os.path.join(secrets_dir, file_name)
+
+            # Set umask to 0 to ensure that we set the correct permissions.
+            old_umask = os.umask(0)
+            try:
+                with os.fdopen(os.open(file_path, os.O_WRONLY | os.O_CREAT, 0o600), 'w') as raw_file:
```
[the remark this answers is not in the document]
kostko: > Will fdesc be closed automatically or does it have to be closed manually?

Yes, `fdopen` returns a file-like object backed by the file descriptor and since we use its context manager, it is closed after exiting the with block.

tests/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Ignore files created during test runs.
+.test_*/test_*
```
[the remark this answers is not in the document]
kostko: The `/tests` prefix is not needed as this `.gitignore` is not at the top level.

resolwe/flow/managers/base.py
```diff
@@ -353,15 +356,29 @@ def _prepare_context(self, data_id, data_dir, runtime_dir, **kwargs):
         }
 
         # extend the settings with whatever the executor wants
-        self.executor.extend_settings(data_id, files)
+        self.executor.extend_settings(data_id, files, secrets)
 
-        # save the settings into the various files in the data dir
+        # save the settings into the various files in the runtime dir
```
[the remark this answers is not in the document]
kostko: It should have been a sentence in the first place. I'll fix it.

tests/.gitignore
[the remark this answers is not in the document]
kostko: Also `{data,runtime,upload}` is not a valid pattern for `.gitignore`. I'll move it to a separate commit and add cases for each of them.

resolwe/flow/models/data.py
```diff
@@ -195,6 +196,48 @@ def save_storage(self, instance, schema):
                 # `value` is copied by value, so `fields[name]` must be changed
                 fields[name] = storage.pk
 
+    def resolve_secrets(self):
+        """Retrieve handles for all basic:secret: fields on input.
```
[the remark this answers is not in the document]
kostko: This line seems ok, I've wrapped the ones below.

resolwe/flow/executors/docker/prepare.py
tjanez: It's not yet implemented in pycodestyle, but the PR @ https://github.com/PyCQA/pycodestyle/pull/674 seems on right track and that it will got approval from pycodestyle's maintainer.

resolwe/flow/managers/base.py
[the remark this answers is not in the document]
tjanez: > Yes, `fdopen` returns a file-like object backed by the file descriptor and since we use its context manager, it is closed after exiting the with block.

Ok, perfect.

resolwe/flow/models/data.py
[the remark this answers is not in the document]
tjanez: 👍 , I meant all lines in the docstring.

resolwe/flow/executors/run.py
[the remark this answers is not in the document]
jberci: It is the case, the comment is about Redis deleting the list once all items have been popped. This behaviour was not obvious to me when I was doing this. If memory serves, the code was also structured so that one would expect explicit key deletion at the end, so I put that comment there. Feel free to get rid of it.

tests/.gitignore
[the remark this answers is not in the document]
tjanez: Sorry for the confusion about the pattern, I did't check if it accepted by `.gitignore`, I just wanted to concisely write what should be ignored 🙂 .

The new commit and git ignore pattern looks great!

resolwe/flow/executors/docker/run.py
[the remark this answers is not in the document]
tjanez: I see. I think for the time being, we can assume everything will be in the same place and use the `volumes =+ [...]` approach. Please, change this.

I didn't notice the first time I reviewed the code, the name `self.mappings_tools` is actually a left-over from the `FLOW_DOCKER_MAPPINGS` times. I think we should take this opportunity to also rename it to `self.volumes_tools` or `self.tools_volumes`.

resolwe/flow/executors/run.py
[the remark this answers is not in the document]
tjanez: Thanks for the explanation. I like @kostko's rewording and I think we can leave it in.

resolwe/flow/executors/docker/run.py
```diff
@@ -115,26 +152,32 @@ def start(self):
...
+                [tool]
+            )
+
+        volumes += self.mappings_tools
+
+        # Add any extra volumes verbatim.
+        volumes += SETTINGS.get('FLOW_DOCKER_EXTRA_VOLUMES', [])
+
+        # Create Docker --volume parameters from volumes.
+        command_args['volumes'] = ' '.join(['--volume="{src}":"{dest}":{options}'.format(**volume)
```
dblenkus: @kostko Will this work if `options` is empty? Or do you have to omit the colon?

kostko: Yes, it works if the options are empty (no need to omit the colon).

resolwe/flow/executors/docker/run.py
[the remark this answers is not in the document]
kostko: Done.

PR #408, opened 2017-12-07
Fix race condition when scanning for resolving Data objects

jberci:
Suppose two manager instances scan the database at roughly the same
time and enter the checking transaction for the same object. Both get to
the end of the transaction at the same time, having checked that the
object is indeed resolving. Both also add it to their own to-do queues
before exiting the transaction. At the end, exactly one of the
transactions will succeed, because they both affect the same rows, so
the other will roll back, however: the object will still stay in that
manager's queue. This means that both managers will proceed to try to
execute this Data object, even though one of them actually had a failed
status update transaction.

The fix is to make adding to the queue dependent on transaction success
by setting a flag in its transaction.on_commit event.

### DIFF
--- a/resolwe/flow/managers/base.py
+++ b/resolwe/flow/managers/base.py
@@ -204,7 +204,7 @@ class BaseManager(object):
         # such avoidance possible, all settings lookups in the worker
         # should use a private dictionary instead of the global
         # configuration. This variable is maintained around
-        # `_communicate()` calls, effectively emulating Django's
+        # _data_scan() calls, effectively emulating Django's
         # settings but having no effect on anything outside the worker
         # code (in particular, the signal triggers).
         self.settings_actual = {}
@@ -457,11 +457,11 @@ class BaseManager(object):
 
         if cmd == WorkerProtocol.COMMUNICATE:
             try:
-                self._communicate(**message.content[WorkerProtocol.COMMUNICATE_EXTRA])
+                self._data_scan(**message.content[WorkerProtocol.COMMUNICATE_EXTRA])
             finally:
                 # Clear communicate() claim on the semaphore.
                 new_sema = self.state.sync_semaphore.add(-1)
-                logger.debug(__("Manager changed sync_semaphore DOWN to {} after _communicate().", new_sema))
+                logger.debug(__("Manager changed sync_semaphore DOWN to {} after _data_scan().", new_sema))
 
         elif cmd == WorkerProtocol.FINISH:
             data_id = message.content[WorkerProtocol.DATA_ID]
@@ -481,11 +481,11 @@ class BaseManager(object):
                 new_sema = self.state.sync_semaphore.add(1)
                 logger.debug(__("Manager changed sync_semaphore UP to {} in spawn handler.", new_sema))
                 try:
-                    self._communicate(**message.content[WorkerProtocol.FINISH_COMMUNICATE_EXTRA])
+                    self._data_scan(**message.content[WorkerProtocol.FINISH_COMMUNICATE_EXTRA])
                 finally:
                     # Clear communicate() claim on the semaphore.
                     new_sema = self.state.sync_semaphore.add(-1)
-                    logger.debug(__("Manager changed sync_semaphore DOWN to {} after _communicate().", new_sema))
+                    logger.debug(__("Manager changed sync_semaphore DOWN to {} after _data_scan().", new_sema))
 
             self.state.executor_count.add(-1)
             # Clear execution claim on the semaphore.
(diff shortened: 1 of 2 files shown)


resolwe/flow/managers/base.py
```diff
@@ -624,39 +628,47 @@ def _communicate(self, run_sync=False, verbosity=1, executor='resolwe.flow.execu
...
-                        except OSError as err:
-                            logger.error(__(
-                                "OSError occurred while preparing data {} (will skip): {}",
-                                data.id, err
-                            ))
-                            continue
-
-                        queue.append((data.id, priority, runtime_dir, argv))
+                    # Signal transaction success through the flag.
+                    transaction.on_commit(lambda: transaction_success.append(True))
```
kostko: This is quite convoluted, why don't you instead push to queue on transaction success instead of introducing an intermediate flag-list?

jberci:
If the manager crashes between transaction end and before execution start, the object will be lost. To handle that, I suppose we'd have to have basically the whole `_communicate` call in a transaction, but we'd have to get rid of the queue. If there's a crash anywhere, then the started objects shouldn't roll back, which implies the transactions should be per-object. But for the duration of the manager's processing, up to the point where the process is started, things should be atomic to the outside world.

resolwe/flow/managers/base.py
jberci: This was an explicit scoping decision. When an object is vetted for execution, there's still quite a bit of code to execute, but with enough state in the local scope to make a separate function awkward. The flag list is necessary because lambdas can't contain assignment expressions, so a primitive wouldn't work. Given your comment though, we could refactor this entirely so that `_communicate` only does the scanning and then calls a special commit function `on_commit`. This way, it would be easier to correctly handle transactions and get rid of the queue. Thoughts?

kostko: > Given your comment though, we could refactor this entirely so that _communicate only does the scanning and then calls a special commit function on_commit. This way, it would be easier to correctly handle transactions and get rid of the queue. Thoughts?

This sounds better. So the run methods would be called one by one (after commit) instead of first going to the queue?

resolwe/flow/managers/base.py
```diff
@@ -204,7 +204,7 @@ def __init__(self, *args, **kwargs):
         # such avoidance possible, all settings lookups in the worker
         # should use a private dictionary instead of the global
         # configuration. This variable is maintained around
-        # `_communicate()` calls, effectively emulating Django's
+        # `_data_scan()` calls, effectively emulating Django's
```
tjanez: Drop backticks around `_data_scan()` since this is not Markdown.

resolwe/flow/managers/base.py
```diff
@@ -577,7 +577,74 @@ def communicate(self, run_sync=False, verbosity=1, save_settings=True):
...
+        if verbosity >= 1:
+            print("Running", runtime_dir)
+        self.state.executor_count.add(1)
+        # Set execution claim on the semaphore.
+        new_sema = self.state.sync_semaphore.add(1)
+        logger.debug(__("Manager changed sync_semaphore UP to {} on executor start.", new_sema))
+        self.run(data.id, runtime_dir, argv, priority=priority, verbosity=verbosity)
+
+    def _data_scan(self, run_sync=False, verbosity=1, executor='resolwe.flow.executors.local', **kwargs):
         """Resolve task dependencies and run the task.
```
tjanez: Would `Resolwe task dependencies, mark Data objects for execution.` be more appropriate after this refactoring?

resolwe/flow/managers/base.py
```diff
@@ -590,6 +590,7 @@ def _communicate(self, run_sync=False, verbosity=1, executor='resolwe.flow.execu
             discovered in this pass.
         """
         queue = []
+        transaction_success = []
```
tjanez: Remove this, it is not needed.

resolwe/flow/managers/base.py
```diff
@@ -654,44 +658,52 @@ def _communicate(self, run_sync=False, verbosity=1, executor='resolwe.flow.execu
...
-                        queue.append((data.id, priority, runtime_dir, argv))
+                    # Signal transaction success through the flag.
+                    transaction.on_commit(lambda: transaction_success.append(True))
 
                     # All data objects created by the execution engine are commited after this
                     # point and may be processed by other managers running in parallel. At the
-                    # same time, lock for the current data object is released.
+                    # same time, the lock for the current data object is released.
+
+                # End of the with clause
```
tjanez: Remove this line, in the next commit, this is the end of the function.

resolwe/flow/managers/base.py
jberci: Not entirely, since the functionality hasn't changed in any way. Objects aren't marked for anything either, they're handed directly to the runner function.

tjanez: I want the docstring to reflect the change of splitting the execution part to the `_data_exection()` method.
What about `Resolve task dependencies and trigger task execution` since we use Django's signals to trigger execution after transaction is commited?

jberci: We don't use signals for this. The behaviour of these two functions is precisely the same as it was when they were part of `_communicate`.

tjanez: @jberci, thanks for the explanation, I agree with your argument.

PR #410, opened 2017-12-11
 Speed up dependency resolver by using parent relation

dblenkus:


resolwe/flow/managers/base.py
```diff
@@ -53,37 +52,31 @@ def default(self, o):  # pylint: disable=method-hidden
...
     """
-    for field_schema, fields in iterate_fields(data.input, data.process.input_schema):
-        if (field_schema['type'].lower().startswith('data:') or
-                field_schema['type'].lower().startswith('list:data:')):
-            name = field_schema['name']
-            value = fields[name]
+    parents_status = set(
+        DataDependency.objects.filter(
+            child=data, kind=DataDependency.KIND_IO
+        ).values_list('parent__status', flat=True)
```
kostko: Since you are creating a set in Python, you could use distinct to reduce the number of entries on the database server?

dblenkus: Good point, i will fix it.

resolwe/flow/managers/base.py
```diff
@@ -53,37 +52,31 @@ def default(self, o):  # pylint: disable=method-hidden
...
     STATUS_DONE .. all dependencies have done status
     None .. other
 
     """
-    for field_schema, fields in iterate_fields(data.input, data.process.input_schema):
-        if (field_schema['type'].lower().startswith('data:') or
-                field_schema['type'].lower().startswith('list:data:')):
-            name = field_schema['name']
-            value = fields[name]
+    parents_status = set(
```
tjanez: I would rename this to `parent_statuses` or `parents_statuses` since the lines below are a bit confusing when you check if `len(parents_status)` is `0` or `1` and the name implies there is one status in the variable.

PR #413, opened 2017-12-12
Lock object for update while updating over API

dblenkus:


resolwe/flow/views/mixins.py
```diff
@@ -102,7 +103,55 @@ def update(self, request, *args, **kwargs):
                         'Invalid descriptor_schema slug "{}" - object does not exist.'.format(ds_query)]},
                     status=status.HTTP_400_BAD_REQUEST)
 
-        return super(ResolweUpdateModelMixin, self).update(request, *args, **kwargs)
+        # return super(ResolweUpdateModelMixin, self).update(request, *args, **kwargs)
```
kostko: Add a comment above this line, saying that "Use the original method instead when support for locking is added" + link to issue (as below). Otherwise it is not immediately clear why there is a commented-out block of code here.

resolwe/flow/views/mixins.py
```diff
@@ -102,7 +103,55 @@ def update(self, request, *args, **kwargs):
...
+            'Expected view %s to be called with a URL keyword argument '
+            'named "%s". Fix your URL conf, or set the `.lookup_field` '
+            'attribute on the view correctly.' %
+            (self.__class__.__name__, lookup_url_kwarg)
+        )
+
+        filter_kwargs = {self.lookup_field: self.kwargs[lookup_url_kwarg]}
+        obj = get_object_or_404(queryset.select_for_update(), **filter_kwargs)
+
+        # May raise a permission denied
```
kostko: Dot at the end.

resolwe/flow/views/mixins.py
```diff
@@ -102,7 +103,55 @@ def update(self, request, *args, **kwargs):
...
+
+    # NOTE: This is a copy of rest_framework.mixins.UpdateModelMixin.update().
+    #       The self.get_object() was replaced with the
+    #       self.get_object_with_lock() to lock the object while updating.
+    #       Use the original method when suport for locking is added:
+    #       https://github.com/encode/django-rest-framework/issues/4675
+    def _update(self, request, *args, **kwargs):
+        """Update a resource."""
+        partial = kwargs.pop('partial', False)
+        instance = self.get_object_with_lock()
```
kostko: I think it would be nice to explicitly mark the changed line, even though this is mentioned in the comment above.

resolwe/flow/views/mixins.py
```diff
@@ -102,7 +103,55 @@ def update(self, request, *args, **kwargs):
...
+
+        assert lookup_url_kwarg in self.kwargs, (
+            'Expected view %s to be called with a URL keyword argument '
+            'named "%s". Fix your URL conf, or set the `.lookup_field` '
+            'attribute on the view correctly.' %
+            (self.__class__.__name__, lookup_url_kwarg)
+        )
+
+        filter_kwargs = {self.lookup_field: self.kwargs[lookup_url_kwarg]}
+        obj = get_object_or_404(queryset.select_for_update(), **filter_kwargs)
```
kostko: I think it would be nice to explicitly mark the changed line, even though this is mentioned in the comment above.

dblenkus:
I have no idea how to write test as it would require to do 2 parallel requests and to make sure that both reads from the database before writing in it. @kostko do you have any idea?

PR #414, opened 2017-12-14
Manager dispatcher and SLURM support

jberci:
This PR introduces dispatching functionality to the manager, which selects and instantiates workload manager connectors based on Data object attributes.

Manager subclassing is no longer supported; workload management system support is done through connector classes which are much lighter weight than subclassing the main manager.

The initial SLURM support is a side dish.


resolwe/flow/managers/dispatcher.py
```diff
@@ -0,0 +1,50 @@
...
+        if 'DISPATCHER_MAPPING' not in getattr(settings, 'FLOW_MANAGER', {}):
+            raise ImproperlyConfigured("Required key 'DISPATCHER_MAPPING' not in FLOW_MANAGER settings.")
+
+        mapping = settings.FLOW_MANAGER['DISPATCHER_MAPPING']
+        self.scheduling_class_map = {choice[0]: choice[1] for choice in Process.SCHEDULING_CLASS_CHOICES}
+
+        scheduling_classes = set(self.scheduling_class_map.values())
+        map_keys = set(mapping.keys())
+        if not scheduling_classes <= map_keys:
+            raise ImproperlyConfigured("Dispatcher manager mapping settings incomplete.")
```
dblenkus: Add a list of missing keys.

resolwe/flow/managers/dispatcher.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def run(self, data, dest_dir, argv, run_sync=False, verbosity=1):
+        """Select a concrete manager and run the object through it.
+
+        For details, see
+        :meth:`~resolwe.flow.managers.base.BaseManager.run`.
+        """
+        process_scheduling = self.scheduling_class_map[data.process.scheduling_class]
+        manager_class = settings.FLOW_MANAGER['DISPATCHER_MAPPING'][process_scheduling]
+        manager_module = import_module(manager_class)
```
dblenkus: Import all managers in `__init__`.

resolwe/flow/managers/slurm.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def run(self, data, dest_dir, argv, run_sync=False, verbosity=1):
+        """Run process with SLURM.
+
+        For details, see
+        :meth:`~resolwe.flow.managers.base.BaseManager.run`.
+        """
+        limits = data.process.get_resource_limits()
+        command = ['/usr/bin/env', 'sbatch']
+        command.append('--mem={}M'.format(limits['memory'] + 200))
```
dblenkus: Add a comment why 200 is added.

resolwe/flow/models/process.py
```diff
@@ -156,3 +157,35 @@ class Meta(BaseModel.Meta):
...
+        # Get limit defaults and overrides.
+        limit_defaults = settings.get('FLOW_DOCKER_LIMIT_DEFAULTS', {})
+        limit_overrides = settings.get('FLOW_DOCKER_LIMIT_OVERRIDES', {})
+
+        limits = {}
+
+        resources = self.requirements.get('executor', {}).get('resources', {})
+
+        if 'cores' in resources:
+            limits['cores'] = int(resources['cores'])
```
dblenkus: @kostko, @tjanez would it ok to set cores requirement to 1 by default?

resolwe/flow/managers/dispatcher.py
dblenkus: Would it make sense to use [`load manager`](https://github.com/genialis/resolwe/blob/master/resolwe/flow/managers/__init__.py#L15-L37)?

resolwe/flow/models/process.py
kostko: I think it would be ok.

resolwe/flow/managers/dispatcher.py
```diff
@@ -0,0 +1,50 @@
...
+        manager = manager_module.Manager(internal=True)
```
kostko: Can we be sure that creating a new internal manager instance will not cause any issues due to there now being multiple managers (one singleton instance, and another internal manager)? Can settings (and loaded executors, engines, etc.) differ in any way? Because this new manager will also perform discovery of all engines and create new instances of them, which may cause issues.

resolwe/flow/managers/dispatcher.py
jberci: `load_manager` is obsolete and gone, as discussed.

resolwe/flow/managers/dispatcher.py
jberci: There's only one instance again. The actual running logic that was in subclasses is now in a separate class hierarchy. I've also reinstated your singleton check in `Manager.__init__`.

resolwe/flow/executors/docker/run.py
```diff
@@ -36,24 +36,17 @@ def start(self):
...
 
-        # Get limit defaults and overrides.
+        # Get limit defaults.
         limit_defaults = SETTINGS.get('FLOW_DOCKER_LIMIT_DEFAULTS', {})
-        limit_overrides = SETTINGS.get('FLOW_DOCKER_LIMIT_OVERRIDES', {})
 
         # Set resource limits.
         limits = []
-        if 'cores' in self.resources:
+        if 'cores' in self.process['resource_limits']:
```
dblenkus: You can omit if, as `cores` is always there.

resolwe/flow/managers/workload_connectors/base.py
```diff
@@ -0,0 +1,31 @@
...
+
+    The main :class:`~resolwe.flow.managers.dispatcher.Manager` instance
+    in :data:`~resolwe.flow.managers.manager` uses connectors to handle
+    communication with concrete backend workload management systems,
+    such as Celery and SLURM. The connectors need not worry about how
+    jobs are discovered or how they're prepared for execution; this is
+    all done by the manager.
+    """
+
+    def submit(self, data, dest_dir, argv, verbosity=1):
```
dblenkus: Also add a method for cancelling jobs, as this is the next feature we have to support.

resolwe/flow/managers/dispatcher.py
```diff
@@ -250,21 +285,22 @@ def _include_environment_variables(self, program):
...
             :class:`~resolwe.flow.models.Data` object should be run from.
         :param argv: The argument vector used to spawn the executor.
-        :param priority: The execution priority for this job.
-        :param run_sync: If ``True``, the method will not return until
-            the manager is finished processing everything from this call
-            onwards.
         :param verbosity: Integer logging verbosity level.
         """
-        raise NotImplementedError("Subclasses of BaseManager must implement a run() method.")
+        process_scheduling = self.scheduling_class_map[data.process.scheduling_class]
```
dblenkus: Prepare `scheduling_class` -> `connector` mapping in `__init__`.

resolwe/flow/managers/workload_connectors/slurm.py
```diff
@@ -0,0 +1,65 @@
...
+
+                script.write('#SBATCH --cpus-per-task={}\n'.format(limits['cores']))
+
+                # Render the argument vector into a command line.
+                line = ' '.join(map(shlex.quote, argv))
+                script.write(line + '\n')
+                os.chmod(script.fileno(), 0o555)
+
+            command = ['/usr/bin/env', 'sbatch', script_path]
+            subprocess.Popen(
```
dblenkus: Command returns id of the job which has to be saved in [`Data.process_pid`](https://github.com/genialis/resolwe/blob/master/resolwe/flow/models/data.py#L107) (the field should probably also be renamed) and will be used for canceling jobs.

dblenkus:
@kostko please do another review pass.

resolwe/flow/managers/workload_connectors/slurm.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        limits = data.process.get_resource_limits()
+        logger.debug(__(
+            "Connector '{}.{}' running {}.",
+            self.__class__.__module__,
+            self.__class__.__name__,
+            repr(argv)
+        ))
+        try:
+            script_path = os.path.join(dest_dir, 'slurm.sh')
```
kostko: Where is this script stored in (it is not obvious what `dest_dir` is from above comments)? Is this in the runtime or the data directory? It should be in the runtime directory.

resolwe/flow/managers/workload_connectors/slurm.py
```diff
@@ -0,0 +1,65 @@
...
+            with open(script_path, 'wt') as script:
+                script.write('#!/bin/bash\n')
+
+                # We add 200MB to the memory limit to account for
+                # executor overhead, since the executor is running in
+                # the same environment as the process.
+                script.write('#SBATCH --mem={}M\n'.format(limits['memory'] + 200))
```
kostko: Why 200MB? Should this be a module-level constant (e.g., `EXECUTOR_MEMORY_OVERHEAD`) instead?

resolwe/flow/managers/workload_connectors/slurm.py
```diff
@@ -0,0 +1,65 @@
...
+
+                script.write('#SBATCH --cpus-per-task={}\n'.format(limits['cores']))
+
+                # Render the argument vector into a command line.
+                line = ' '.join(map(shlex.quote, argv))
+                script.write(line + '\n')
+                os.chmod(script.fileno(), 0o555)
```
kostko: Why is this done like this? The file should be created with the correct mode, not chmoded later.

resolwe/flow/managers/workload_connectors/slurm.py
jberci: It is in the runtime directory. I'll try changing parameter names and docstrings to make it more obvious.

resolwe/flow/managers/workload_connectors/slurm.py
jberci: @dblenkus ?

dblenkus: It's just an arbitrary chosen amount that is big enough to work without problems. As clusters normally have enough memory, it is safer to use a bit larger number.

You can put it into a constant, but it shouldn't be controllable in settings.

resolwe/flow/managers/workload_connectors/base.py
jberci: I think it would be better to add that when we actually get around to implementing it. Having just a stub method which nobody can use seems a bit pointless, especially because we don't have any binary compatibility to worry about.

resolwe/flow/managers/workload_connectors/slurm.py
jberci: See above. I can add a comment here to remind us about this when we get to it.

resolwe/flow/managers/slurm.py
```diff
@@ -0,0 +1,64 @@
...
+        limits = data.process.get_resource_limits()
+        logger.debug(__(
+            "Manager '{}.{}' running {}.",
+            self.__class__.__module__,
+            self.__class__.__name__,
+            repr(argv)
+        ))
+        try:
+            script_path = os.path.join(dest_dir, 'slurm.sh')
+            file_descriptor = os.open(script_path, os.O_WRONLY | os.O_CREAT, mode=0o555)
```
tjanez: Please, add a comment that explains why we want this low-level type of file creation.

resolwe/flow/managers/dispatcher.py
```diff
@@ -267,17 +267,14 @@ def _include_environment_variables(self, program):
...
-    def run(self, data, dest_dir, argv, run_sync=False, verbosity=1):
+    def run(self, data, dest_dir, argv, verbosity=1):
         """Select a concrete runner and run the process through it.
 
         :param data: The :class:`~resolwe.flow.models.Data` object that
             is to be run.
         :param dest_dir: The directory the
             :class:`~resolwe.flow.models.Data` object should be run from.
         :param argv: The argument vector used to spawn the executor.
-        :param run_sync: If ``True``, the method will not return until
```
tjanez: Can you extend the comment message with an explanation about why we want and can get rid of `run_sync` flag?

jberci: The commit log already says the flag is disused. By definition, this means there's no code anywhere that uses it, so it's redundant and can be removed, which the commit does.

resolwe/flow/models/process.py
```diff
@@ -156,3 +157,35 @@ class Meta(BaseModel.Meta):
...
     """
+
+    def get_resource_limits(self):
+        """Get the core count and memory usage limits for this process.
+
+        :return: A dictionary with the resource limits, containing the
+            following keys:
+
+            - ``memory``: Memory usage limit, in MB.
+            - ``cores``: Optional. Core count limit.
```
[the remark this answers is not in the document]
jberci: That was a leftover, it used to be optional. Fixed.

resolwe/flow/models/process.py
```diff
@@ -156,3 +157,35 @@ class Meta(BaseModel.Meta):
...
+
+        :rtype: dict
+        """
+        # Get limit defaults and overrides.
+        limit_defaults = getattr(settings, 'FLOW_DOCKER_LIMIT_DEFAULTS', {})
```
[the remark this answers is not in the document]
jberci: I left those as they were, as they're entirely out of scope for this PR.
@dblenkus @kostko ?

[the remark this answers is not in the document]
dblenkus: I don't like the idea with another layer of nested dicts. I would just rename `DOCKER` to `EXECUTOR`.

resolwe/flow/managers/dispatcher.py
tjanez: I guess I read too quickly and missed the *disused* part 🙂 .

PR #420, opened 2018-01-09
 Pass data_id to communicate method

dblenkus:


kostko:
Rebase.

dblenkus:
Rebased.

mstajdohar:
Should we add a Django command to trigger the communicate manually in case of overflow?

PR #427, opened 2018-01-16
Don't trim predefined slugs

dblenkus:


resolwe/flow/models/fields.py
```diff
@@ -110,8 +110,11 @@ def pre_save(self, instance, add):
                 slug = ''
 
         if slug:
-            # Make sure that potentially added sequence won't excede maximal length.
-            slug = slug[:(self.max_length - MAX_SLUG_SEQUENCE_DIGITS - 1)]
+            # Make sure that auto generated slug with added sequence
+            # won't excede maximal length.
+            # Validation of predefined slugs is handled by Django.
+            if not predefined_slug:
```
kostko: Can you combine with the if above?

dblenkus: No, just trimming has to be skipped for predefined slugs. The rest of the above if is still needed to determine availability of the slug

kostko: Ah I didn't see there is code below in the same if, nevermind.

tjanez:
I've run all Tox environments locally and all of them succeeded so I think we can merge this since we don't know how long [Travis CI will need to restore their service](https://www.traviscistatus.com/incidents/c1s4dlyxd4lf).

PR #428, opened 2018-01-17
Propagate internal errors without Django ORM

kostko:


resolwe/flow/managers/dispatcher.py
```diff
@@ -753,26 +753,47 @@ def process_data_object(data):
...
+                        with connection.cursor() as cursor:
+                            cursor.execute(
+                                """
+                                    UPDATE {table} SET status = %s, process_error = process_error || (%s)::varchar[]
+                                    WHERE id = %s
+                                """.format(
+                                    table=Data._meta.db_table
+                                ),
+                                [
+                                    Data.STATUS_ERROR,
```
dblenkus: Use named params.

dblenkus:
And add a change log :)

kostko:
We already have "Make manager more robust to unhandled failures during data object processing", so I am not sure about this :-)

PR #430, opened 2018-01-18
Trim process_* fields before saving them

dblenkus:


resolwe/flow/tests/processes/tests.yml
```diff
@@ -40,6 +40,23 @@
     program: |
       echo "{\"output\":{\"file\":\"i-dont-exist.zip\"}}"
 
+- slug: test-process-notofications
```
kostko: Typo (`notofications`).

PR #433, opened 2018-01-23
Enable manager prefix overriding from the env

jberci:
Enable manager prefix overriding from the environment, in case someone wants to run multiple instances of the test suite at once.

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -83,7 +83,7 @@ DATABASES = {
         'HOST': pghost,
         'PORT': pgport,
         'TEST': {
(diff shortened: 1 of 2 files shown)


tests/settings.py
```diff
@@ -83,7 +83,7 @@
         'HOST': pghost,
         'PORT': pgport,
         'TEST': {
-            'NAME': 'resolwe_test' + toxenv
+            'NAME': pgname + '_test'
```
tjanez: Please, add this as a separate commit.

tjanez:
Also, don't forget to rebase.

PR #435, opened 2018-01-24
Remove Jinja support in DescriptorSchema's default values

dblenkus:


docs/CHANGELOG.rst
```diff
@@ -14,6 +14,8 @@ Changed
 -------
 - **BACKWARD INCOMPATIBLE:** Remove Ubuntu 17.04 base Docker image due to end
   of lifetime
+- **BACKWARD INCOMPATIBLE:** Remove support for Jinja in ``DescriptorSchema``'s
```
mstajdohar: Maybe write that we are removing this because this is some leftover, dead code (not used within our platform) that is causing problems when updating a descriptor of the data objects with deleted inputs.

PR #436, opened 2018-01-24
Use filesystem-level lock for the data directory

kostko:
This prevents the executor from being run twice on the same data directory. In this case, the second executor will abort.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -36,6 +36,7 @@ Fixed
 - Make parallel test suite worker threads clean up after initialization failures
 - Add mechanism to override the manager's control channel prefix from the
   environment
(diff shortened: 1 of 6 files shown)


kostko:
BTW, why is stdout/jsonout stored in the data directory? I think it would be better to store it in the runtime directory instead, to protect it from the process running in the container, which may corrupt these files and make debugging much harder.

tjanez:
> BTW, why is stdout/jsonout stored in the data directory? I think it would be better to store it in the runtime directory instead, to protect it from the process running in the container, which may corrupt these files and make debugging much harder.

Agreed. I don't see a reason why we would not move it there (before executor separation, we didn't have this directory, therefore, we had it in the data directory).
However, this would be a breaking change since we rely on this stdout/json location when serving them through the web server.

PR #437, opened 2018-01-24
Propagate log messages from executors to the master

dblenkus:


resolwe/flow/executors/manager_comm.py
```diff
@@ -0,0 +1,66 @@
+"""Utility functions for communicating with the manager."""
```
kostko: This file has a strange name `manager_comm`? Can you use a full name instead?

dblenkus: Fixed.

PR #439, opened 2018-02-07
Fix delete of Data objects belonging to two Entyties

dblenkus:


mstajdohar:
Typo in the commit message, Entyties.

PR #441, opened 2018-02-15
Don't set Data object's status to error if executor is run twice

dblenkus:


resolwe/flow/executors/run.py
```diff
@@ -104,6 +104,8 @@ def run(self, data_id, script, verbosity=1):
         if finish_fields is not None:
             send_manager_command(ExecutorProtocol.FINISH, extra_fields=finish_fields)
 
+        print("Kje je piskotek?")
```
kostko: WTF?

resolwe/flow/managers/listener.py
```diff
@@ -420,6 +420,33 @@ def handle_finish(self, obj):
...
         })
 
+    def handle_abort(self, obj):
+        """Handle an incoming ``Data`` abort processing request.
+
+        .. IMPORTANT::
+
+            This only makes manager's state consistent and doesn't
+            affect Data object in any way. Any changes to the Data
+            must be applied over ``habdle_update`` method.
```
kostko: Typo, `handle_update`.

resolwe/flow/managers/listener.py
```diff
@@ -420,6 +420,33 @@ def handle_finish(self, obj):
             },
         })
 
+    def handle_abort(self, obj):
+        """Handle an incoming ``Data`` abort processing request.
+
+        .. IMPORTANT::
+
+            This only makes manager's state consistent and doesn't
```
kostko: It makes it consistent only in case the protocol was followed. In case the executor was executed twice (and the manager did not actually dispatch twice), this will actually make the state inconsistent.

resolwe/flow/executors/run.py
dblenkus: :)

PR #442, opened 2018-02-15
Container image changes

kostko:
Tests currently fail because `resolwe/archiver` and `resolwe/upload-tab-file` have no image tag specified (tag `latest` is no longer allowed). We should tag the images and update these processes, perhaps `1.0.0`?

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -16,6 +16,10 @@ Changed
   of lifetime
 - **BACKWARD INCOMPATIBLE:** Remove support for Jinja in ``DescriptorSchema``'s
   default values
+- **BACKWARD INCOMPATIBLE:** Remove ``CONTAINER_IMAGE`` configuration option
+  from the Docker executor; if no container image is specified for a process
+  using the Docker executor, the same pre-defined default image is used
+  (currently this is ``resolwe/base:ubuntu-16.04``)
 - Add mechanism to change test database name from the environment, appending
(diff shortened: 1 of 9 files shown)


tests/settings.py
```diff
@@ -142,6 +142,9 @@
 # Disable SECCOMP if set via environment variable.
 FLOW_DOCKER_DISABLE_SECCOMP = strtobool(os.environ.get('RESOLWE_DOCKER_DISABLE_SECCOMP', '0'))
 
+# Ensure all container images follow a specific format.
```
dblenkus: I would prefer more specific comment mentioning that `latest` tag is not allowed.

Do we want to check only that the tag is not `latest` or would it be better to allow only tags in form of `x.y.z`?

docs/CHANGELOG.rst
```diff
@@ -16,6 +16,10 @@ Changed
   of lifetime
 - **BACKWARD INCOMPATIBLE:** Remove support for Jinja in ``DescriptorSchema``'s
   default values
+- **BACKWARD INCOMPATIBLE:** Remove ``CONTAINER_IMAGE`` configuration option from
```
tjanez: Can you wrap the lines at 80 characters?

tests/settings.py
tjanez: As we've discussed, we have tags like `ubuntu-16.04` so we can't use `x.y.z`.

PR #447, opened 2018-02-20
Hydrate refs in file and dir fields

dblenkus:


resolwe/flow/models/utils.py
```diff
@@ -249,17 +249,21 @@ def hydrate_storage(storage_id):
         if 'type' in field_schema:
             if field_schema['type'].startswith('basic:file:'):
                 value['file'] = hydrate_path(value['file'])
+                value['refs'] = [hydrate_path(ref) for ref in value.get('refs', [])]
```
mstajdohar: Do we want `value['refs']` set to `[]` if the `refs` key did not exist in `value` before?

mstajdohar: Maybe that's a feature, because then you always have `refs` in processes.

dblenkus: Yes, i think it is easier to write processes if `refs` is always there, that's why I did it like this.

PR #450, opened 2018-03-07
Add FLOW_MAX_CORES_PER_PROCESS setting

dblenkus:


mstajdohar:
We could also call this `FLOW_PROCESS_CORE_LIMIT`.

mstajdohar:
@tjanez, @jberci You usually pick good words for parameter names. What do you suggest?

resolwe/flow/models/process.py
```diff
@@ -180,6 +180,10 @@ def get_resource_limits(self):
 
         limits['cores'] = int(resources.get('cores', 1))
 
+        max_cores = getattr(settings, 'FLOW_MAX_CORES_PER_PROCESS', None)
```
tjanez: I would suggest `FLOW_PROCESS_MAX_CORES` since it would be similar to `FLOW_PROCESS_RESOURCE_DEFAULTS` and `FLOW_PROCESS_RESOURCE_OVERRIDES`.

mstajdohar:
Thanks. Just a side note - we may want to consider renaming `FLOW_PROCESSES_FINDERS` and `FLOW_PROCESSES_DIRS` to `FLOW_PROCESS_FINDERS` and `FLOW_PROCESS_DIRS` at some point.

PR #452, opened 2018-03-14
Fix processing of dependencies in ES indices

dblenkus:


resolwe/elastic/builder.py
```diff
@@ -26,6 +27,60 @@
...
 
+# UUID used in tests to make sure that no index is re-used.
+TESTING_UUID = str(uuid.uuid4())
+
+
+class BuildArgumentsCache:
+    """Cache for storing arguments for index builder.
+
+    If set value contains queryset, it is evaluated and later recreated
+    to prevend influence from database changes.
```
kostko: prevent

resolwe/elastic/builder.py
```diff
@@ -26,6 +27,60 @@
...
 )
 
+# UUID used in tests to make sure that no index is re-used.
+TESTING_UUID = str(uuid.uuid4())
+
+
+class BuildArgumentsCache:
+    """Cache for storing arguments for index builder.
+
+    If set value contains queryset, it is evaluated and later recreated
```
kostko: a queryset

resolwe/elastic/builder.py
```diff
@@ -26,6 +27,60 @@
...
+            return
+
+        if 'queryset' in build_kwargs:
+            self.model = build_kwargs['queryset'].model
+            self.pks = list(build_kwargs['queryset'].values_list('pk', flat=True))
+
+        elif 'obj' in build_kwargs:
+            self.obj = build_kwargs['obj']
+
+    def get(self):
```
kostko: Since the cache is cleaned, perhaps a better name would be `take`?

resolwe/elastic/builder.py
```diff
@@ -98,12 +183,18 @@ def __init__(self, field):
         """Construct m2m dependency."""
         super(ManyToManyDependency, self).__init__(field.rel.model)
         self.field = field
+        self.delete_cache = BuildArgumentsCache()
+        self.remove_cache = BuildArgumentsCache()
```
kostko: The naming is non-obvious. Delete vs. remove? Can this be documented somewhere or renamed?

PR #453, opened 2018-03-15
Switch Data API viewset to use Elasticsearch

kostko:
**WARNING: This is a breaking change for the Data API endpoint. Any filter usage that is not explicitly tested in `TestDataViewSetFilters` may not work anymore or may behave differently.**

TODO
* [x] Add tests for reverse and self m2m relations in ES.
* [x] Add lookup expressions for ES viewsets.
* [x] Port tests for filters, check if all filters work.


resolwe/flow/elastic_indexes/data.py
```diff
@@ -0,0 +1,83 @@
...
+    process_name = dsl.String()
+    tags = dsl.String(multi=True, index='not_analyzed')
+
+    collection = dsl.Integer(multi=True)
+    parents = dsl.Integer(multi=True)
+    children = dsl.Integer(multi=True)
+    entity = dsl.Integer(multi=True)
+
+    # Output fields (may not exist on all data objects).
+    source = dsl.String(index='not_analyzed')
```
dblenkus: This does not belong into resolwe.

kostko: Oh good point. Hm, this is annoying as we would need to augment the index in resolwe-bio?

resolwe/flow/views/data.py
```diff
@@ -33,9 +37,55 @@ class DataViewSet(ResolweCreateModelMixin,
...
+    def custom_filter_text(self, value, search):
+        """Support general query using the 'text' attribute."""
+        if isinstance(value, list):
+            value = ' '.join(value)
+
+        search = search.query(
+            'bool',
+            should=[
+                Q('match', slug={'query': value, 'operator': 'and'}),
+                Q('match', name={'query': value, 'operator': 'and'}),
```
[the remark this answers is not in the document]
kostko: My local version has boosts all over the place.

kostko:
@dblenkus this is now ready for review.

resolwe/test/testcases/__init__.py
```diff
@@ -61,8 +59,17 @@ def _clean_up(self):
...
         """Prepare environment for test."""
+        from resolwe.elastic.builder import index_builder
+
         super().setUp()
 
+        # Reset Elastic search indices.
+        index_builder.destroy()
+        index_builder.discover_indexes()
+        index_builder.create_mappings()
+        index_builder.unregister_signals()
```
dblenkus: Do you need to call `unregister_signals `? It is already called as a part of `destroy()`.

kostko: I like to be on the safe side and I think this operation is quite fast.

resolwe/elastic/viewsets.py
```diff
@@ -96,7 +96,15 @@ def filter_search(self, search):
 
         """
         builder = QueryBuilder(self.filtering_fields, self)
-        search, _ = builder.build(search, self.get_query_params())
+        search, unmatched = builder.build(search, self.get_query_params())
+
+        # Ensure that no unsupported arguments were used.
+        unmatched.pop('ordering', None)
+        unmatched.pop('offset', None)
+        unmatched.pop('limit', None)
```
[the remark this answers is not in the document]
kostko: Done.

PR #454, opened 2018-03-16
Add sort argument to _assert_file

JureZmrzlikar:


mstajdohar:
And write a simple test.

JureZmrzlikar:
@mstajdohar , @dblenkus : Do we have a testing module for these functions? Could you show me where is it? I looked around a bit but did not find anything.

tjanez:
> @mstajdohar , @dblenkus : Do we have a testing module for these functions? Could you show me where is it? I looked around a bit but did not find anything.

@JureZmrzlikar, you can find 2 tests for `assertFile()` [here](https://github.com/genialis/resolwe/blob/master/resolwe/flow/tests/test_test_framework.py#L28).

Also, please document the new sort functionality in the `assertFile()`'s docstring.

JureZmrzlikar:
@dblenkus , please review again. I've moved tests from `resolwe/flow/tests/test_test_framework.py` to `resolwe/test/tests.py` and added tests for `sort` functionality. I've also added descriptions of `sort` parameter to `assertFile(s)` docstrings.

I don't understand why [Travis/linter tests fail](https://travis-ci.org/genialis/resolwe/jobs/355771930) with `Using deprecated method ...` for methods that are **not**  [deprecated](https://docs.python.org/3/library/unittest.html#deprecated-aliases).

resolwe/test/tests.py
```diff
@@ -1,73 +1,86 @@
...
+        with mock.patch.object(builtins, 'open', open_mock):
+            obj_mock = mock.MagicMock()
+            ProcessTestCase._assert_file(self.dummy_case, obj_mock, "", "",  # pylint: disable=protected-access
+                                         sort=True)
+
+    @mock.patch("os.path.isfile")
+    def test_assert_file_sort_func(self, isfile_mock):
+        isfile_mock.return_value = True
+        output1_file = io.BytesIO(b"A\nB\nC\n")
+        output2_file = io.BytesIO(b"A\nC\nB\n")
```
dblenkus: Change to `b"a\nc\nb\n"` to really cover the case that you are testing.

resolwe/test/tests.py
```diff
@@ -1,73 +1,86 @@
...
-    @unittest.skipIf(six.PY2, "Needs Python 3 version of the gzip module")
+    @mock.patch("os.path.isfile")
+    def test_assert_file_sort_none(self, isfile_mock):
+        isfile_mock.return_value = True
+        output1_file = io.BytesIO(b"A\nB\nC\n")
+        output2_file = io.BytesIO(b"A\nC\nB\n")
+        open_mock = mock.MagicMock(side_effect=[output1_file, output2_file])
+        with mock.patch.object(builtins, 'open', open_mock):
+            obj_mock = mock.MagicMock()
+            self.assertRaises(AssertionError, ProcessTestCase._assert_file,  # pylint: disable=protected-access
```
dblenkus: ProcessTestCase._assert_file(  # pylint: disable=protected-access
        self.dummy_case, obj_mock, "", "", sort=lambda x: str(x).lower()
    )

Also change this in tests below.

resolwe/test/tests.py
```diff
@@ -149,9 +150,11 @@ def join_side_effect(path, *paths):
...
         with mock.patch.object(os.path, 'join', join_mock):
             with mock.patch.object(os.path, 'isfile', isfile_mock):
-                with six.assertRaisesRegex(self, AssertionError,
-                                           'Output file .* missing so it was created.'):
-                    ProcessTestCase.assertJSON(dummy_case, obj_mock, storage_mock, '', 'foo.gz')
+                self.assertRaisesRegex(
+                    AssertionError,
+                    'Output file .* missing so it was created.',
+                    ProcessTestCase.assertJSON,
+                    self.dummy_case, obj_mock, storage_mock, '', 'foo.gz')
```
dblenkus: Put the closing parenthesis in to a new line.

resolwe/test/tests.py
JureZmrzlikar: Note that sorting function is just for sorting purpose, it does *not* modify file contents. So the elements in both `output1_file` and `output2_file` should be the same. With `str.lower()` we just check that a callable can be used. Any other callable would be also fine...

PR #455, opened 2018-03-21
Terminate executor if Data object does not exist

dblenkus:


resolwe/flow/executors/local/run.py
```diff
@@ -21,7 +21,6 @@ def __init__(self, *args, **kwargs):
         """Initialize attributes."""
         super(FlowExecutor, self).__init__(*args, **kwargs)
 
-        self.processes = {}
```
dblenkus: @kostko was there any good reason for this dict instead of just using `self.proc`?

kostko: No idea, I didn't add this?

PR #459, opened 2018-03-26
Track allocated memory and cores on data objects

JureZmrzlikar:


resolwe/flow/managers/dispatcher.py
```diff
@@ -738,6 +738,11 @@ def process_data_object(data):
                     # additional locks are introduced. Currently, we only take an explicit lock on
                     # the currently processing object.
                     program = self.get_execution_engine(execution_engine).evaluate(data)
+
+                    # Set allocated resources:
```
dblenkus: Move this out from the `try`/`except` block.

PR #463, opened 2018-03-29
Ignore 'fields' argument when raising errors

kostko:


dblenkus:
Yes, this is an even better idea. You should be able to configure exceptions on per-view basis as they heavily depend on what parameters are supported in the view.

PR #478, opened 2018-05-02
Migrate to Channels 2

jberci:


setup.py
```diff
@@ -77,8 +77,8 @@
         # have to pin it
         'django-filter~=1.0.0',
         'redis~=2.10.6',
-        'asgi-redis~=1.4.3',
-        'channels~=1.1.6',
+        'channels~=2.0',
```
kostko: Please depend on channels 2.1 as it has some important fixes relevant for observers.

setup.py
```diff
@@ -77,8 +77,8 @@
...
+        'channels_redis~=2.0',
```
kostko: Please depend on channels-redis 2.1.

resolwe/flow/managers/consumer.py
```diff
@@ -1,24 +1,96 @@
 """Manager Channels consumer."""
-from channels.generic import BaseConsumer
+import asyncio
 
-from . import manager, state
+import async_timeout
+from asgiref.sync import async_to_sync
+from channels.consumer import AsyncConsumer
+from channels.layers import get_channel_layer
+from channels.testing import ApplicationCommunicator
```
kostko: Hm, testing primitive used outside tests? I don't like this.

resolwe/flow/managers/consumer.py
```diff
@@ -1,24 +1,96 @@
...
+        :param message: The message to send to the manager workers.
+        """
+        packet = {
+            'type': 'control_event',  # This is used as the method name in the consumer.
+            'content': message,
+        }
+        await get_channel_layer().send(state.MANAGER_CONTROL_CHANNEL, packet)
+
+    @staticmethod
+    async def flush_queue():
```
kostko: These static methods are in the consumer but they are called from the manager? And the consumer calls the manager, getting us a full circle. Why aren't these methods in the manager if they are only used there?

jberci: Also in response to the previous comment.

I think it makes sense to segregate everything to do with Channels into the consumer. This way, the main manager code can be mostly about what the manager actually is, along with some additional utility functions (because we don't want the world to know about the consumer at all). The consumer is a specific implementation for outside infrastructure (Channels in this case) on which the manager sits, so it's sensible to make it handle everything to do with that infrastructure.

The manager itself used to have lots of test-specific functionality and that code wasn't in the tests either. Now some of it moved to the consumer, because it makes more sense there (and also the dispatcher is getting rather large, that's why I tried keeping as much Channels code out of it as possible).

As for the circle: it could be broken by moving all the low level Channels management code into the manager, but I think that would make everything much more ugly. The circle is controlled in that the manager only calls consumer methods which are call graph leaves. I could make those methods top-level instead of static.

PR #489, opened 2018-05-28
Add `dirname` and `relative_path` Jinja filters

JureZmrzlikar:


mstajdohar:
Is it unhydrate or dehydrate? Ha ha :)

JureZmrzlikar:
@dblenkus , thanks for the hint, it helped a lot!
@mstajdohar Nice suggestion, i renamed filter to `get_dehydrated_path`.
Please have another review pass.

JureZmrzlikar:
After spending some time on this, i've found a solution that works without adding a custom filter to Resolwe. Will set this as WIP until we decide what exactly do we want to do.

JureZmrzlikar:
I found a different solution, but decided to keep the PR. @dblenkus @mstajdohar please review again.

PR #490, opened 2018-06-05
Remove Python 2 heritage

dblenkus:


mstajdohar:
Nice work, @dblenkus! ❤️

PR #491, opened 2018-06-05
Add request statistics tracking to the listener

jberci:


resolwe/flow/managers/stat_utils.py
```diff
@@ -0,0 +1,162 @@
+"""Statistical utilities for the listener."""
```
tjanez: I'm not sure `resolwe.flow.managers` is the best place for such a static utility module.

@dblenkus, you probably have the best overview of the project's directory structure. Where should this reside?

resolwe/flow/tests/test_manager.py
```diff
@@ -171,3 +172,98 @@ def test_communicate(self):
         async_to_sync(manager.communicate)(run_sync=True)
 
         self.assertEqual(Data.objects.filter(status=Data.STATUS_RESOLVING).count(), 0)
+
+
+class TestListenerStats(TestCase):
```
tjanez: I would also move this to a separate module since this only really tests the `stat_utils` module and not the integration into the Listener.

resolwe/flow/managers/stat_utils.py
dblenkus: You can put it in `resolwe/flow/utils/`.

PR #496, opened 2018-06-18
Fix Redis connection leaks and switch to channels_redis_persist

jberci:


resolwe/flow/executors/__main__.py
```diff
@@ -53,3 +54,4 @@ def run_executor():
 
 if __name__ == '__main__':
     run_executor()
+    manager_commands.redis_conn = None
```
dblenkus: Add a comment why you set this to `None`.

resolwe/flow/managers/listener.py
```diff
@@ -494,35 +494,38 @@ def run(self):
...
+                            "Executor command handling error:\n\n{}",
+                            traceback.format_exc()
+                        ))
+                else:
+                    logger.error(
+                        __("Unknown executor command '{}'.", command),
+                        extra={'decoded_packet': obj}
+                    )
+        finally:
+            self._redis = None
```
dblenkus: Same here.

setup.py
```diff
@@ -78,7 +78,7 @@
         'django-filter~=1.0.0',
         'redis~=2.10.6',
         'channels~=2.1',
-        'channels_redis~=2.1',
+        'channels_redis_persist~=2.1',
```
tjanez: Change this to `~=2.2` or `~=2.2.1`.

PR #497, opened 2018-06-18
Migrate the listener to asyncio and aioredis

jberci:


resolwe/flow/managers/listener.py
```diff
@@ -36,7 +36,7 @@
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 
-class ExecutorListener(Thread):
+class ExecutorListener():
```
dblenkus: Omit parentheses.

resolwe/flow/managers/listener.py
```diff
@@ -48,58 +48,61 @@ def __init__(self, *args, **kwargs):
...
 
-    def clear_queue(self):
-        """Reset the executor queue channel to an empty state."""
-        self._redis.delete(state.MANAGER_EXECUTOR_CHANNELS.queue)
+    async def _make_connection(self):
+        """Construct a connection to Redis."""
+        return await aioredis.create_redis(
+            'redis://{}:{}'.format(
+                self._redis_params.get('host', 'localhost'),
+                self._redis_params.get('port', 56379)
```
dblenkus: I will be much easier to debug if you don't specify default values. You can add a check on start to make sure these settings are set.

PR #501, opened 2018-06-21
Generally fix documentation for the manager and executor

jberci:


resolwe/flow/executors/__init__.py
```diff
@@ -5,6 +5,8 @@
 ==============
 
 .. automodule:: resolwe.flow.executors.__main__
+.. automodule:: resolwe.flow.executors.run
```
tjanez: Could you refactor `resolwe.flow.executors.run`'s docstring so that the _Flow Executors_ title is not repeated twice?

Perhaps you could title it _Base Class_, like in _Flow Managers_ documentation?

resolwe/flow/managers/__init__.py
```diff
@@ -1,6 +1,30 @@
...
+
+Workflow workload managers.
+
+.. data:: manager
+
+    The global manager instance.
+
+    :type: :class:`~resolwe.flow.managers.dispatcher.Manager`
+
+.. automodule:: resolwe.flow.managers.dispatcher
```
tjanez: I would change the title of the `resolwe.flow.managers.dispatcher`'s docstring to _Dispatcher_.

resolwe/flow/tasks.py
```diff
@@ -8,8 +8,13 @@
 from __future__ import absolute_import
 
 import subprocess
+import sys
 
-from celery import shared_task  # pylint: disable=import-error
+if 'sphinx' not in sys.modules:
```
tjanez: Please, add a comment explaining why this work-around is needed.

resolwe/flow/managers/workload_connectors/celery.py
```diff
@@ -21,11 +21,12 @@
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 
-try:
-    import celery  # pylint: disable=unused-import
-except ImportError:
-    logger.error("Please install Celery using 'pip install celery'", file=sys.stderr)
-    sys.exit(1)
+if 'sphinx' not in sys.modules:
```
tjanez: Please, add a comment explaining why this work-around is needed.

resolwe/flow/managers/dispatcher.py
```diff
@@ -193,9 +193,9 @@ def __init__(self):
     class _SettingsManager(object):
         """Context manager for settings overrides.
 
-        Because Django's :meth:`~django.test.override_settings` is a
-        context manager, it would make the code awkward if the manager's
-        support for this wasn't also in the form of a context manager.
+        Because Django's ``django.test.override_settings`` is a context
```
tjanez: I'm curious why did you remove the `:meth:` part?

I noticed that this is not included in the docs since `_SettingsManager` is a private class, hence is not really important, but still...

tjanez:
Also, I think this work deserves a CHANGELOG entry.

resolwe/flow/managers/dispatcher.py
jberci: Sphinx was complaining about missing reference targets so I took the easy way out because Django objects are external to Resolwe; Sphinx was probably expecting to be able to resolve it with objects within the local codebase. I can look into this a bit more.

tjanez: I figured out what was wrong, we tried to reference a method, but apparently, Sphinx considers `override_settings` to be function.

Thus, changing ``:meth:`~django.test.override_settings` `` to `` func:`django.test.override_settings` `` works like a charm.

PR #504, opened 2018-07-03
Refactor Resolwe API

JureZmrzlikar:


resolwe/flow/serializers.py
```diff
@@ -1,318 +0,0 @@
-""".. Ignore pydocstyle D400.
```
dblenkus: Change commit message to: `Split flow's serializers into module`

resolwe/flow/serializers/process.py
```diff
@@ -11,8 +11,26 @@ class Meta:
...
-        fields = ('slug', 'name', 'data_name', 'version', 'type', 'flow_collection', 'category',
-                  'persistence', 'description', 'input_schema', 'output_schema', 'requirements',
-                  'run', 'scheduling_class') + update_protected_fields + read_only_fields
+        read_only_fields = (
+            'id',
+            'created',
+            'modified',
+        )
+        update_protected_fields = (
+            'name',
```
dblenkus: Is there any logic how these are ordered? If not, you can order them alphabetically.

resolwe/flow/serializers/base.py
```diff
@@ -35,20 +35,22 @@ class of child class.
     """
 
     contributor = ContributorSerializer()
+    name = serializers.CharField(required=False)
```
dblenkus: This commit in not an update, so you should use 'prettify' or 'refactor'.

resolwe/flow/views/data.py
```diff
@@ -184,6 +159,29 @@ def get_or_create(self, request, *args, **kwargs):
...
         return self.create(request, *args, **kwargs)
 
+    def perform_get_or_create(self, request, *args, **kwargs):
+        """Perform "get_or_create" - return existing object if found."""
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        process = serializer.validated_data.get('process')
+        process_input = request.data.get('input', {})
+
+        # use default values if they are not given
```
dblenkus: This can/should be moved to model's utils in the future.

resolwe/flow/serializers/fields.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __init__(self, slug_field=None, **kwargs):
+        """Initialize attributes."""
+        assert slug_field is not None, 'The `slug_field` argument is required.'
+        self.slug_field = slug_field
+        super(ResolweSlugRelatedField, self).__init__(**kwargs)
+
+    def to_internal_value(self, data):
+        """Convert to internal value."""
+        user = getattr(self.context.get('request', None), 'user', None)
```
dblenkus: You should probably swap Django's `AnonymousUser` for guardian's one, as it is done [here](https://github.com/genialis/resolwe/pull/504/commits/644739a977cc695a7b856e01775159e65aa7c490#diff-363866a33132f2ba0b58211776554c4dR27).
How does `get_objects_for_user` work with `None` as the user?

resolwe/flow/serializers.py
JureZmrzlikar: From Python docs: "[A module is a file containing Python definitions and statements](https://docs.python.org/3/tutorial/modules.html)". If so, it would be better to rename commit message to "Split flow's serializers into module**s**"(plural). Would you agree?

resolwe/flow/serializers/process.py
JureZmrzlikar: No logic, but the order is preserved in the output data. Do we care about the order?

resolwe/flow/views/data.py
JureZmrzlikar: In the future or now? :-)

resolwe/flow/serializers/collection.py
```diff
@@ -9,14 +9,16 @@
...
+from .fields import ResolweSlugRelatedField
 
 
 class CollectionSerializer(ResolweBaseSerializer):
     """Serializer for Collection objects."""
 
     settings = ProjectableJSONField(required=False)
     descriptor = ProjectableJSONField(required=False)
+    descriptor_schema = ResolweSlugRelatedField(
+        slug_field='slug', required=False, queryset=DescriptorSchema.objects.all())
```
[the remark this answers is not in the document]
JureZmrzlikar: Any particular reason for this. Linters don't seem to bother?

resolwe/flow/serializers/fields.py
```diff
@@ -0,0 +1,51 @@
...
+        queryset = self.get_queryset()
+        app_label = queryset.model._meta.app_label  # pylint: disable=protected-access
+        model_name = queryset.model._meta.model_name  # pylint: disable=protected-access
+        permission = '{}.view_{}'.format(app_label, model_name)
+        try:
+            return get_objects_for_user(user, permission, queryset.filter(**{self.slug_field: data})).latest()
+        except ObjectDoesNotExist:
+            self.fail('does_not_exist', slug_name=self.slug_field, value=smart_text(data), model_name=model_name)
```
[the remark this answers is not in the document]
JureZmrzlikar: Any reason for this? I can understand it from readabilty view but the line-length limit is 120, right?

resolwe/flow/serializers/fields.py
JureZmrzlikar: Thanks for pointing this out: `get_objects_for_user` will fail if `user=None`. What exactly do you mean by "swap Django's AnonymousUser for guardian's one"? I don't understand. What is the difference?

resolwe/flow/serializers.py
dblenkus: You are right, or you can use `package` instead.

resolwe/flow/views/data.py
dblenkus: You can do it now. [Data model](https://github.com/genialis/resolwe/blob/master/resolwe/flow/models/data.py#L353) uses the same code.

resolwe/flow/views/storage.py
```diff
@@ -12,4 +13,6 @@ class StorageViewSet(mixins.RetrieveModelMixin,
 
     queryset = Storage.objects.all().prefetch_related('contributor')
     serializer_class = StorageSerializer
-    filter_fields = ('contributor', 'name', 'created', 'modified', 'slug')
+    filter_class = StorageFilter
```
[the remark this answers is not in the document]
dblenkus: Then just remove it (including `ListModelMixin `) and we will see what happens.

Don't forget to mention this in change log.

JureZmrzlikar:
@dblenkus I've adressed your comments, please have another review pass.

resolwe/flow/serializers/fields.py
dblenkus: `AnonymouseUser` in Django is a different class than the normal `User`, so you cannot permissions to it. Guardian solves this with creating an actual user object, which is treated as a public user.

Actually tis may already be solved in `get_objects_for_user`, but please make sure that it works as intended.

resolwe/flow/models/data.py
```diff
@@ -10,7 +10,8 @@
 from django.db import models, transaction
 
 from resolwe.flow.expression_engines.exceptions import EvaluationError
-from resolwe.flow.utils import dict_dot, get_data_checksum, iterate_fields, iterate_schema
+from resolwe.flow.models.utils import fill_with_defaults
```
dblenkus: This commit includes several changes:

- `ResolweSlugRelatedField`
- `fill_with_defaults`
- `perform_get_or_create`

Please split them in separate commits.

JureZmrzlikar: Done.

resolwe/flow/views/mixins.py
```diff
@@ -44,27 +22,21 @@ class ResolweCreateModelMixin(mixins.CreateModelMixin):
...
-                    status=status.HTTP_400_BAD_REQUEST)
-
+    def define_contributor(self, request):
+        """Define contibutor by adding it to request.data."""
         if request.user.is_anonymous:
             request.data['contributor'] = get_anonymous_user().pk
         else:
             request.data['contributor'] = request.user.pk
 
+        return request
```
dblenkus: You don't need to return modified object as original object (passed to the function) is also changed.

resolwe/flow/views/mixins.py
```diff
@@ -44,27 +22,21 @@ class ResolweCreateModelMixin(mixins.CreateModelMixin):
...
+
+    def create(self, request, *args, **kwargs):
+        """Create a resource."""
+        request = self.define_contributor(request)
+
         try:
             return super().create(request, *args, **kwargs)
-
```
dblenkus: Leave an empty line after the `return` statement.

resolwe/flow/serializers/fields.py
```diff
@@ -0,0 +1,51 @@
+"""Resolwe custom serializer fields."""
+from django.core.exceptions import ObjectDoesNotExist
+from django.utils.encoding import smart_text
+
+from rest_framework.relations import RelatedField
+
+from resolwe.permissions.shortcuts import get_objects_for_user
+
+
+class ResolweSlugRelatedField(RelatedField):
```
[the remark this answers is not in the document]
dblenkus: Have you added tests?

PR #505, opened 2018-07-04
Add debug log messages to executor and listener

dblenkus:


resolwe/flow/executors/docker/run.py
```diff
@@ -184,12 +189,32 @@ def new_volume(kind, base_dir_name, volume, path=None, read_only=True):
...
-            universal_newlines=True)
+            shlex.split(docker_command),
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            universal_newlines=True,
+            bufsize=1,
+        )
+
+        # Wait for Docker container to start.
```
tjanez: Can you rewrite the comment to be more along the lines of:

```
# Write something to stdout to measure container's start-up time.
```

resolwe/flow/managers/listener.py
```diff
@@ -606,6 +606,7 @@ def check_critical_load(self):
             self.check_critical_load()
             _, item = ret
             try:
+                logger.debug(__("Got command from executor: {}", item))
```
tjanez: Does this work?

According to the next line, `item` is a `bytes` object that needs to be decoded.

resolwe/flow/executors/run.py
```diff
@@ -161,6 +161,8 @@ def _run(self, data_id, script):
             stdout = self.get_stdout()
             while True:
                 line = stdout.readline()
+                logger.debug("Process's output: {}".format(line.strip()))
```
tjanez: Process's -> Process'

PR #514, opened 2018-08-06
Convert collection/entity endpoints to ES

kostko:


resolwe/flow/elastic_indexes/collection.py
```diff
@@ -0,0 +1,61 @@
...
+    class Meta:
+        """Meta class for collection search document."""
+
+        index = 'collection'
+
+
+class CollectionIndexMixin:
+    """Mixin for indices for collection objects used in ``CollectionDocument``."""
+
+    def extract_descriptor(self, obj):
```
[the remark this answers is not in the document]
kostko: But this method should be considered exported to subclasses as some may want to extract data from other descriptors (e.g., entity extracts descriptors of all child objects).

resolwe/flow/views/collection.py
```diff
@@ -34,9 +39,49 @@ class CollectionViewSet(ResolweCreateModelMixin,
     )
     serializer_class = CollectionSerializer
     permission_classes = (get_permissions_class(),)
-    filter_class = CollectionFilter
-    ordering_fields = ('id', 'created', 'modified', 'name')
-    ordering = ('id',)
+    document_class = CollectionDocument
+
+    filtering_fields = ('id', 'slug', 'name', 'created', 'modified', 'contributor', 'owners', 'text')
```
JureZmrzlikar: I don't fully understand ES syntax so this might be a stupid question: will it be still possible to filter by data (id) on collection endpoint? A common question for me is: "I have a data ID and want to know in which collections it is located?" Will such queries still be supported and if not, what is the suggested alternative?

kostko: Hm, missed that part, so currently this is not supported. Also it seems that this is not covered by any tests as the tests are passing.

> I have a data ID and want to know in which collections it is located?

Is this the only reason why such filters existed? Because to me it would be much more logical to fetch the given data item and get a list of collections it is in.

lukaw3d: (frontend is also using `api.Sample.queryOne({ data: data.id })` to get samples in which data id is located)

JureZmrzlikar: > it would be much more logical to fetch the given data item and get a list of collections it is in

Totally agree, even better. But this would require modifying `DataSerializer` which seems out of context for this PR... So someone has to do it sometimes soon. In the end, i don't care how, as long as i can get the desired info... ;-)

kostko: @JureZmrzlikar done in this PR.

kostko: @lukaw3d note that with this PR, you can get the list of samples directly on the data object by passing `hydrate_entities=1` query argument to the data detail endpoint. Similar for collections using `hydrate_collections=1`. (By default it returns only primary keys instead of full objects.)

PR #520, opened 2018-08-16
Add Python execution engine

kostko:


resolwe/flow/execution_engines/workflow/__init__.py
```diff
@@ -15,6 +16,36 @@ class ExecutionEngine(BaseExecutionEngine):
 
     name = 'workflow'
 
+    def discover_process(self, path):
```
mstajdohar: This is pretty much the same as for bash. Couldn't you reuse the code?

resolwe/flow/tests/processes/python.py
```diff
@@ -0,0 +1,31 @@
...
+        integer = IntegerField(label="My integer")
+        my_float = FloatField(label="My float")
+        my_json = JsonField(label="Blah blah")
+        my_optional = StringField(label="Optional", required=False, default='default value')
+
+    class Output:
+        my_output = StringField(label="My output")
+
+    def run(self, inputs, outputs):
+        print(inputs)
```
mstajdohar: This solution will not give us code completion for inputs and outputs, right?

resolwe/process/fields.py
```diff
@@ -0,0 +1,286 @@
...
+    # TODO: Fix the underlying field into basic:float once that is renamed.
+    field_type = 'basic:decimal'
+
+    def to_python(self, value):
+        if value is not None:
+            return float(value)
+
+
+class DateField(Field):
+    field_type = 'basic:date'
```
mstajdohar: Should we support Python `datetime.date`?

resolwe/process/fields.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+class DateTimeField(Field):
+    field_type = 'basic:datetime'
```
mstajdohar: Should we support Python `datetime.datetime`?

resolwe/process/fields.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+class UrlField(StringField):
+    # Url types.
+    DOWNLOAD = 'download'
+    VIEW = 'view'
+    LINK = 'link'
+
+    URL_TYPES = (DOWNLOAD, VIEW, LINK)
```
mstajdohar: I like how you've handled the three types of URL field.

resolwe/process/fields.py
```diff
@@ -0,0 +1,286 @@
...
+    def get_field_type(self):
+        return 'basic:url:{}'.format(self.url_type)
+
+
+class SecretField(Field):
+    field_type = 'basic:secret'
+
+
+class FileField(Field):
+    field_type = 'basic:file'
```
mstajdohar: This will require some thought. One thing that comes to mind is that `__repr__` and `__str__` could return an empty value.

mstajdohar:
Don't forget a change log entry. I also suggest to include the new docstrings in reference documentation.

mstajdohar:
This test fails randomly. I've got [the same error](https://travis-ci.org/genialis/resolwe/jobs/439338083) when I was making the release where just changed the version and the change log.

PR #521, opened 2018-08-17
WIP: Custom serializer field: DictRelatedField

JureZmrzlikar:


resolwe/flow/serializers/fields.py
```diff
@@ -11,63 +11,65 @@
...
+        else:
+            self.fail('no_data', name=self.field_name)
+
         try:
             return get_objects_for_user(
                 user,
                 permission,
-                queryset.filter(**{self.slug_field: data}),
+                queryset.filter(**kwargs),
             ).latest()
```
mstajdohar: Is this OK? I know we set `get_latest_by` in the base model. What if one of the models overrides that, or if we use this field in other use cases at some point. Do we want to be more explicit here and pass `version` to `latest()`?

mstajdohar:
Don't forget about the change log—very important in this case.

resolwe/flow/serializers/fields.py
```diff
@@ -11,63 +11,65 @@
 from .descriptor import DescriptorSchemaSerializer
 
 
-class ResolweSlugRelatedField(RelatedField):
+class DictRelatedField(RelatedField):
```
mstajdohar: Should we call this `ResolweRelatedField`? That would be a dict field that can contain either id or slug, and can be hydrated or not.

JureZmrzlikar:
Closed in favor of [#631](https://github.com/genialis/resolwe/pull/631)

PR #522, opened 2018-08-17
Skip current_user_permissions field when applying projection

JureZmrzlikar:


resolwe/permissions/mixins.py
```diff
@@ -49,6 +49,11 @@ def to_representation(serializer_self, instance):  # pylint: disable=no-self-arg
                 """Object serializer."""
                 # TODO: These permissions queries may be expensive. Should we limit or optimize this?
                 data = super().to_representation(instance)
+
+                if 'fields' in self.request.query_params \
```
mstajdohar: I would reformat this to `return data` only once, for example:
```
if 'fields' not in self.request.query_params or \
        'current_user_permissions' in self.request.query_params['fields']:
    data['current_user_permissions'] = get_object_perms(instance, self.request.user)
```

resolwe/permissions/mixins.py
```diff
@@ -47,9 +47,12 @@ def get_fields(serializer_self):  # pylint: disable=no-self-argument
 
             def to_representation(serializer_self, instance):  # pylint: disable=no-self-argument
                 """Object serializer."""
-                # TODO: These permissions queries may be expensive. Should we limit or optimize this?
                 data = super().to_representation(instance)
-                data['current_user_permissions'] = get_object_perms(instance, self.request.user)
+
+                if 'fields' not in self.request.query_params \
```
dblenkus: Use brackets instead of backslash.

From PEP8:
> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

resolwe/permissions/mixins.py
```diff
@@ -47,9 +47,12 @@ def get_fields(serializer_self):  # pylint: disable=no-self-argument
...
+                        or 'current_user_permissions' in self.request.query_params.get('fields', []):
```
dblenkus: You can use `self.request.query_params['fields']` as the key is always present - if not, evaluation is broken at first condition.

PR #523, opened 2018-08-17
Fix premature counter zeroing and aggressive Channels flushing in tests

jberci:
Communicate calls weren't protected by a counter claim, so it could
happen that a communicate command was issued, then the synchronization
counter dropped to 0 before the command was picked up by Channels on the
other side. Fixed by increasing the count when the command is generated,
and decreasing it once the _data_scan function finishes.

There was also a problem with parallel testing. The Channels Redis layer
is global (per server) and its flush command clears all asgi-related
channels from the server. In the case of parallelized test suites, this
means that one process can still be doing work while another finished -
upon finishing, it will "clean up after itself" by flushing, which will
also clean keys used by other testing processes, interrupting their flow
and likely hanging them (if the deleteion happens e.g. between a
communicate command being generated and then received, since the
relevant counter claim won't be undone). The somewhat awkward solution
is to make the Manager infrastructure clean up only specific keys on
startup and rely on expiration for others.

### DIFF
--- a/resolwe/flow/managers/listener.py
+++ b/resolwe/flow/managers/listener.py
@@ -111,7 +111,15 @@ class ExecutorListener:
         """Reset the executor queue channel to an empty state."""
         conn = await self._make_connection()
         try:
-            await conn.delete(state.MANAGER_EXECUTOR_CHANNELS.queue)
+            script = """
+                local keys = redis.call('KEYS', ARGV[1])
+                redis.call('DEL', unpack(keys))
+            """
+            await conn.eval(
+                script,
+                keys=[],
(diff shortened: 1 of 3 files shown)


resolwe/flow/managers/listener.py
```diff
@@ -549,6 +557,7 @@ def _make_stats(self):
         try:
             serialized = json.dumps(snapshot)
             await self._call_redis(aioredis.Redis.set, state.MANAGER_LISTENER_STATS, serialized)
+            await self._call_redis(aioredis.Redis.expire, state.MANAGER_LISTENER_STATS, 3600)
```
tjanez: Please, add a comment above this line explaining why this is needed and why here.

mstajdohar:
@jberci Please rebase and merge.

PR #524, opened 2018-08-26
Relations v2

dblenkus:


resolwe/flow/models/entity.py
```diff
@@ -58,27 +59,59 @@ class Meta(BaseCollection.Meta):
...
+    UNIT_HOUR = 'hr'
+    UNIT_CHOICES = (
+        (UNIT_HOUR, 'Hour'),
+    )
+
+    def validate_unique(self, exclude=None):
+        """Check unique constraints on the model.
+
+        Separate check for the combination of ``collection`` and ``category``
+        is required as ``category`` is optional and thus
```
JureZmrzlikar: Could all this be avoided by just providing a default value for `category`and then use `Meta.unique_together`?

resolwe/flow/models/entity.py
```diff
@@ -58,27 +59,59 @@ class Meta(BaseCollection.Meta):
...
+
     #: type of the relation
     type = models.ForeignKey('RelationType', on_delete=models.PROTECT)
 
     #: collection in which relation is
     collection = models.ForeignKey('Collection', null=True, on_delete=models.SET_NULL)
 
-    #: entities in the relation
-    entities = models.ManyToManyField(Entity, through='PositionInRelation')
+    #: enteties in the relation
```
JureZmrzlikar: `enteties` --> `entities`

resolwe/flow/models/entity.py
```diff
@@ -48,7 +49,7 @@ def __str__(self):
 class Relation(BaseModel):
     """Relations between entities."""
 
-    class Meta(BaseCollection.Meta):
+    class Meta(BaseModel.Meta):
```
JureZmrzlikar: [Django coding style guidelines](https://docs.djangoproject.com/en/2.0/internals/contributing/writing-code/coding-style/#model-style) suggest that `Meta` class should be placed after field names definitions. I know this is inconsistent with how things currently are in this project, but still, we may want to gradually move in that direction...

resolwe/flow/migrations/0016_refactor_relations_2.py
```diff
@@ -0,0 +1,24 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def position_to_label(apps, schema_editor):
+    RelationPartition = apps.get_model('flow', 'RelationPartition')
+    for partition in RelationPartition.objects.all():
+        if partition.position:
+            partition.label = partition.position
```
mstajdohar: Do you expect issues with the type of old `partition.position`? This was JSONField but the new `partition.label` accepts strings. I suggest you check existing values in DB.

resolwe/flow/models/entity.py
```diff
@@ -58,27 +59,59 @@ class Meta(BaseCollection.Meta):
             ("owner_relation", "Is owner of the relation"),
         )
 
+    UNIT_HOUR = 'hr'
+    UNIT_CHOICES = (
```
mstajdohar: @darlingt What kind-of units do you expect in time-series? Hour, seconds, days? We would like to code to be future-ready so we are thinking ahead here.

resolwe/flow/models/entity.py
mstajdohar: Yeah, I really don't like this. Remind me why do we have to support NULL category? Could we have an empty string by default?

resolwe/flow/models/entity.py
```diff
@@ -58,27 +59,59 @@ class Meta(BaseCollection.Meta):
...
+
+        duplicate_exists = Relation.objects.exclude(id=self.id).filter(
+            collection_c=self.collection, category=self.category, category__isnull=False
+        ).exists()
+
+        if duplicate_exists:
+            raise ValidationError("Combination of 'collection' and 'category' must be unique.")
+
     #: type of the relation
     type = models.ForeignKey('RelationType', on_delete=models.PROTECT)
```
mstajdohar: We could simply use a `CharField` with choices now. We do not need the `ordered` field of the `RelationType` anymore, since we have designed the `RelationPartition` with `label` and `position` fields.

resolwe/flow/models/entity.py
```diff
@@ -58,27 +59,59 @@ class Meta(BaseCollection.Meta):
...
 
     #: collection in which relation is
     collection = models.ForeignKey('Collection', null=True, on_delete=models.SET_NULL)
```
mstajdohar: I should probably delete the relations if I delete the collection. Otherwise it may break the `validate_unique` constraint. This would cause deleting the collection to fail... Say you delete a collection with Treatment category. Then you delete another collection with Treatment category. This will break the unique constraint because you can not have two Treatment relations with NULL collection.

resolwe/flow/models/entity.py
```diff
@@ -48,7 +49,7 @@ def __str__(self):
 class Relation(BaseModel):
     """Relations between entities."""
```
mstajdohar: Include the model documentation in the reference. We need to start documenting out DB models.

resolwe/flow/models/entity.py
```diff
@@ -58,27 +59,59 @@ class Meta(BaseCollection.Meta):
...
 
 
-class PositionInRelation(models.Model):
+class RelationPartition(models.Model):
     """Intermediate model to connect Entities with Relations."""
 
+    class Meta:
+        """RelationPartition Meta options."""
+
+        unique_together = ("entity", "relation")
```
mstajdohar: Finally! This will prohibit many inconsistencies, thanks.

resolwe/flow/models/entity.py
```diff
@@ -58,27 +59,59 @@ class Meta(BaseCollection.Meta):
...
+
     #: entity in the relation
     entity = models.ForeignKey(Entity, on_delete=models.CASCADE)
 
     #: relation to which entity belongs
     relation = models.ForeignKey(Relation, on_delete=models.CASCADE)
 
-    #: position of the entity in the relation (relevant if `relation.type.ordered` is True)
-    position = JSONField(null=True, blank=True)
+    #: position of the entity in the relation (relevant if `relation.type.odrered` is True)
```
mstajdohar: > relevant if `relation.type.odrered` is True

I don't think this is needed anymore. Position is relevant when it is used. If position is NULL, the label is relevant.

resolwe/flow/models/entity.py
```diff
@@ -58,27 +59,59 @@ class Meta(BaseCollection.Meta):
...
+    position = models.PositiveSmallIntegerField(null=True, blank=True, db_index=True)
```
mstajdohar: I hope `PositiveSmallIntegerField` will cover all the use cases. I agree we start with that and change it only if there is a viable use case.

resolwe/flow/serializers/relation.py
```diff
@@ -59,22 +57,36 @@ def get_fields(self):
...
+            )
+
     def create(self, validated_data):
         """Create ``Relation`` object and add ``Entities``."""
-        # `entities` field is renamed to `positioninrelation_set` based on source of nested serializer
+        # `entities` field is renamed to `relationpartition_set` based on source of nested serializer
+        entities = validated_data.pop('relationpartition_set', {})
 
-        entities = validated_data.pop('positioninrelation_set', {})
+        with transaction.atomic():
```
mstajdohar: +1 Exactly what I was looking for here.

resolwe/flow/serializers/relation.py
```diff
@@ -4,29 +4,27 @@
...
-        fields = ('entity', 'position')
+        model = RelationPartition
+        fields = ('id', 'entity', 'position', 'label')
 
 
 class RelationSerializer(ResolweBaseSerializer):
     """Serializer for Relation objects."""
 
-    entities = PositionInRelationSerializer(source='positioninrelation_set', many=True)
+    entities = RelationPartitionSerializer(source='relationpartition_set', many=True)
```
mstajdohar: It's a bit misleading to call this `entities`, since this is not a list of `entities` but of RelationPartition objects which include the entity, label and position.

mstajdohar: Don't ask me how to call this better :) Maybe @jberci can help ;)

resolwe/flow/serializers/relation.py
```diff
@@ -59,22 +57,36 @@ def get_fields(self):
...
 
-        # Prevent "Failed to populate slug Relation.slug from name" output
-        validated_data['name'] = 'Relation'
+        return instance
+
+    def update(self, instance, validated_data):
+        # `entities` field is renamed to `relationpartition_set` based on source of nested serializer
+        entities = validated_data.pop('relationpartition_set', {})
+
+        instance = super().update(instance, validated_data)
```
mstajdohar: Why not have this in `transaction.atomic()`?

resolwe/flow/views/relation.py
```diff
@@ -48,9 +48,9 @@ def _filter_queryset(self, queryset):
 
         id_ = query_params.get('id', None)
         relation_type = query_params.get('type', None)
-        label = query_params.get('label', None)
+        category = query_params.get('category', None)
```
mstajdohar: Don't forget to update the function docstring.

mstajdohar: I suspect we will have to rethink the filtering soon. @JureZmrzlikar, @hadalin Think what you need for your use cases.

resolwe/flow/models/entity.py
dblenkus: `NULL` means that you are not using categories, so you can have multiple relations of the same type. Setting a default value (e.g. empty string) would restrain this, which is, in my opinion, not a good option.

Alternative would be to make the category mandatory.

resolwe/flow/models/entity.py
dblenkus: This should be a model, so different applications (e.g. resolwe-bio) can define different types of relations. You cannot predefine all of them here.

resolwe/flow/serializers/relation.py
dblenkus: Good point :)

resolwe/flow/models/entity.py
darlingt: seconds (s), minutes (min), hours (h), days (d), weeks (wk).

Weeks is not a standard ISO unit of time so I've used a common abbreviation. All others are ISO units of time and use the ISO abbreviation. I added in weeks because I think some users may use this field for age of their mice.

resolwe/flow/models/entity.py
mstajdohar: > Alternative would be to make the category mandatory.

Let's do that. I would not like the to offer no-category alternative to the users.

resolwe/flow/models/entity.py
mstajdohar: True

resolwe/flow/views/relation.py
```diff
@@ -59,23 +59,24 @@ def _filter_queryset(self, queryset):
         if relation_type:
             queryset = queryset.filter(type__name=relation_type[0])
 
-        if label:
-            queryset = queryset.filter(label=label[0])
+        if category:
+            queryset = queryset.filter(category=category[0])
```
hadalin: Can you make this filter case insensitive?

resolwe/flow/models/entity.py
```diff
@@ -58,27 +59,59 @@ class Meta(BaseCollection.Meta):
...
+        """Check unique constraints on the model.
+
+        Separate check for the combination of ``collection`` and ``category``
+        is required as ``category`` is optional and thus
+        ``Meta.unique_together`` cannot be used.
+        """
+        super().validate_unique(exclude)
+
+        duplicate_exists = Relation.objects.exclude(id=self.id).filter(
+            collection_c=self.collection, category=self.category, category__isnull=False
```
hadalin: Can you make the category case insensitive?

resolwe/flow/serializers/relation.py
```diff
@@ -40,12 +37,12 @@ class Meta:
         )
         update_protected_fields = (
             'contributor',
-            'entities',
             'type',
         )
         fields = read_only_fields + update_protected_fields + (
```
hadalin: Should `slug` also be on this list? It appears in other serializer's.

resolwe/flow/migrations/0017_refactor_relations_3.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.14 on 2018-08-06 02:28
+from __future__ import unicode_literals
```
mstajdohar: Not needed.

mstajdohar:
Don't forget about the change log.

resolwe/flow/serializers/relation.py
```diff
@@ -4,30 +4,27 @@
...
+        fields = ('id', 'entity', 'position', 'label')
 
 
 class RelationSerializer(ResolweBaseSerializer):
     """Serializer for Relation objects."""
 
-    entities = PositionInRelationSerializer(source='positioninrelation_set', many=True)
-    collection = serializers.PrimaryKeyRelatedField(queryset=Collection.objects.all(), required=True)
+    entities = RelationPartitionSerializer(source='relationpartition_set', many=True)
+    collection = serializers.PrimaryKeyRelatedField(queryset=Collection.objects.all())
```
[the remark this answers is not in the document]
dblenkus: This is still required, but does not need to be specified explicitly, as `null=True` has been removed.

docs/CHANGELOG.rst
```diff
@@ -14,6 +14,15 @@ Changed
...
   to use Elasticsearch
+- **BACKWARD INCOMPATIBLE:** Refactor ``Relation`` model, which includes:
+
+  - renaming ``position`` to ``partiiton``
+  - renaming ``label`` to ``category`` and making it required
+  - adding ``unit``
+  - making ``collection`` field required
+  - requireing uniqe combination of ``collection`` and ``category``
+  - renaming partition's ``position`` to ``label``
+  - adding (integer) ``position`` to partition (used for sorting)
```
mstajdohar: Do you need to add one extra line after the sub-list?

tjanez: I would say no. Also, we don't add these extra lines in [Resolwe Bio's CHANGELOG](https://raw.githubusercontent.com/genialis/resolwe-bio/master/docs/CHANGELOG.rst).

resolwe/flow/models/entity.py
```diff
@@ -46,39 +46,110 @@ def __str__(self):
...
+
+    #: category of the relation
+    category = models.CharField(max_length=100)
+
+    #: unit used in the partitions' positions (where applicable, e.g. for serieses)
+    unit = models.CharField(max_length=3, choices=UNIT_CHOICES, null=True, blank=True)
+
+    def save(self, *args, **kwargs):
+        """Lowercase the ``category`` field and save the model."""
+        self.category = self.category.lower()
```
hadalin: Is it ok to save category as lowercase? You loose some information.

hadalin: Maybe the category should be saved as it is and fix the filter to be case insensitive.

mstajdohar: This was fixed in #532

docs/CHANGELOG.rst
mstajdohar: So, here's the catch. This works:
```
- First bullet

  - Sub-bullet 1
  - Sub-bullet 2
- Second bullet
```
But this doesn't:
```
- First bullet

  - Sub-bullet 1
  - Sub-bullet 2
  Some more text under the first bullet
```
The [standard says](http://docutils.sourceforge.net/docs/user/rst/quickref.html#bullet-lists) that a blank line is required before the first item and after the last.

tjanez: Ok, the text under the sub-bullets is a nice thing to have.

We can reformat the CHANGELOG (and the rest of the documentation, if such cases exist?) to include this extra blank line.

PR #527, opened 2018-08-30
Slug

hadalin:


resolwe/flow/serializers/base.py
```diff
@@ -56,3 +59,10 @@ def __init__(self, instance=None, data=empty, **kwargs):
     def request(self):
         """Extract request object from serializer context."""
         return self.context.get('request', None)
+
+    def save(self):
+        """Override save method to catch handled errors and repackage them as 400 errors."""
+        try:
+            super(ResolweBaseSerializer, self).save()
+        except DatabaseError as database_error:
```
hadalin: Not the best idea.

A better solution would be to raise a specific resolwe error [here](https://github.com/genialis/resolwe/blob/7b8ee3e4de21affa7cf18d73e9f0acb5556c06ec/resolwe/flow/models/fields.py#L178) and catch it here. That would entail an `excpetions` package with that error defined.

resolwe/flow/serializers/base.py
```diff
@@ -56,3 +59,10 @@ def __init__(self, instance=None, data=empty, **kwargs):
     def request(self):
         """Extract request object from serializer context."""
         return self.context.get('request', None)
+
+    def save(self):
+        """Override save method to catch handled errors and repackage them as 400 errors."""
+        try:
+            return super(ResolweBaseSerializer, self).save()
```
dblenkus: `return super().save(**kwargs)`

resolwe/flow/serializers/base.py
```diff
@@ -56,3 +59,10 @@ def __init__(self, instance=None, data=empty, **kwargs):
     def request(self):
         """Extract request object from serializer context."""
         return self.context.get('request', None)
+
+    def save(self):
```
dblenkus: `def save(self, **kwargs):`

resolwe/flow/serializers/base.py
dblenkus: Actually a really bad idea :).

Create a custom error and put it in `flow/exceptions.py`.

resolwe/flow/serializers/base.py
```diff
@@ -36,7 +36,7 @@ class of child class.
 
     contributor = ContributorSerializer()
     name = serializers.CharField(required=False)
-    slug = serializers.CharField(required=False)
+    slug = serializers.CharField(required=False, allow_null=True)  # Allow null to support slug autogeneration
```
dblenkus: Put the comment in a separate line above. And don't forget to put a dot at the end.

resolwe/flow/tests/test_api.py
```diff
@@ -301,6 +301,35 @@ def test_change_descriptor_schema(self):
...
+        response = self.collection_detail_viewset(request, pk=collection1.pk)
+        self.assertEqual(response.data['slug'], 'collection')
+
+        collection3 = Collection.objects.create(name="Collection", contributor=self.contributor)
+        self.assertEqual(collection3.slug, 'collection-3')
+
+        request = factory.patch(self.detail_url(collection2.pk), {'name': 'Collection', 'slug': None}, format='json')
+        force_authenticate(request, self.admin)
+        response = self.collection_detail_viewset(request, pk=collection2.pk)
+        self.assertEqual(response.data['slug'], 'collection-4')
```
dblenkus: This is probably not the desired behaviour and should be fixed at some point, so i think it is not a good idea to include it in tests.

hadalin: Is it the problem that the new slug should be `collection-2`?

hadalin: Fixed in https://github.com/genialis/resolwe/pull/527/commits/97a6c69854191f3024e4b8ae6aaa006a34d3ea16

resolwe/flow/serializers/base.py
hadalin: Fixed in https://github.com/genialis/resolwe/pull/527/commits/b45d2f9abff4c82ab13a43515d89c99afad5c738

resolwe/flow/tests/test_api.py
```diff
@@ -301,6 +301,30 @@ def test_change_descriptor_schema(self):
...
+        collection2 = Collection.objects.create(name="Collection", contributor=self.contributor)
+        self.assertEqual(collection1.slug, 'collection')
+        self.assertEqual(collection2.slug, 'collection-2')
+
+        request = factory.patch(self.detail_url(collection1.pk), {'name': 'Collection', 'slug': None}, format='json')
+        force_authenticate(request, self.admin)
+        response = self.collection_detail_viewset(request, pk=collection1.pk)
+        self.assertEqual(response.data['slug'], 'collection')
+
+        collection3 = Collection.objects.create(name="Collection", contributor=self.contributor)
```
dblenkus: Now you can remove this.

resolwe/flow/tests/test_api.py
```diff
@@ -301,6 +301,30 @@ def test_change_descriptor_schema(self):
...
+        self.assertEqual(collection3.slug, 'collection-3')
+
+        request = factory.patch(self.detail_url(collection2.pk), {'slug': 'collection-2'}, format='json')
```
dblenkus: This should probably be `collection-3`.

resolwe/flow/serializers/base.py
```diff
@@ -61,9 +62,9 @@ def request(self):
...
         return self.context.get('request', None)
 
-    def save(self):
+    def save(self, **kwargs):
         """Override save method to catch handled errors and repackage them as 400 errors."""
         try:
-            return super(ResolweBaseSerializer, self).save()
-        except DatabaseError as database_error:
-            raise ParseError(database_error)
+            return super(ResolweBaseSerializer, self).save(**kwargs)
```
dblenkus: `ResolweBaseSerializer, self` part is no longer needed in Pyhton 3.

resolwe/flow/tests/test_api.py
dblenkus: It is not consistent with the first use-case in this test, where slugs stays the same. We should at least unify the behaviour.

PR #530, opened 2018-09-05
Retire obsolete processes

mstajdohar:


resolwe/flow/tests/test_commands.py
```diff
@@ -77,6 +77,34 @@ def test_inherit_perms(self):
...
+        out, err = StringIO(), StringIO()
+        call_command('register', stdout=out, stderr=err)
+
+        # Check that all registered processes are active
+        initial_processes = Process.objects.count()
+        active_processes = Process.objects.filter(is_active=True).count()
+        self.assertGreater(initial_processes, 0)
+        self.assertEqual(initial_processes, active_processes)
+
+        # Check that retired processes without data are removed and
```
dblenkus: Add a test that `register` without `--retire` doesn't do this.

resolwe/flow/views/data.py
```diff
@@ -165,6 +165,11 @@ def perform_get_or_create(self, request, *args, **kwargs):
 
     def perform_create(self, serializer):
         """Create a resource."""
+        process = serializer.validated_data.get('process')
+        if not process.is_active:
+            raise exceptions.ParseError('Process retired (id: {}, slug: {}/{}).'.format(
```
dblenkus: New line after opening bracket of `ParseError` and before closing one.

resolwe/flow/tests/test_commands.py
mstajdohar: done

resolwe/flow/views/data.py
mstajdohar: done

PR #531, opened 2018-09-06
Increase stdout buffer size in the Docker executor

jberci:


dblenkus:
What about change log?

tjanez:
> What about change log?

We should include it when making the release.

mstajdohar:
Do we want the buffer size configurable or not? AFAIK this is a temporary fix that will be reverted when the new communication protocol is in place, so I do not see an immediate need to make it configurable (except for other potential users of Resolwe).

PR #532, opened 2018-09-06
Improve Relations v2

dblenkus:


resolwe/flow/models/entity.py
```diff
@@ -133,6 +127,12 @@ class Meta(BaseModel.Meta):
             ("owner_relation", "Is owner of the relation"),
         )
 
+    def save(self, *args, **kwargs):
+        """Lowercase the ``category`` field and save the model."""
+        self.category = self.category.lower()
```
mstajdohar: Why do we want the category lower case? We want to keep the original case, but make uniqueness and search invariant of the case. The reason we want the original case is that the category really acts like name (or label) in the sense that it will be displayed in many places in UI. And if the category is abbreviation of something it should be all caps, maybe they want just the first letter capitalized, maybe the category will be a name of something and all words will have the first letter capitalized... Image the user name a relation (sets the category), and then this relation has a different case in UI. Maybe category will be _Homo sapiens vs Mus musculus study_.  Biologists would be pissed if we mess up the case there.

resolwe/flow/serializers/relation.py
```diff
@@ -47,6 +47,13 @@ class Meta:
             'unit',
         )
 
+    def validate_partitions(self, partitions):
+        """Raise validation error if list of partitions is empty."""
+        if not partitions:
+            raise serializers.ValidationError("List of partitions must not be empty.")
```
mstajdohar: Are we sure we want this?

docs/CHANGELOG.rst
```diff
@@ -25,6 +25,7 @@ Changed
   - adding (integer) ``position`` to partition (used for sorting)
 - **BACKWARD INCOMPATIBLE:** Remove rarely used parameters of the ``register``
   command ``--path`` and ``--schemas``.
+  - deleting ``Relation`` when the last ``Entity`` is removed
```
mstajdohar: Check the indent.

resolwe/flow/serializers/relation.py
dblenkus: Yes, we automatically delete Relation when the last entity is removed, so we don't want to allow to create such "invalid" state.

mstajdohar: Hm, thinking about this more, @hadalin may have an issue with this. The user workflow is such: user clicks a button to create a relation. Then the user selects some sample to create a group on this collection, then another group. User may remove a group from collection. Or change groups if she likes...

So, when the user creates a collection, Primoz will not be able to create it in the backend yet (until the user creates at least one group). So if the user navigates away and back, she will lose her work—the "empty" collection will disappear.

mstajdohar: Maybe we don't want to delete empty collections automatically after all.

mstajdohar: I worry this really complicates the implementation in the web frontend.

hadalin: The workflow is as such: when a user creates a new relation, it is created together with groups (partitions). A user can remove all relations on a collection or a particular relation (a table column), but not groups. So if the only concern is the list of partitions "must not be empty" then I don't think we have a problem.

mstajdohar: > So if the only concern is that the list of partitions "must not be empty", then I don't think we have a problem.

That was my only concern, thanks!

PR #536, opened 2018-09-17
Raise DirtyError if descriptor_completed and descriptor incomplete

mzganec:


mstajdohar:
We will repurpose the `descriptor_completed` field in the near future.

PR #538, opened 2018-09-21
Add tags to collections

dblenkus:


resolwe/flow/models/collection.py
```diff
@@ -30,6 +30,9 @@ class Meta(BaseModel.Meta):
     #: indicate whether `descriptor` doesn't match `descriptor_schema` (is dirty)
     descriptor_dirty = models.BooleanField(default=False)
 
+    #: tags for categorizing objects
+    tags = ArrayField(models.CharField(max_length=255), default=list)
```
lukaw3d: Filtering by tags doesn't work :sob:

mstajdohar: Why not? @dblenkus when can you fix it?

PR #541, opened 2018-10-09
Prepare release 13

mstajdohar:


mstajdohar:
@dblenkus Should we do a major release? I know there's BACKWARDS INCOMPATIBLE flag in the release notes, but the change does not require the update of downstream packages. It merely changes the index definition (the result set).

mstajdohar:
@dblenkus Relations test fails. Please investigate.

mstajdohar:
Weird. I've added a pre-release commit and the bug is gone.

PR #548, opened 2018-10-27
Support multiple entity types

dblenkus:


resolwe/flow/tests/test_executors.py
```diff
@@ -232,6 +232,7 @@ def test_cpu_resource(self):
         self.run_process('test-cpu-resource-1core')
         self.run_process('test-cpu-resource-2core')
 
+    @unittest.skipIf(settings.FLOW_DOCKER_DISABLE_SECCOMP, "Docker seccomp is disabled.")
```
mstajdohar: Why?

resolwe/flow/tests/test_commands.py
```diff
@@ -139,7 +139,7 @@ def setUp(self):
         super().setUp()
 
         # Make sure the test processes are in the database
-        self._register_schemas(path=[PROCESSES_DIR])
+        self._register_schemas(processes_path=[PROCESSES_DIR])
```
mstajdohar: I would write `_paths` in plural since the input is a list of paths. Thus I suggest `process_paths` or `proc_paths`.

docs/proc.rst
```diff
@@ -202,6 +203,24 @@ We have predefined three top categories: upload, import and analyses. Processes
...
+Entity
+------
+
+With defining the ``entity`` field in the process, new data objects will be
+automatically attached to a new or existing Entity, depending on it's parents
+and the definition of the field.
+
+``entity`` field has 3 subfields:
+
+* ``type`` is required and defines the type of entity that created ``Data``
```
mstajdohar: created -> the new

docs/proc.rst
```diff
@@ -202,6 +203,24 @@ We have predefined three top categories: upload, import and analyses. Processes
...
+  object is attached to
+* ``input`` limits the group of parents' enteties to a single field (comma
```
mstajdohar: dot separated path?

resolwe/flow/management/commands/register.py
```diff
@@ -124,6 +124,18 @@ def register_processes(self, process_schemas, user, force=False, verbosity=1):
         log_templates = []
 
         for p in process_schemas:
+            # TODO: Remove this when all processes are migrated to the
+            #       new syntax.
+            if 'flow_collection' in p:
+                if 'entity' in p:
+                    self.stderr.write(
+                        "Skip processor {}: only one of 'flow_collection' and 'entity' fields "
+                        "can be used".format(p['slug'])
```
mstajdohar: can be used -> allowed

resolwe/flow/management/commands/register.py
```diff
@@ -124,6 +124,18 @@ def register_processes(self, process_schemas, user, force=False, verbosity=1):
         log_templates = []
 
         for p in process_schemas:
+            # TODO: Remove this when all processes are migrated to the
+            #       new syntax.
+            if 'flow_collection' in p:
+                if 'entity' in p:
+                    self.stderr.write(
+                        "Skip processor {}: only one of 'flow_collection' and 'entity' fields "
```
mstajdohar: We should rename processor to process (everywhere)

resolwe/flow/management/commands/register.py
```diff
@@ -139,6 +151,25 @@ def register_processes(self, process_schemas, user, force=False, verbosity=1):
             if not self.valid(p, PROCESSOR_SCHEMA):
                 continue
 
+            if 'entity' in p:
+                if 'type' not in p['entity']:
+                    self.stderr.write(
+                        "Skip processor {}: 'entity.type' is required if 'entity' is defined".format(p['slug'])
```
mstajdohar: `Skip process {}: 'entity.type' required if 'entity' defined".format(p['slug']`

resolwe/flow/management/commands/register.py
```diff
@@ -139,6 +151,25 @@ def register_processes(self, process_schemas, user, force=False, verbosity=1):
...
+                    )
+                    continue
+
+                p['entity_type'] = p['entity']['type']
+                p['entity_descriptor_schema'] = p['entity'].get('descriptor_schema', p['entity_type'])
```
mstajdohar: Remove the trailing colon (`:`) of the default value.

resolwe/flow/management/commands/register.py
```diff
@@ -342,8 +373,10 @@ def handle(self, *args, **options):
                 self.find_schemas(desc_path, schema_type=SCHEMA_TYPE_DESCRIPTOR, verbosity=verbosity))
 
         user_admin = users.first()
-        self.register_processes(process_schemas, user_admin, force, verbosity=verbosity)
         self.register_descriptors(descriptor_schemas, user_admin, force, verbosity=verbosity)
+        # NOTE: Descriptor schemas must be registered first, so
+        #       processes can validate 'entity_descriptor_schema' field.
+        self.register_processes(process_schemas, user_admin, force, verbosity=verbosity)
```
mstajdohar: Shouldn't we use logger instead of the verbosity and writing to stderr directly? I know this would require quite some refactoring...

resolwe/flow/models/data.py
```diff
@@ -319,31 +322,58 @@ def create_entity(self):
...
+
+            if entity_input:
+                input_id = dict_dot(self.input, entity_input, default=lambda: None)
+                if input_id is None:
+                    logger.warning("Skipping creation of entity due to missing input.")
+                    return
+                if isinstance(input_id, int):
+                    entity_query = Entity.objects.filter(data__pk=input_id)
+                elif isinstance(input_id, list):
+                    entity_query = Entity.objects.filter(data__pk__in=input_id)
```
mstajdohar: Filter by `type` too. You may want to only define the data filter within `if entity_input:` and then run the query outside.

resolwe/flow/models/data.py
```diff
@@ -319,31 +322,58 @@ def create_entity(self):
...
 
-            # Inherite collections from entity.
-            for collection in entity.collections.all():
-                collection.data.add(self)
+            else:
+                logger.info("Skipping creation of entity due to multiple entities found.")
+                entity = None
+
+            if entity:
+                entity.data.add(self)
```
mstajdohar: Maybe all this (entity creation, attaching data to entity and collections...) should be done in transaction?

resolwe/flow/models/process.py
```diff
@@ -120,20 +120,25 @@ class Meta(BaseModel.Meta):
...
-    data-flow. If all input ``Data`` objects belong to the same
-    collection, add newly created ``Data`` object to it, otherwise
-    create a new collection.
-    If :class:`~resolwe.flow.models.DescriptorSchema` object with
-    ``type`` matching this field exists, reference it in the
-    collection's
-    :attr:`~resolwe.flow.models.collection.BaseCollection.descriptor_schema`
-    field.
+    with this process to an :class:`~resolwe.flow.models.Entity` object
+    representing a data-flow.
```
mstajdohar: Add a blank line or merge with the above line.

resolwe/flow/models/entity.py
```diff
@@ -27,6 +27,9 @@ class Meta(BaseCollection.Meta):
     #: indicate whether `descriptor` is completed (set by user)
     descriptor_completed = models.BooleanField(default=False)
 
+    #: entity type
+    type = models.CharField(max_length=100, null=True, blank=True)
```
mstajdohar: We probably want index on this field. Add it to serializer and support filtering on the endpoint.

resolwe/flow/tests/test_executors.py
dblenkus: Because it doesn't make sense to test seccomp if it is disabled :)

dblenkus:
What should be supported in Python processes?

mstajdohar:
> What should be supported in Python processes?

Defining the entity field on the process.

resolwe/flow/management/commands/register.py
mstajdohar: Ah, the above code confused me a bit:
```
 if p['type'][-1] != ':':
    p['type'] += ':'
```
I see now this is process type...

resolwe/flow/models/data.py
```diff
@@ -319,31 +322,55 @@ def create_entity(self):
...
+                input_id = dict_dot(self.input, entity_input, default=lambda: None)
+                if input_id is None:
+                    logger.warning("Skipping creation of entity due to missing input.")
+                    return
+                if isinstance(input_id, int):
+                    data_filter['data__pk'] = input_id
+                elif isinstance(input_id, list):
+                    data_filter['data__pk__in'] = input_id
+                else:
+                    raise ValueError("")
```
mstajdohar: You could also print out something like Skipping creation of entity due to invalid value of field {}. We should check the field in register that it is valid under process schema and that the field type is `data:*` or `list:data:*`. Bioinformaticians will make typos.

resolwe/flow/models/data.py
mstajdohar: Or is the whole `save()` method called in transaction? Could you write a test?

resolwe/flow/models/entity.py
```diff
@@ -27,6 +27,9 @@ class Meta(BaseCollection.Meta):
     #: indicate whether `descriptor` is completed (set by user)
     descriptor_completed = models.BooleanField(default=False)
 
+    #: entity type
+    type = models.CharField(max_length=100, db_index=True, null=True, blank=True)
```
mstajdohar: Do we really want to support blank type?

resolwe/flow/serializers/entity.py
```diff
@@ -18,6 +18,7 @@ class Meta(CollectionSerializer.Meta):
         fields = CollectionSerializer.Meta.fields + (
             'collections',
             'descriptor_completed',
+            'type',
```
mstajdohar: Don't forget about the filter. We need to be able to query for entities of given type.

mstajdohar:
Approved. Rebase and merge, please.

PR #550, opened 2018-11-05
Fix manager shutdown in the test runner

jberci:
If an unrecoverable exception occured while running a test, and never
got caught (e.g. an unpicklable exception in a parallel test worker),
the listener would not get terminated properly, leading to a hang.


mstajdohar:
huh, ugly. Thanks for the fix!

PR #553, opened 2018-11-08
Support extending viewsets with custom filter methods

dblenkus:


resolwe/elastic/viewsets.py
```diff
@@ -71,6 +73,12 @@ def __init__(self, *args, **kwargs):
             self.ordering_fields = self.ordering_fields + tuple(ordering_fields)
             self.ordering_map.update(ordering_map)
 
+            for field_name in filtering_fields:
+                custom_filter_name = 'custom_filter_' + field_name
```
kostko: Use format.

PR #554, opened 2018-11-14
Use raw filtering for collection and data name

JureZmrzlikar:


kostko:
But `__exact` is exactly how it should already work?

kostko:
I mean I remember implementing the `exact` lookup a while back:
https://github.com/genialis/resolwe/blob/0ca8d6bc8f0bfc114cdf3d3e7532e22601c5297f/resolwe/elastic/tests/test_viewsets.py#L355-L371

kostko:
Oh, but it currently only works for direct field names (e.g., `name`), not subfields (e.g., `name.ngrams`). Should be easy to change though.

kostko:
In any case, if having exact matches by default is the desired behavior this PR seems ok to me.

dblenkus:
I think this filter is not used anywhere. So we have 3 options how it can work: with ngram analyzer, with keyword analyzer or with no analyzer.

Depending on how other similar filters works, we can also support `name_exact` filter to keep the behaviour in sync.

@kostko does this mean to only add `field = field.rstrip('.ngrams')` [here](https://github.com/genialis/resolwe/blob/0ca8d6bc8f0bfc114cdf3d3e7532e22601c5297f/resolwe/elastic/lookup.py#L101)?

kostko:
> @kostko does this mean to only add `field = field.rstrip('.ngrams')` here?

Something more general, but yeah. We should likely split by dots and extract the first component instead of hardcoding `.ngrams`.

JureZmrzlikar:
@dblenkus , @mstajdohar ping, I would like to merge this ASAP.

PR #555, opened 2018-11-15
Switch channels_redis dependency to usptream version

dblenkus:


setup.py
```diff
@@ -59,7 +59,7 @@
         'django-filter~=1.0.0',
         'redis~=2.10.6',
         'channels~=2.1',
-        'channels_redis_persist~=2.2',
+        'channels_redis~=2.3',
```
tjanez: Please, fix the typo in the commit message:

usptream -> upstream

PR #559, opened 2018-11-29
Kill Docker executor if container doesn't start for 30 seconds

dblenkus:


resolwe/flow/executors/docker/run.py
```diff
@@ -218,13 +221,27 @@ def new_volume(kind, base_dir_name, volume, path=None, read_only=True):
...
+        async def wait_for_container():
+            """Wait for Docker container to start to avoid blocking the code that uses it."""
+            self.proc.stdin.write(('echo PING' + os.linesep).encode('utf-8'))
+            await self.proc.stdin.drain()
+            while True:
+                line = await self.proc.stdout.readline()
+                stdout.append(line)
+                if line.rstrip() == b'PING':
+                    break
+                await asyncio.sleep(0.01)
```
jberci: Why the sleep here? `readline()` won't return until it gets something anyway.

dblenkus: It looks like it doesn't block.

dblenkus: Ok, it makes sense that it doesn't block if stdout is already closed. But this means that something strange has happened. I've handled this case.

PR #564, opened 2018-12-10
Mini fixes

dblenkus:


resolwe/flow/managers/listener.py
```diff
@@ -511,6 +514,8 @@ def handle_finish(self, obj):
             },
         })
 
+        batcher.commit()
```
mstajdohar: Consider enclosing batcher with `try/finally` or using a `with` statement.

PR #573, opened 2018-12-17
Fix mock_file_upload to handle nested paths.

JureZmrzlikar:
@dblenkus , Is this the solution that you had in mind?
@romunov Please verify that this solves the issue. I've made a test that replicates the problem, but there might be more missing pieces.


resolwe/toolkit/tests/files/upload_file/input_file.txt
```diff
@@ -0,0 +1,2 @@
+Foo
```
dblenkus: You can use the same file as the input and output.

PR #575, opened 2018-12-17
Make sure that tmp dir exists in Docker executor

dblenkus:


resolwe/flow/executors/docker/run.py
```diff
@@ -200,6 +201,10 @@ def new_volume(kind, base_dir_name, volume, path=None, read_only=True):
         # Add any extra volumes verbatim.
         volumes += SETTINGS.get('FLOW_DOCKER_EXTRA_VOLUMES', [])
 
+        # Make sure that tmp dir exists.
+        with suppress(OSError):
```
mstajdohar: Why suppress?

resolwe/flow/executors/docker/run.py
```diff
@@ -200,6 +201,10 @@ def new_volume(kind, base_dir_name, volume, path=None, read_only=True):
...
+            os.mkdir(constants.TMPDIR)
```
mstajdohar: I suggest to use `os.makedirs` in case `TMPDIR` would be changed to include multiple subfolders sometime in the future.

mstajdohar: Should we set the same `mode` as for data dir [here](https://github.com/genialis/resolwe/blob/58120be940bf8b6092763ed159da233a8e075adf/resolwe/flow/managers/dispatcher.py#L421)?

PR #576, opened 2018-12-18
Skip docker image pull in executor if set via envar

JureZmrzlikar:


dblenkus:
What is the use-case for this?

Next command (`docker run`) requires the image anyway and will have to pull it if it is not already present. This chunk of code just makes sure that the executor doesn't timeout due to the pulling overhead. So there is (almost) no performance benefit if we disable it.

And even if we add this, we should probably use a different setting, as `FLOW_DOCKER_DONT_PULL` has a different use-case that conflicts with this one.

JureZmrzlikar:
Sorry for not providing context... Use case is: developer is working with yet un-published version of docker image. Because of this, docker pull will always fail. This means that developer has to skip(comment) this part of  code for tests to pass. Do you have any alternative solution in mind?

We could just check the existence of the image and raise error if it is missing. And maybe raise warninfg if it cant be pulled?

mstajdohar:
I leave you to discuss the best solution. I only ask to include the context in the commit message and change log.

dblenkus:
The image has to be on the system after this point, you cannot defer the pull any longer. And you have to pull it if it is missing as `FLOW_DOCKER_DONT_PULL` (and other things) relies on this.

I think that the best (and the simplest) solution would be to pull the image only if it doesn't exist locally, e.g. something like `docker image inspect <image_name> >/dev/null 2>&1 && echo yes || echo no`

resolwe/flow/executors/docker/run.py
```diff
@@ -216,22 +216,35 @@ def new_volume(kind, base_dir_name, volume, path=None, read_only=True):
         # A non-login Bash shell should be used here (a subshell will be spawned later).
         command_args['shell'] = '/bin/bash'
 
-        pull_command = '{command} pull {container_image}'.format(**command_args)
+        # Check if image exists locally. If not, command will exit with non-zero returncode
+        check_command = '{command} image inspect {container_image}'.format(**command_args)
 
-        logger.info("Pulling docker image: {}".format(command_args['container_image']))
+        logger.info("Checking existence of docker image: {}".format(command_args['container_image']))
```
dblenkus: This should be a debug log as the following command is local and should execute instantly.

docs/CHANGELOG.rst
```diff
@@ -13,6 +13,9 @@ This project adheres to `Semantic Versioning <http://semver.org/>`_.
 Fixed
 -----
 - Make sure that tmp dir exists in Docker executor
+- Skip docker image pull if image exists locally. This solves the issue
+  where pull would fail if process uses an image that is only used locally
+  for testing.
```
dblenkus: No dot at the end. You can also omit "for testing".

PR #577, opened 2018-12-18
Add assertFilesExist method to ProcessTestCase

JureZmrzlikar:


resolwe/test/testcases/process.py
```diff
@@ -710,6 +710,31 @@ def assertFileExists(self, obj, field_path):  # pylint: disable=invalid-name
...
+        :param str filenames: Expected names of the files.
+        """
+        field = dict_dot(obj.output, field_path)
+
+        output_files = []
+        for item in field:
+            output_file = os.path.join(settings.FLOW_EXECUTOR['DATA_DIR'], str(obj.pk), item['file'])
+            output_files.append(output_file)
+            if not os.path.isfile(output_file):
+                self.fail(msg="File(s) in output field {} don't exist.".format(field_path))
```
dblenkus: This will fail for the first missing file, so there is no need for `(s)`.

resolwe/test/testcases/process.py
```diff
@@ -710,6 +710,31 @@ def assertFileExists(self, obj, field_path):  # pylint: disable=invalid-name
...
+
+        if filenames:
```
dblenkus: This is an unnecessary thing as in our model it only matters that output field references a file and that the file exists on the disk, the naming doesn't matter. 

But if you want to keep it, make sure that `assertFilesExist` and `assertFileExists` are in sync and both support the same options.

docs/CHANGELOG.rst
```diff
@@ -34,6 +34,7 @@ Added
   seconds
 - Set ``TMPDIR`` environment variable in Docker executor to ``.tmp`` dir in
   data directory to prevent filling up container's local storage
+- Add ``assertFilesExist`` method to ``ProcessTestCase``
```
dblenkus: This should go under unreleased.

resolwe/test/testcases/process.py
JureZmrzlikar: This comes really handy in the following case: You expect multiple files to be created, but a different set of them, depending on some input. That way, you can check that correct files have been made, but do not need to check their sha256

resolwe/test/testcases/process.py
```diff
@@ -710,6 +710,34 @@ def assertFileExists(self, obj, field_path):  # pylint: disable=invalid-name
...
+        :param str filenames: Expected names of the files.
+        """
+        field = dict_dot(obj.output, field_path)
+
+        output_files = []
+        for item in field:
+            output_file = os.path.join(settings.FLOW_EXECUTOR['DATA_DIR'], str(obj.pk), item['file'])
+            output_files.append(output_file)
+            if not os.path.isfile(output_file):
+                self.fail(msg="File in output field {} does not exist.".format(field_path))
```
dblenkus: It would be really handy to also include the missing filename.

resolwe/test/testcases/process.py
dblenkus: I agree. But my point was that platform would (or may) work even if filenames change, but this will fail in such cases. So I am not in favour of exposing such anti-patterns, as they give the feeling that this is intended and can cause a lot of problems when get widely used.

resolwe/test/testcases/process.py
```diff
@@ -693,7 +693,7 @@ def assertFiles(self, obj, field_path, fn_list, **kwargs):  # pylint: disable=in
         for fn_tested, fn_correct in zip(field, fn_list):
             self._assert_file(obj, fn_tested['file'], fn_correct, **kwargs)
 
-    def assertFileExists(self, obj, field_path):  # pylint: disable=invalid-name
+    def assertFileExists(self, obj, field_path, filename=None):  # pylint: disable=invalid-name
```
dblenkus: Don't forget to mention this in change log.

JureZmrzlikar:
@dblenkus I've decided to leave out the `filename(s)` parameter. Please have another review pass when you find some time.

PR #578, opened 2018-12-27
Duplicate

hadalin:
Requires https://github.com/genialis/resolwe/pull/571


resolwe/flow/models/entity.py
```diff
@@ -30,6 +58,59 @@ class Meta(BaseCollection.Meta):
...
+        duplicate.pk = None
+        duplicate.slug = None
+        name_max_length = Entity._meta.get_field('name').max_length
+        duplicate.name = 'Copy of {}'.format(self.name)[:name_max_length]
+        duplicate.contributor = contributor
+        duplicate.duplicated = datetime.datetime.now()
+        duplicate.save(force_insert=True)  # pylint: disable=protected-access
+
+        # TODO
+        # Create duplication data migration history entry.
```
dblenkus: This is not needed here.

resolwe/flow/models/entity.py
```diff
@@ -30,6 +58,59 @@ class Meta(BaseCollection.Meta):
...
+
+        assign_contributor_permissions(duplicate)
+
+        # Fields to inherit from original entity.
```
dblenkus: Commend from the `Data.duplicate` also applies here.

resolwe/flow/models/entity.py
```diff
@@ -30,6 +58,59 @@ class Meta(BaseCollection.Meta):
...
+        duplicate.created = self.created
+        duplicate.save()
+
+        # Duplicate entity's data objects.
+        # pylint: disable=no-member
```
dblenkus: Don't disable this for the whole function.

resolwe/flow/utils/__init__.py
```diff
@@ -104,3 +104,40 @@ def get_apps_tools():
...
+    """Rewire inputs of provided data objects.
+
+    Input parameter is a list of original and derivative data object
+    model instances: ``[{'original': original, 'derivative': derivative}]``.
+    This function finds which objects reference other objects
+    (in the list) on the input and replaces original objects with
+    the derivatives (mutates derivatives' inputs).
+
+    """
+    for data in data_list:
```
dblenkus: `data` is not the best name, as it implies that this is a Data object.

resolwe/flow/models/entity.py
```diff
@@ -30,6 +58,59 @@ class Meta(BaseCollection.Meta):
...
     type = models.CharField(max_length=100, db_index=True, null=True, blank=True)
 
+    #: duplication date and time
+    duplicated = models.DateTimeField(blank=True, null=True)
+
+    def is_duplicate(self):
+        """Return True if entity is a duplicate."""
+        return bool(self.duplicated)
+
+    def _duplicate(self, contributor, inherit_collection_membership):
```
dblenkus: Set a default value for `inherit_collection_membership`.

resolwe/flow/models/entity.py
```diff
@@ -30,6 +58,59 @@ class Meta(BaseCollection.Meta):
...
+
+        assign_contributor_permissions(duplicate)
+
+        # Fields to inherit from original entity.
+        duplicate.created = self.created
+        duplicate.save()
+
+        # Duplicate entity's data objects.
+        # pylint: disable=no-member
+        orig_and_duplicated_data = self.data.all().duplicate(contributor, return_mapping=True)
```
dblenkus: `orig_and_duplicated_data` -> `copied_data`

resolwe/flow/models/entity.py
```diff
@@ -30,6 +58,59 @@ class Meta(BaseCollection.Meta):
...
+        if inherit_collection_membership:
+            collections = get_objects_for_user(
+                contributor,
+                'add_collection',
+                self.collections.all()  # pylint: disable=no-member
+            )
+            for collection in collections:
+                collection.entity_set.add(duplicate)
+
+                # Add duplicated data to collection only if the original
```
dblenkus: If an entity in the collection it implies that also all data from the entity should also be in the collection, so you can just add it.

resolwe/flow/models/entity.py
dblenkus: And considering the comment bellow you can greatly simplify it with omitting `return_mapping=True`.

resolwe/flow/views/entity.py
```diff
@@ -86,6 +89,9 @@ def add_to_collection(self, request, pk=None):
         """Add Entity to a collection."""
         entity = self.get_object()
 
+        # TODO use `self.get_ids` (and elsewhere).
+        # Backwards incompatible because raised error's
```
dblenkus: 72

resolwe/flow/views/entity.py
```diff
@@ -156,3 +162,23 @@ def patched_get_queryset():
         resp = super().update(request, *args, **kwargs)
         self.get_queryset = orig_get_queryset
         return resp
+
+    @list_route(methods=['post'])
+    def duplicate(self, request, *args, **kwargs):
+        """Duplicate (make copy of) ``Entity`` models."""
+        inherit_collection_membership = request.data.get('inherit_collection_membership', False)
```
dblenkus: Why not just `inherit_collections`?

resolwe/flow/views/entity.py
```diff
@@ -156,3 +162,23 @@ def patched_get_queryset():
...
+        ids = self.get_ids('ids', request.data)
+        queryset = get_objects_for_user(request.user, 'view_entity', Entity.objects.filter(id__in=ids))
+        actual_ids = queryset.values_list('id', flat=True)
+        missing_ids = list(set(ids) - set(actual_ids))
+        if missing_ids:
+            raise exceptions.ParseError("Entities with the following ids not found: {}"
```
dblenkus:
```
            raise exceptions.ParseError(
                "Entities with the following ids not found: {}".format(', '.join(map(str, missing_ids)))
            )
````

File: resolwe/flow/models/entity.py [hunk 11]
```diff
@@ -30,6 +30,20 @@ def duplicate(self, contributor, inherit_collection_membership=False):
...
         ]
 
+    @transaction.atomic
+    def move_to_collection(self, source_collection, destination_collection):
+        """Move entities from source to destination collection."""
+        for entity in self:
+            # Remove from collection.
+            entity.collections.remove(source_collection)
+            for data in entity.data.all():
+                source_collection.data.remove(data)
```
dblenkus: Can you just use `source_collection.data.remove(entity.data.all())`?

resolwe/flow/views/entity.py
```diff
@@ -48,6 +48,19 @@ def _check_collection_permissions(self, collection_id, user):
             else:
                 raise exceptions.NotFound()
 
+        return collection
```
dblenkus: This is a bad idea to return the collection. If you want to do this, you have to rename the function to `_get_collection_for_user` (in a separate commit, of course).

resolwe/flow/views/entity.py
```diff
@@ -48,6 +48,19 @@ def _check_collection_permissions(self, collection_id, user):
...
+
+    def _get_entities(self, user, ids):
+        """Return entities queryset based on provided entity ids."""
+        queryset = get_objects_for_user(user, 'view_entity', Entity.objects.filter(id__in=ids))
+        actual_ids = queryset.values_list('id', flat=True)
+        missing_ids = list(set(ids) - set(actual_ids))
+        if missing_ids:
+            raise exceptions.ParseError("Entities with the following ids not found: {}"
```
dblenkus: New line after opening parenthesis.

resolwe/flow/views/entity.py
```diff
@@ -138,6 +151,21 @@ def add_data(self, request, pk=None):
 
         return resp
 
+    @list_route(methods=['post'])
+    def move_to_collection(self, request, *args, **kwargs):
+        """Move samples from source to destination collection."""
+        entity_ids = self.get_ids('entity_ids', request.data)
```
dblenkus: `entity_ids` -> `ids` as this is on the entity endpoint.

resolwe/flow/models/entity.py
```diff
@@ -30,6 +30,20 @@ def duplicate(self, contributor, inherit_collection_membership=False):
             for entity in self
         ]
 
+    @transaction.atomic
+    def move_to_collection(self, source_collection, destination_collection):
```
dblenkus: Move the logic in to the model, so you can also move a single object. You can keep a wrapper here, to use it on the queryset.

resolwe/flow/views/entity.py
```diff
@@ -138,6 +151,21 @@ def add_data(self, request, pk=None):
...
+        source_collection_id = self.get_id('source_collection', request.data)
+        destination_collection_id = self.get_id('destination_collection', request.data)
+
+        source_collection = self._check_collection_permissions(source_collection_id, request.user)
+        destination_collection = self._check_collection_permissions(destination_collection_id, request.user)
+
+        entities_queryset = self._get_entities(request.user, entity_ids)
```
dblenkus: `entities_queryset` -> `entity_qs`

resolwe/flow/views/entity.py
```diff
@@ -138,6 +151,21 @@ def add_data(self, request, pk=None):
 
         return resp
 
+    @list_route(methods=['post'])
+    def move_to_collection(self, request, *args, **kwargs):
+        """Move samples from source to destination collection."""
+        entity_ids = self.get_ids('entity_ids', request.data)
+        source_collection_id = self.get_id('source_collection', request.data)
```
dblenkus: `src_collection_id`, `dst_collection_id`, `src_collection`, `dst_collection`. For variables, you can keep the long form in the request.

resolwe/flow/views/mixins.py
```diff
@@ -156,3 +156,14 @@ def get_ids(self, parameter_name, request_data):
             raise ParseError("`{}` parameter contains non-integers".format(parameter_name))
 
         return ids
+
+    def get_id(self, parameter_name, request_data):
```
dblenkus: `    def get_id(self, request_data, parameter_name='id'):`

resolwe/flow/views/data.py
```diff
@@ -137,6 +138,22 @@ def create(self, request, *args, **kwargs):
...
+        """Duplicate (make copy of) ``Data`` objects."""
+        ids = self.get_ids('ids', request.data)
+        queryset = get_objects_for_user(request.user, 'view_data', Data.objects.filter(id__in=ids))
+        actual_ids = queryset.values_list('id', flat=True)
+        missing_ids = list(set(ids) - set(actual_ids))
+        if missing_ids:
+            raise exceptions.ParseError("Data objects with the following ids not found: {}"
+                                        .format(', '.join(map(str, missing_ids))))
+
+        duplicated = queryset.duplicate(contributor=self.resolve_user(request.user))
```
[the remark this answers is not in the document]
hadalin: The contributor should be the one creating the duplicate.

----

>so you can just set it to `request.user`

Isn't think endpoint available to anonymous user too?

resolwe/flow/models/data.py
```diff
@@ -447,6 +476,56 @@ def delete(self, *args, **kwargs):
...
+        duplicate.pk = None
+        duplicate.slug = None
+        name_max_length = Data._meta.get_field('name').max_length
+        duplicate.name = 'Copy of {}'.format(self.name)[:name_max_length]
+        duplicate.contributor = contributor
+        duplicate.duplicated = datetime.datetime.now()
+        duplicate._perform_save(force_insert=True)  # pylint: disable=protected-access
+
+        # TODO
+        # Create duplication data migration history entry.
```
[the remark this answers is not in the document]
hadalin: This one, but please give directions how to do this.

resolwe/flow/models/migrations.py
```diff
@@ -31,10 +28,10 @@ def __str__(self):
 class ProcessMigrationHistory(MigrationHistoryBase):
     """Model for storing process migration history."""
 
-    process = models.ForeignKey(Process, related_name='migration_history', on_delete=models.CASCADE)
+    process = models.ForeignKey('Process', related_name='migration_history', on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
hadalin: No need.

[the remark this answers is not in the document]
hadalin: No, it _is_ needed, because of circular imports.

resolwe/flow/tests/test_models.py
```diff
@@ -521,6 +524,244 @@ def test_raw_process(self):
         self.assertNotEqual(response.data['id'], self.data.pk)
 
 
+class DuplicateTestCase(APITestCase):
```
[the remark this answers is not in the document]
hadalin: Good point, not needed. Probably a copy/paste leftover.

resolwe/flow/tests/test_models.py
```diff
@@ -521,6 +524,244 @@ def test_raw_process(self):
...
+
+    def setUp(self):
+        super().setUp()
+
+        user_model = get_user_model()
+        self.user = user_model.objects.create(username='test_user', password='test_pwd')
+        self.duplication_user = user_model.objects.create(username='duplication_user', password='duplication_pwd')
+
+    def test_data_duplicate(self):
```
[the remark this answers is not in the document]
hadalin: There's a lot of properties to test when duplicating. Splitting the test into smaller parts would mean duplicating the objects in `setUp` and then testing different properties in each test. Is that what you meant?

resolwe/flow/tests/test_models.py
```diff
@@ -521,6 +524,244 @@ def test_raw_process(self):
...
+        data_location.save()
+        data_location.data.add(data3)
+
+        # Duplicate.
+        duplicates = Data.objects.filter(id__in=[data1.id, data2.id, data3.id]).duplicate(self.duplication_user)
+        [duplicate1, duplicate2, duplicate3] = duplicates
+
+        self.assertEqual(duplicate2.input['data_field'], duplicate1.id)
+        self.assertEqual(duplicate2.input['data_list_field'], [duplicate1.id])
+        self.assertEqual(duplicate3.input['data_field'], duplicate2.id)
```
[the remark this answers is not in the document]
hadalin: No. Changed the test a bit.

resolwe/flow/views/entity.py
hadalin: Because it's not explicit enough. Current form rings a bell sooner.

resolwe/flow/models/entity.py
hadalin: `source_collection.data.remove(*entity.data.all())` will do yes.

resolwe/flow/models/data.py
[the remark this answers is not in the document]
dblenkus: You just duplicate all [`DataMigrationHistory`](https://github.com/genialis/resolwe/blob/7ee372a9700a4f3fedd621cf551567354741d9aa/resolwe/flow/models/migrations.py#L37) entries.

resolwe/flow/views/data.py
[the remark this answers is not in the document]
dblenkus: No, it should only be available to logged-in users.

resolwe/flow/models/data.py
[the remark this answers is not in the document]
hadalin: This was meant to log the relation between parent and child object. Before there was a new relation type `DataDependency`, but we discarded the idea. The question is how to log original -> duplicate.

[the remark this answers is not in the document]
hadalin: But since we're here, I'll also duplicate `DataMigrationHistory`.

resolwe/flow/tests/test_models.py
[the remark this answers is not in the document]
dblenkus: Yes, sometimes it can be quite hard to read a test if it is more than 100 lines long.

Other thing is that you should only test reasonable attributes, e.g.
1. It doesn't make sense to test the name of a copy of a copy as it is only copied and truncated
2. It does make sense to test relations of copy of a copy or even relations of a 2nd copy of the original object

resolwe/flow/models/data.py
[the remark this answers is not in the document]
hadalin: >But since we're here, I'll also duplicate DataMigrationHistory

This is already done [apparently](https://github.com/genialis/resolwe/pull/578/files#diff-aa73ad8a382e284ba1af2170c5e136caR524).

resolwe/flow/models/data.py
```diff
@@ -457,6 +487,59 @@ def delete(self, *args, **kwargs):
...
+        # Override fields that are automatically set on create.
+        duplicate.created = self.created
+        duplicate._perform_save()  # pylint: disable=protected-access
+
+        if self.location:
+            self.location.data.add(duplicate)  # pylint: disable=no-member
+
+        duplicate.storages.set(self.storages.all())  # pylint: disable=no-member
+
+        migration_queryset = DataMigrationHistory.objects.filter(data=self).order_by('created')
```
hadalin: Change the query from
`DataMigrationHistory.objects.filter(data=self).order_by('created')`
to
`self.migration_history.all()`?

resolwe/flow/models/data.py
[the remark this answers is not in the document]
hadalin: Regarding the `DataMigrationHistory` entry.. what about something like

```py
DataMigrationHistory.objects.create(
    migration='identifier_please_suggest',
    data=duplicate,
    metadata={'parent_id': self.id},
)
```

resolwe/flow/tests/test_models.py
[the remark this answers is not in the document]
hadalin: I don't know, tried breaking it up, but it's even messier. Creating data object in `setUp` seems confusing, because each test case needs to create its own specific objects. Sure, it's a long test, but at least everything's contained and separated by comments.

Creating another `TestCase` just for this and then having another `TestCase` to test other things (`test_input_rewiring`, `test_data_status_not_done`,..) seems like a unnecessary overhead.

resolwe/flow/models/data.py
```diff
@@ -69,6 +76,22 @@ def delete_chunked(self, chunk_size=500):
         """
         return DataQuerySet._delete_chunked(self, chunk_size=chunk_size)
 
+    @transaction.atomic
+    def duplicate(self, contributor):
```
dblenkus: `contributor=None`

resolwe/flow/tests/test_models.py
```diff
@@ -521,6 +524,197 @@ def test_raw_process(self):
...
+            contributor=self.user,
+            input_schema=[
+                {'name': 'data_field', 'type': 'data:test:first:'},
+            ],
+            output_schema=[
+                {'name': 'json_field', 'type': 'basic:json:'},
+            ],
+        )
+
+        data = Data.objects.create(
```
dblenkus: Name this one `child_data` or `input_data` and the other one just `data` to make it cleaner.

resolwe/flow/tests/test_models.py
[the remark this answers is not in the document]
dblenkus: I don't see any benefit of checking every attribute of `duplicate_of_duplicate` in this test. In this case, I would prefer to split this test in two:

- create a minimal example of duplication and assert all attributes
- focus only on dependencies in the second and check what happens if you copy already copied object, what happens if you copy a source object twice,...
- you can add other tests for more complex parts of the process if there are any

I agree it is not the best thing to add a separate test case for each test, but maybe it would be beneficial to add a separate test case for all duplication tests.

resolwe/flow/tests/test_models.py
```diff
@@ -521,6 +524,197 @@ def test_raw_process(self):
...
+        duplicate = duplicates[0]
+        self.assertTrue(duplicate.is_duplicate())
+
+        duplicates = Data.objects.filter(id=duplicate.id).duplicate(self.contributor)
+        self.assertEqual(len(duplicates), 1)
+        duplicate_of_duplicate = duplicates[0]
+        self.assertTrue(duplicate_of_duplicate.is_duplicate())
+
+        # Convert original and duplicated to dict.
+        data2_dict = Data.objects.filter(id=data2.id).values()[0]
```
dblenkus: You can use:
```
from django.forms.models import model_to_dict
data2_dict = model_to_dict(data2)
```

resolwe/flow/utils/__init__.py
```diff
@@ -104,3 +104,40 @@ def get_apps_tools():
...
+
+    for bundle in data_list:
+        updated = False
+        copy = bundle['copy']
+
+        for field_schema, fields in iterate_fields(copy.input, copy.process.input_schema):
+            name = field_schema['name']
+            value = fields[name]
+
+            if field_schema.get('type', '').startswith('data:') and value in mapped_ids:
```
dblenkus: Type is a required key, so you can just use `field_schema['type']`.

resolwe/flow/utils/__init__.py
```diff
@@ -104,3 +104,40 @@ def get_apps_tools():
...
+                fields[name] = mapped_ids[value]
+                updated = True
+
+            elif field_schema.get('type', '').startswith('list:data:') and any([id in mapped_ids for id in value]):
```
dblenkus: `id` is a build-in function, use `id_` instead.

resolwe/flow/models/entity.py
```diff
@@ -30,6 +59,46 @@ class Meta(BaseCollection.Meta):
...
+        return bool(self.duplicated)
+
+    def duplicate(self, contributor, inherit_collection_membership):
+        """Duplicate (make a copy)."""
+        duplicate = Entity.objects.get(id=self.id)
+        duplicate.pk = None
+        duplicate.slug = None
+        name_max_length = Entity._meta.get_field('name').max_length
+        duplicate.name = 'Copy of {}'.format(self.name)[:name_max_length]
+        duplicate.contributor = contributor
```
dblenkus: Support a duplication without a given contributor.

resolwe/flow/views/entity.py
dblenkus: `inherit_collection_membership ` is too long for my taste, try to find something shorter, e.g. `inherit_collections`, `keep_collections`,...

resolwe/flow/models/collection.py
```diff
@@ -61,3 +77,44 @@ class Meta(BaseCollection.Meta):
...
+        duplicate.pk = None
+        duplicate.slug = None
+        name_max_length = Collection._meta.get_field('name').max_length
+        duplicate.name = 'Copy of {}'.format(self.name)[:name_max_length]
+        duplicate.contributor = contributor
+        duplicate.duplicated = datetime.datetime.now()
+        duplicate.save(force_insert=True)  # pylint: disable=protected-access
+
+        # TODO
+        # Create duplication data migration history entry.
```
dblenkus: Remove this.

resolwe/flow/models/collection.py
```diff
@@ -61,3 +77,44 @@ class Meta(BaseCollection.Meta):
...
+        return bool(self.duplicated)
+
+    def _duplicate(self, contributor):
+        """Duplicate (make a copy)."""
+        duplicate = Collection.objects.get(id=self.id)
+        duplicate.pk = None
+        duplicate.slug = None
+        name_max_length = Collection._meta.get_field('name').max_length
+        duplicate.name = 'Copy of {}'.format(self.name)[:name_max_length]
+        duplicate.contributor = contributor
```
dblenkus: Also here.

resolwe/flow/models/collection.py
```diff
@@ -61,3 +77,44 @@ class Meta(BaseCollection.Meta):
...
+        duplicate.created = self.created
+        duplicate.save()
+
+        # Duplicate collection's entities.
+        duplicated_entities = self.entity_set.all().duplicate(contributor=contributor)  # pylint: disable=no-member
+        duplicate.entity_set.add(*duplicated_entities)
+
+        # Add duplicated data objects to collection.
+        for duplicated_entity in duplicate.entity_set.all():  # pylint: disable=no-member
+            for data in duplicated_entity.data.all():
```
dblenkus: Can you do this in a single query?

resolwe/flow/views/collection.py
```diff
@@ -174,3 +176,22 @@ def remove_data(self, request, pk=None):
...
+        """Duplicate (make copy of) ``Collection`` models."""
+        if not request.user.is_authenticated:
+            raise exceptions.NotFound
+
+        ids = self.get_ids(request.data)
+        queryset = get_objects_for_user(request.user, 'view_collection', Collection.objects.filter(id__in=ids))
+        actual_ids = queryset.values_list('id', flat=True)
+        missing_ids = list(set(ids) - set(actual_ids))
+        if missing_ids:
+            raise exceptions.ParseError("Collections with the following ids not found: {}"
```
dblenkus: New line after opening bracket.

resolwe/flow/views/entity.py
dblenkus: This still needs to be fixed.

resolwe/flow/views/entity.py
```diff
@@ -138,6 +152,21 @@ def add_data(self, request, pk=None):
 
         return resp
 
+    @list_route(methods=['post'])
+    def move_to_collection(self, request, *args, **kwargs):
+        """Move samples from source to destination collection."""
+        ids = self.get_ids(request.data)
+        src_collection_id = self.get_id(request.data, 'source_collection', )
```
dblenkus: No need for the coma at the end.

resolwe/flow/views/entity.py
dblenkus: I see this was fixed in the next commit. Switch commits, to make it nicer and move this line to the other commit.

resolwe/flow/tests/test_models.py
hadalin: I tried that but there's no `contributor_id` if `model_to_dict` is used.

resolwe/flow/models/entity.py
```diff
@@ -30,6 +59,48 @@ class Meta(BaseCollection.Meta):
...
+
+        duplicate.save(force_insert=True)
+
+        assign_contributor_permissions(duplicate)
+
+        # Override fields that are automatically set on create.
+        duplicate.created = self.created
+        duplicate.save()
+
+        # Duplicate entity's data objects.
```
dblenkus: Check permissions here.

resolwe/flow/models/collection.py
```diff
@@ -61,3 +77,42 @@ class Meta(BaseCollection.Meta):
...
+
+        duplicate.save(force_insert=True)  # pylint: disable=protected-access
+
+        assign_contributor_permissions(duplicate)
+
+        # Fields to inherit from original data object.
+        duplicate.created = self.created
+        duplicate.save()
+
+        # Duplicate collection's entities.
```
dblenkus: Check permissions.

docs/CHANGELOG.rst
```diff
@@ -24,6 +24,7 @@ Added
 - Add ``entity_names`` field to ``DataSerializer``
 - Support duplication of ``Data``
 - Support duplication of ``Entity``
+- Support duplication of ``Collection``
```
dblenkus: Combine the three lines.

PR #581, opened 2019-01-21
Bump DRF, Django-filters and DRF filters version

JureZmrzlikar:


resolwe/permissions/mixins.py
```diff
@@ -101,7 +101,7 @@ def detail_permissions(self, request, pk=None):
 
         return Response(get_object_perms(obj))
 
-    @list_route(methods=['get', 'post'], url_path='permissions')
+    @action(detail=False, methods=['get', 'post'], url_path='permissions')
```
dblenkus: Also set `url_name` here.

setup.py
```diff
@@ -36,8 +36,9 @@
     python_requires='>=3.6, <3.7',
     install_requires=[
         'Django~=1.11.0',
-        'djangorestframework~=3.7.0',
-        'djangorestframework-filters~=0.10.0',
+        'djangorestframework~=3.9.0',
+        'djangorestframework-filters~=1.0.0.dev0',
```
dblenkus: We cannot depend on `dev` version. Are there any plans for a next release?

resolwe/flow/filters.py
```diff
@@ -14,12 +14,12 @@
 class BaseResolweFilter(filters.FilterSet):
     """Base filter for Resolwe's endpoints."""
 
-    id = filters.AllLookupsFilter()  # pylint: disable=invalid-name
-    slug = filters.AllLookupsFilter()
-    name = filters.AllLookupsFilter()
+    id = filters.AutoFilter(lookups='__all__')  # pylint: disable=invalid-name
```
dblenkus: It would be great if you can explicitly list lookups. I know it is quite some work, but it will be a lot easier in the future, if we know which use-cases should be supported (e.g. in Elasticsearch), what should be tested,...

setup.py
dblenkus: There were always conflicts between drf, drf-filters and django_filters, but there was no other way to achieve what we needed in the past. It would be great if we can remove drf-filters. Even if it requires few days of work.

JureZmrzlikar: DRF-filters removed, tests pass.

resolwe/flow/filters.py
JureZmrzlikar: While removing DRF-fiters, some of the syntax had to be changed. I tried to limit filter functionality to filter explicitly declared fields/lookups. Tests pass, but there might be cases where desired functionality is not tested. @lukaw3d and @hadalin should advise if there are which cases that are not covered by tests...

resolwe/flow/filters.py
```diff
@@ -5,26 +5,32 @@
...
-    created = filters.AutoFilter(lookups='__all__')
-    modified = filters.AutoFilter(lookups='__all__')
-
     class Meta:
         """Filter configuration."""
 
-        fields = ['id', 'slug', 'name', 'contributor', 'created', 'modified']
+        fields = {
+            'id': ['exact', 'in'],
+            'slug': ['exact', 'startswith'],
```
lukaw3d: slug__in

resolwe/flow/tests/test_filtering.py
```diff
@@ -551,10 +552,10 @@ def _apply_filter(self, filters, expected):
         self.assertCountEqual(filtered.qs, expected)
 
     def test_descriptor_completed(self):
-        self._apply_filter({'descriptor_completed': 'true'}, [self.entity_1])
-        self._apply_filter({'descriptor_completed': '1'}, [self.entity_1])
-        self._apply_filter({'descriptor_completed': 'false'}, [self.entity_2])
-        self._apply_filter({'descriptor_completed': '0'}, [self.entity_2])
+        self._apply_filter({'descriptor_completed': True}, [self.entity_1])
```
dblenkus: Filters are normally passed as a string from a client, so tests have to work the sam way.

resolwe/flow/tests/test_filtering.py
```diff
@@ -111,6 +111,7 @@ def test_filter_contributor(self):
         self._apply_filter({'contributor': self.user_1.pk}, [self.data_1, self.data_2])
 
     def test_filter_created(self):
+        self._apply_filter({'created': datetime.datetime(2016, 7, 30, 13, 59)}, [self.data_1])
```
dblenkus: Look at the comment below.

resolwe/flow/tests/test_filtering.py
```diff
@@ -604,3 +604,48 @@ def test_is_active(self):
...
+            contributor=self.user,
+            type='data:expression:',
+            category='analyses:',
+            scheduling_class='IN',
+        )
+
+    def _check_filter(self, query_args, expected, print_url=True, status_code=status.HTTP_200_OK):
+        request = factory.get('/', query_args, format='json')
+        if print_url:
+            print(request.build_absolute_uri())
```
dblenkus: Don't forget to remove prints.

resolwe/flow/filters.py
```diff
@@ -137,3 +171,11 @@ class Meta(BaseResolweFilter.Meta):
 
         model = Relation
         fields = BaseResolweFilter.Meta.fields
+
+    def get_always_allowed_arguments(self):
+        """Get always allowed query arguments."""
+        return super().get_always_allowed_arguments().union({
```
dblenkus: Add a comment that these arguments are used in thew view for manual filtering.

JureZmrzlikar:
This PR is ready for review.
@lukaw3d @dblenkus @mstajdohar

resolwe/flow/filters.py
```diff
@@ -65,54 +90,66 @@ def __init__(self, *args, **kwargs):
...
+    collection = filters.ModelChoiceFilter(field_name='collections', queryset=Collection.objects.all())
+    descriptor_completed = filters.rest_framework.filters.BooleanFilter(field_name='descriptor_completed')
     tags = TagsFilter()
 
     class Meta(CollectionFilter.Meta):
         """Filter configuration."""
 
         model = Entity
-        fields = BaseResolweFilter.Meta.fields + ['collections', 'descriptor_completed', 'tags']
+        fields = CollectionFilter.Meta.fields
```
dblenkus: You can probably omit this row as keeps the same value.

resolwe/flow/models/entity.py
```diff
@@ -10,6 +9,12 @@
 from .base import BaseModel
 from .collection import BaseCollection
 
+if settings.USE_TZ:
```
dblenkus: This pattern is used a lot. It would be nice to move it to `resolwe.utils` and import it from there. But it can be done in a separate pull request.

resolwe/flow/filters.py
JureZmrzlikar: Good point, thanks for spotting this.

resolwe/flow/models/entity.py
JureZmrzlikar: Good point, but let's do that in in a separate effort. I'll write it down.

PR #582, opened 2019-01-24
Add input Jinja filter to access input fields

mstajdohar:


mstajdohar:
cc @jvrakor

mstajdohar:
@jvrakor Please test this locally if it works for your use case.

resolwe/flow/tests/test_expression_filters.py
```diff
@@ -153,6 +174,7 @@ def test_templatetags(self):
         self.assertEqual(data.output['safe'], 'this')
         self.assertEqual(data.output['description_text'], 'This is test Data object.')
         self.assertEqual(data.output['list_description_text'], 'This is test Data object.')
+        self.assertEqual(data.output['input_id'], 1)
```
dblenkus: Change `1` to `inputs_input_data.pk` as this can change over time.

resolwe/flow/expression_engines/jinja/filters.py
```diff
@@ -36,6 +38,17 @@ def slug(data):
     return apply_filter_list(lambda datum: _get_data_attr(datum, 'slug'), data)
 
 
+def input_(data, field_path):
+    """Return a hydrated value of the ``input`` field."""
+    data_obj = Data.objects.get(id=data['__id'])
+
+    inputs = copy.deepcopy(data_obj.input)
+    hydrate_input_references(inputs, data_obj.process.input_schema)
```
dblenkus: This can be optimized by hydrating only the required field. But this would require a major refactoring. So just add a comment for now.

mstajdohar: Added

resolwe/flow/tests/test_expression_filters.py
mstajdohar: Changed

PR #584, opened 2019-01-28
Store ES build arguments cache in thread's local storage

dblenkus:


mstajdohar:
Add a change log entry.

PR #588, opened 2019-02-13
Log plumbum commands to standard output

jkokosar:


resolwe/process/runtime.py
```diff
@@ -8,6 +9,15 @@
 
 try:
     from plumbum import local as Cmd  # pylint: disable=unused-import
+    # Log plumbum commands to standard output.
+    root = logging.getLogger('plumbum.local')
+    root.setLevel(logging.DEBUG)
+
+    handler = logging.StreamHandler(sys.stdout)
+    handler.setLevel(logging.DEBUG)
```
dblenkus: Does this need to be set, as there is only one handler and level is already set for the logger?

resolwe/process/runtime.py
```diff
@@ -8,6 +9,15 @@
...
+    formatter = logging.Formatter('%(asctime)s - %(message)s')
```
dblenkus: Make this more verbose, to make it clear where the logs come from, e.g. `%(asctime)s - %(levelname)s - %(name)s[%(process)s]: %(message)s`.

resolwe/process/runtime.py
```diff
@@ -8,6 +9,15 @@
...
+    handler.setFormatter(formatter)
+    root.addHandler(handler)
```
dblenkus: Group the configurations together, i.e. first all `formatter` configuration, then `handler` and `root` at the end.

resolwe/process/runtime.py
```diff
@@ -8,6 +9,15 @@
 
 try:
     from plumbum import local as Cmd  # pylint: disable=unused-import
+    # Log plumbum commands to standard output.
+    root = logging.getLogger('plumbum.local')
```
dblenkus: Rename `root` to `logger`.

resolwe/process/runtime.py
mstajdohar: I kind-off wanted to avoid that in stdout.

dblenkus: The crucial information for me is `%(name)s` to know where the log comes from.

Do you need `%(asctime)s`? Dows any other tool in our processes uses time in logs?

mstajdohar: Why would a bioinformatician that runs the process need to know where the log comes from?

mstajdohar: The time is useful because the order of commands is switched sometimes in stdout.

dblenkus: > Why would a bioinformatician that runs the process need to know where the log comes from?

This is probably not useful now that there is only one source of logs, but it will be much more transparent when we add other sources, e.g. it would be nice to pipe all stdouts through logging.

resolwe/process/runtime.py
```diff
@@ -8,6 +9,16 @@
 
 try:
     from plumbum import local as Cmd  # pylint: disable=unused-import
+    # Log plumbum commands to standard output.
+    FORMATTER = logging.Formatter('%(asctime)s - %(levelname)s - %(name)s[%(process)s]: %(message)s')
```
dblenkus: These are not constants, so you should keep them lowercase - and disable pylint.

docs/CHANGELOG.rst
```diff
@@ -10,6 +10,11 @@ This project adheres to `Semantic Versioning <http://semver.org/>`_.
 Unreleased
 ==========
 
+Added
+-----
+- Add ``run_process`` method to ``Process`` to support triggering of a
```
dblenkus: Mention that this applies for python processes - here and also in the commit message.

PR #593, opened 2019-03-04
Add test for order by relevance

lukaw3d:


resolwe/flow/tests/test_filtering.py
```diff
@@ -240,6 +240,16 @@ def _check_filter(self, query_args, expected):
         for item in response.data:
             self.assertIn(item['id'], expected)
 
+    def _check_filter_ordered(self, query_args, expected):
```
JureZmrzlikar: 1. You should use this method instead of `_check_filter`. I guess you forgot... ;-)

2. My personal preference would be to add something like `check_ordering=False` parameter to `_check_filter` method instead of implementing this one. There is a lot of repetition in this method and it would take little effort to check the correct ordering if `check_ordering=True` in the `_check_filter`... But i think this is a matter of taste.

lukaw3d: Ops, got stuck in the middle between those two options
Fixed

resolwe/flow/tests/test_filtering.py
```diff
@@ -400,6 +403,14 @@ def test_filter_text(self):
         self._check_filter({'text': 'rst'}, [])
         self._check_filter({'text': 'process'}, self.data)
 
+        # Order by newest first.
+        self._check_filter({'text': 'Data 1'}, [self.data[11], self.data[10], self.data[1]], check_ordering=True)
+        # Order by relevance.
+        # Ordering by empty string should override DataViewSet's default ordering (-created)
+        # and return as ordered by Elastic Search.
+        self._check_filter({'text': 'Data 1', 'ordering': ''}, [self.data[1], self.data[11], self.data[10]],
```
lukaw3d: There seems to be some randomness here. One of previous force-pushes succeeded with 
`[self.data[1], self.data[11], self.data[10]]` but now wants `[self.data[11], self.data[1], self.data[10]]`

https://travis-ci.org/genialis/resolwe/builds/502007674 (click commit to view code that succeeded)

mstajdohar:
@dblenkus Could this be related to your seemingly random bug in `genialis-base`?

resolwe/flow/tests/test_filtering.py
```diff
@@ -400,6 +407,20 @@ def test_filter_text(self):
...
+                           [self.data[3], self.data[2], self.data[1], self.data[0]],
+                           check_ordering=True)
+
+        # Order by relevance.
+        # Ordering by empty string should override DataViewSet's default ordering (-created)
+        # and return as ordered by Elastic Search.
+        self._check_filter({'text': 'Hello', 'ordering': ''},
+                           # "Hello Hello", "hello hello", "hello", "hellodata"
+                           [self.data[1], self.data[0], self.data[3], self.data[2]],
+                           check_ordering=True)
```
lukaw3d: Does this look robust enough to overcome randomness?

lukaw3d: Nope

mstajdohar: Rename data objects to a single word, eg. Five and Six

dblenkus:
The error occurs when tests run in parallel.

PR #594, opened 2019-03-06
Add entity_names field to DataSerializer

hadalin:


resolwe/flow/views/entity.py
```diff
@@ -34,6 +34,12 @@ class EntityViewSet(CollectionViewSet):
         'descriptor_completed', 'collections', 'type'
     )
 
+    def get_queryset(self):  # pylint: disable=method-hidden
+        """Return queryset."""
+        # Override parent ``get_queryset`` to prevent prefetching
+        # ``entity_set``.
+        return self.queryset
```
dblenkus: You should all entities of data contained in current entity.

mstajdohar:
rebase

resolwe/flow/views/collection.py
```diff
@@ -56,6 +56,13 @@ class CollectionViewSet(ElasticSearchCombinedViewSet,
     }
     ordering = 'id'
 
+    def get_queryset(self):
+        """Return queryset."""
+        if self.request and self.request.query_params.get('hydrate_data', False):
+            return self.queryset.prefetch_related('entity_set')
```
lukaw3d: I'm appending collection_names and this looks incorrect. Should it be `'data__entity_set'`, like in entity.py/get_queryset https://github.com/genialis/resolwe/blob/784aeb2a21e6e87a398a299c4789dceb00bef951/resolwe/flow/views/entity.py#L37-L42

@hadalin / @dblenkus

PR #595, opened 2019-03-06
Fix linting errors

hadalin:


hadalin:
This is strange because `pylint` and `pycodestyle` are the same versions as when the build was successful.

dblenkus:
The problem is in `isort` linter.

PR #606, opened 2019-03-19
Add get_data_id_by_slug method to Process class

mstajdohar:


mstajdohar:
cc @jkokosar

resolwe/process/tests/processes/python_test.py
```diff
@@ -75,3 +75,6 @@ class PythonProcess2(process.Process):
     name = "Test Python Process 2"
     version = '0.0.1'
     process_type = 'data:python'
+
+    def run(self, inputs, outputs):
```
dblenkus: It would be nice to make this optional.

resolwe/api_urls.py
```diff
@@ -19,5 +19,5 @@
 
 
 urlpatterns = [  # pylint: disable=invalid-name
-    url(r'^', include(api_router.urls, namespace='resolwe-api')),
+    url(r'^api/', include(api_router.urls, namespace='resolwe-api')),
```
dblenkus: This should go into a separate commit.

resolwe/process/tests/processes/python_test.py
mstajdohar: I'll add this to the Python process wishlist

resolwe/api_urls.py
mstajdohar: done

PR #608, opened 2019-03-19
Reproducing test for issues with combined filters

JureZmrzlikar:


dblenkus:
This works as expected after upgrading to Elasticsearch 6 in https://github.com/genialis/resolwe/pull/623

PR #610, opened 2019-03-22
Python process syntax enhancements

mstajdohar:


mstajdohar:
@jkokosar Please test if this works for your scenario

resolwe/process/fields.py
```diff
@@ -593,7 +593,7 @@ def contribute_to_class(self, process, fields, name):
         : param fields: Fields registry to use
         : param name: Field name
         """
-        for field_name in dir(self.field_group):
+        for field_name in self.field_group.__dict__:
```
dblenkus: Add a comment why this is changed. You can also include this link: https://www.python.org/dev/peps/pep-0520/#an-order-preserving-cls-dict

resolwe/process/fields.py
```diff
@@ -458,9 +462,9 @@ class DataDescriptor:
     def __init__(self, data_id, field, cache=None):
         """Construct a data descriptor.
 
-        :param data_id: Data object primary key
-        :param field: Field this descriptor is for
-        :param cache: Optional cached object to use
+        : param data_id: Data object primary key
```
dblenkus: Do you need spaces here? We have such definitions without them all over the code.

resolwe/process/tests/processes/python_test.py
```diff
@@ -59,8 +59,10 @@ class MyGroup:
         my_group = GroupField(MyGroup, label="My group")
 
     class Output:
-        my_output = StringField(label="My output")
+        string_output = StringField(label="My string output")
```
dblenkus: Git history (and also a single commit) is much cleaner if you only change relevant code.

resolwe/process/fields.py
mstajdohar: done

resolwe/process/fields.py
mstajdohar: I have no idea how this got here. Fixed, thanks.

resolwe/process/tests/processes/python_test.py
mstajdohar: Well I had to add a `list_string_output` field for the test, so I had to change this field to be consistent...

mstajdohar:
@JakaKokosar please test if this PR works in your use case of interactive process updates.

JakaKokosar:
@mstajdohar sorry for the late reply, I was busy all day. I see this is already merged so I tested it with the latest master and it seems OK. I do get updated output fields now. Thanks!

docs/CHANGELOG.rst
```diff
@@ -10,9 +10,21 @@ This project adheres to `Semantic Versioning <http://semver.org/>`_.
...
+  ``input_field.file_field.path``.
+- Interactively save results in Python processes
+
 Added
 -----
 - Add get_data_id_by_slug method to Python processes' Process class
+- Python process syntax enhancements:
+
+  - Support ``.entity_name`` in data inputs
+  - Easy access to process resources through ``self.resources``
```
jvrakor: `self.requirements.resources`

PR #613, opened 2019-04-05
Simplify imports of mehtod now

JureZmrzlikar:
We used if-else logic to determine if the timetone support is
enabled in settings. Since this logic is already done in
django.utils.timezone.now we only need to import that.


mstajdohar:
rebase, please

JureZmrzlikar:
Rebased.

dblenkus:
This needs to be rebased again. Tests should be fixed now.

JureZmrzlikar:
Rebased.

PR #616, opened 2019-04-10
Python process enhacements

mstajdohar:


resolwe/flow/managers/listener.py
```diff
@@ -460,7 +462,17 @@ def handle_finish(self, obj):
...
 
-                if process_rc == 0 and not d.status == Data.STATUS_ERROR:
+                if spawning_failed:
+                    process_error = copy.copy(d.process_error)
+                    process_error.append("Error while preparing spawned Data objects")
+
+                    changeset = {
+                        'status': Data.STATUS_ERROR,
+                        'process_progress': 100,
+                        'finished': now(),
```
dblenkus: You can set `process_progress` and `finished` before the if.

resolwe/flow/managers/listener.py
```diff
@@ -460,7 +462,17 @@ def handle_finish(self, obj):
                     async_to_sync(self._send_reply)(obj, {ExecutorProtocol.RESULT: ExecutorProtocol.RESULT_ERROR})
                     return
 
-                if process_rc == 0 and not d.status == Data.STATUS_ERROR:
+                if spawning_failed:
+                    process_error = copy.copy(d.process_error)
+                    process_error.append("Error while preparing spawned Data objects")
```
dblenkus: Errors are appended, so there is no need to pass existing ones: https://github.com/genialis/resolwe/blob/3a1ae8077e27b5d640e3c69b64b27a4101be219b/resolwe/flow/managers/listener.py#L282

resolwe/flow/managers/listener.py
dblenkus: I would also prefere to set `process_progress` in `handle_update` when status is changed to `DONE` or `ERROR`. But this is out of scope for this PR.

resolwe/process/runtime.py
```diff
@@ -176,6 +176,17 @@ def run(self, inputs, outputs):
 
     def run_process(self, slug, inputs):
         """Run a new process from a running process."""
+        def export_files(d):
+            """Export input files of spawned process."""
+            for value in d.values():
+                if isinstance(value, dict):
```
dblenkus: You also need to support lists for `list:basic:file:` fields.

PR #619, opened 2019-04-15
Move validation logic for Data creation from view to serializer

JureZmrzlikar:


resolwe/flow/serializers/data.py
```diff
@@ -143,3 +143,23 @@ def get_fields(self):
...
+                collection = Collection.objects.get(pk=collection_id)
+            except Collection.DoesNotExist:
+                raise serializers.ValidationError('Invalid pk "{}" - object does not exist.'.format(collection_id))
+
+            if not self.request.user.has_perm('add_collection', obj=collection):
+                if self.request.user.has_perm('view_collection', obj=collection):
+                    msg = "You don't have `ADD` permission on collection (id: {}).".format(collection_id)
+                    raise exceptions.PermissionDenied(msg)
+                else:
+                    msg = "Collection not found (id: {}).".format(collection_id)
```
dblenkus: This message should be the same as the one when collection actually does not exist.

resolwe/flow/serializers/data.py
```diff
@@ -143,3 +143,23 @@ def get_fields(self):
...
+        # object will be added to.
+        for collection_id in self.request.data.get('collections', []):
+            try:
+                collection = Collection.objects.get(pk=collection_id)
+            except Collection.DoesNotExist:
+                raise serializers.ValidationError('Invalid pk "{}" - object does not exist.'.format(collection_id))
+
+            if not self.request.user.has_perm('add_collection', obj=collection):
+                if self.request.user.has_perm('view_collection', obj=collection):
+                    msg = "You don't have `ADD` permission on collection (id: {}).".format(collection_id)
```
dblenkus: It is cleaner if you insert the message directly into the exception, without an extra variable.

resolwe/flow/tests/test_serializers.py
```diff
@@ -148,6 +148,7 @@ def test_to_internal_value(self):
             'process': 'test-process',
             'descriptor_schema': 'test-schema',
         }
+        request.data = data
```
dblenkus: This looks like an anti-pattern that you have to pass the same thing twice. You can try to add a read-only `collections` field to the serializer as values will be (probably) passed directly to the `validate` method, but won't affect the save method (you can also use `validate_collections` instead of `validate`).

If this is too complicated, you can leave it like that as it will get much simpler once the new relations will be merged.

JureZmrzlikar: Agree. This was just an attempt to make tests happy. I would rather spare effort here and focus on porting m2m relations to m2one. Whean that is done, we can/should revisit this.

JureZmrzlikar:
Closed in favor of [#631](https://github.com/genialis/resolwe/pull/631).

PR #620, opened 2019-04-16
Always allow delete_content as query parameter

JureZmrzlikar:


resolwe/elastic/viewsets.py
```diff
@@ -146,7 +146,7 @@ def filter_search(self, search):
             unmatched.pop(argument, None)
 
         if unmatched:
-            msg = 'Unsupported filter argument(s): {}. Please use any of the supported arguments: {}.'.format(
+            msg = 'Unsupported parameter(s): {}. Please use any of the supported ones: {}.'.format(
```
mstajdohar: Please use one of:

resolwe/flow/filters.py
```diff
@@ -54,6 +54,7 @@ def get_always_allowed_arguments(self):
             'limit',
             'offset',
             'ordering',
+            'delete_content',
```
dblenkus: This should only go into collection and entity filters.

resolwe/elastic/viewsets.py
dblenkus: Move this to a separate commit.

resolwe/flow/filters.py
```diff
@@ -189,6 +196,12 @@ class Meta(BaseResolweFilter.Meta):
             'started': DATETIME_LOOKUPS[:],
         }}
 
+    def get_always_allowed_arguments(self):
```
dblenkus: This is unnecessary, as `hydrate_data` is never used on data endpoint.

JureZmrzlikar: OK, but why this:
https://github.com/genialis/resolwe/blob/master/resolwe/flow/views/data.py#L65

dblenkus: It can probably be removed: https://github.com/genialis/resolwe/blob/ddd7bdd887be1615d8dc8ff72508a7fc0c20c5c6/resolwe/flow/serializers/data.py#L110-L134

JureZmrzlikar: Done.

PR #626, opened 2019-05-13
Support requirements override in Python processes

mstajdohar:


mstajdohar:
cc @jvrakor, @jkokosar

PR #630, opened 2019-06-18
Add entity_new field to Process model

JureZmrzlikar:


resolwe/flow/management/commands/register.py
```diff
@@ -161,6 +161,7 @@ def register_processes(self, process_schemas, user, force=False, verbosity=1):
                 p['entity_type'] = p['entity']['type']
                 p['entity_descriptor_schema'] = p['entity'].get('descriptor_schema', p['entity_type'])
                 p['entity_input'] = p['entity'].get('input', None)
+                p['entity_new'] = p['entity'].get('new', False)
```
mstajdohar: I'd rather have a more verbose name, e.g. `entity_always_create`

mstajdohar:
Add a change log entry.

resolwe/flow/management/commands/register.py
JureZmrzlikar: Good point, done.

JureZmrzlikar:
@dblenkus Please review and merge if OK, i would like to use this in the downstream package in the following release cycle.

PR #632, opened 2019-06-22
Optimize queries for retrieving permissions in elastic indices

dblenkus:
This produces following queries:
```
SELECT "guardian_userobjectpermission"."user_id" FROM "guardian_userobjectpermission" WHERE ("guardian_userobjectpermission"."object_pk" = '2' AND "guardian_userobjectpermission"."permission_id" = (SELECT U0."id" FROM "auth_permission" U0 WHERE (U0."codename"::text LIKE 'view%' AND U0."content_type_id" = 16)));
```
```
SELECT "guardian_groupobjectpermission"."group_id" FROM "guardian_groupobjectpermission" WHERE ("guardian_groupobjectpermission"."object_pk" = '2' AND "guardian_groupobjectpermission"."permission_id" = (SELECT U0."id" FROM "auth_permission" U0 WHERE (U0."codename"::text LIKE 'view%' AND U0."content_type_id" = 16)));
```
```
SELECT (1) AS "a" FROM "guardian_userobjectpermission" WHERE ("guardian_userobjectpermission"."object_pk" = '2' AND "guardian_userobjectpermission"."permission_id" = (SELECT U0."id" FROM "auth_permission" U0 WHERE (U0."codename"::text LIKE 'view%' AND U0."content_type_id" = 16)) AND "guardian_userobjectpermission"."user_id" = (SELECT U0."id" FROM "auth_user" U0 WHERE U0."username" = 'public'))  LIMIT 1;
```

### DIFF
--- a/resolwe/elastic/indices.py
+++ b/resolwe/elastic/indices.py
@@ -23,7 +23,10 @@ from elasticsearch.helpers import bulk
 from elasticsearch_dsl.connections import connections
 from elasticsearch_dsl.exceptions import IllegalOperation
 
+from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Permission
 from django.contrib.contenttypes.models import ContentType
+from django.db.models import Subquery
 
 from django_priority_batch import PrioritizedBatcher
 from guardian.conf.settings import ANONYMOUS_USER_NAME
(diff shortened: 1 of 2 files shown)


resolwe/elastic/indices.py
```diff
@@ -406,19 +409,29 @@ def get_permissions(self, obj):
...
+
         filters = {
             'object_pk': obj.id,
-            'content_type': ContentType.objects.get_for_model(obj),
-            'permission__codename__startswith': 'view',
+            'permission': permissions_subquery,
         }
         return {
             'users': list(
-                UserObjectPermission.objects.filter(**filters).distinct('user').values_list('user_id', flat=True)
```
mstajdohar: Don't we need `distinct` users?

dblenkus: No, because we don't do any joins, so we cannot get duplicated rows.

PR #634, opened 2019-06-28
Move dependency filtering to a separate API endpoint

dblenkus:


resolwe/flow/views/data.py
```diff
@@ -212,3 +212,25 @@ def perform_create(self, serializer):
                 for entity in entities:
                     entity.collections.add(collection)
                     copy_permissions(collection, entity)
+
+    def _parents_children(self, request, queryset):
```
JureZmrzlikar: This method is much more general than just processing parents and children queryset and it should be named accordingly. `_serialize_queryset` or something similar would be a better choice IMO.

dblenkus: I was considering this, but at the end I've decided to make a more dedicated method, as process the query in a bit specific way, i.e. it applies `get_objects_for_user` and it doesn't do any filtering.

We can still generalize the function and use it elsewhere when we will have a use-case.

JureZmrzlikar: OK, sounds reasonable.

PR #640, opened 2019-08-08
Add projection to DictRelatedField

JureZmrzlikar:
This fix enables the use of `fields` query parameter for DictRelatedField (Process, DescriptorSchema, collection and Entity).

### DIFF
--- a/resolwe/flow/models/entity.py
+++ b/resolwe/flow/models/entity.py
@@ -109,13 +109,22 @@ class Entity(BaseCollection):
 
     @transaction.atomic
     def move_to_collection(self, source_collection, destination_collection):
-        """Move entity from source to destination collection."""
-        # Remove from collection.
-        source_collection.data.remove(*self.data.all())  # pylint: disable=no-member
-        destination_collection.data.add(*self.data.all())  # pylint: disable=no-member
+        """Move entity to destination collection."""
+        if source_collection == destination_collection:
+            return
 
         self.collection = destination_collection
+        if destination_collection:
+            self.tags = destination_collection.tags
+            copy_permissions(destination_collection, self)
         self.save()
+
+        for datum in self.data.all():  # pylint: disable=no-member
+            datum.collection = destination_collection
+            if destination_collection:
+                datum.tags = destination_collection.tags
+                copy_permissions(destination_collection, datum)
+            datum.save()
 
 
 class RelationType(models.Model):
--- a/resolwe/flow/serializers/fields.py
+++ b/resolwe/flow/serializers/fields.py
@@ -72,4 +72,10 @@ class DictRelatedField(relations.RelatedField):
 
     def to_representation(self, obj):
         """Convert to representation."""
-        return self.serializer(obj, required=self.required).data
+        serializer = self.serializer(obj, required=self.required, context=self.context)
+
+        # Manually bind this serializer to field.parent as field projection
+        # relies on parent/child relations between serializers.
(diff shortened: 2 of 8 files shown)


resolwe/flow/serializers/fields.py
```diff
@@ -72,4 +73,5 @@ def to_internal_value(self, data):
 
     def to_representation(self, obj):
         """Convert to representation."""
-        return self.serializer(obj, required=self.required).data
+        data = self.serializer(obj, required=self.required).data
+        return apply_subfield_projection(self, data, deep=True)
```
dblenkus: This is not a good solution, because `.data` calculates all fields and only then applies projection.

dblenkus: Why dont you use [`SelectiveFieldMixin`](https://github.com/genialis/resolwe/blob/7ee372a9700a4f3fedd621cf551567354741d9aa/resolwe/rest/serializers.py#L7)?

JureZmrzlikar: We use the same approach in [ProjectableJSONField](https://github.com/genialis/resolwe/blob/master/resolwe/rest/fields.py#L12).

`SelectiveFieldMixin` is already in use through `ResolweBaseSerializer` but it does not suffice.

dblenkus: Difference is that JSON field is read as a whole from the database and you get a dict. In this case you have to build that dict (serialize an object), so there is a lot overhead if you do it in a wrong order. Actually you lose all benefits gained by limiting the fields on the API.

dblenkus: All nested object in DRF are [presented as serializers](https://www.django-rest-framework.org/api-guide/relations/#nested-relationships), not as fields. I think that this extra field around the actual serializer is causing problems.

JureZmrzlikar: True, thanks for the hint. I believe i have now correctly addressed this.

JureZmrzlikar:
@dblenkus I've added another commit which handles Data.collection assignment when Sample.collection is changed. Review when you find time.

resolwe/flow/models/entity.py
```diff
@@ -109,14 +109,20 @@ def duplicate(self, contributor=None, inherit_collection=False):
 
     @transaction.atomic
     def move_to_collection(self, source_collection, destination_collection):
-        """Move entity from source to destination collection."""
-        # Remove from collection.
-        source_collection.data.remove(*self.data.all())  # pylint: disable=no-member
-        destination_collection.data.add(*self.data.all())  # pylint: disable=no-member
-
+        """Move entity to destination collection."""
```
dblenkus: Add:
```
if source_collection == destination_collection:
    return
```

resolwe/flow/serializers/fields.py
```diff
@@ -41,7 +41,10 @@ def model_name(self):
 
     def to_internal_value(self, data):
         """Convert to internal value."""
-        if data.get('id', None) is not None:
+        # Allow None values only in case field is not required.
+        if 'id' in data and data['id'] is None and not self.required:
```
dblenkus: Is this `required` check needed or is it handled at validation step?

JureZmrzlikar: Not required. Removed. What bothers me is that if a required field is updated to None (for example Data.process=None), the response is http status code 204, but the field stays unchanged. Should we raise error on such requests?

PR #641, opened 2019-08-18
Fix deep projection of nested serializers

JureZmrzlikar:


resolwe/rest/tests.py
```diff
@@ -69,6 +69,27 @@ def test_projection(self):
                 },
             },
         })
+        # Test deep projection: an empty projection means that all fields
+        # should be returned without any projection.
+        data = self.get_projection(['entity'])[0]
+        self.assertEqual(set(data['entity'].keys()), {
+            'created',
```
lukaw3d: you probably want assertCountEqual here

lukaw3d: ops, it's a set

PR #646, opened 2019-08-27
Add 'version' to minimal example

zagm:
In really minimal example, which I always use to see if everything was set up properly is missing version string in order to avoid always modifying it once downloaded, I was wondering if we could update the documentation.


dblenkus:
This is example of a minimal working process and it doesn't need version.

PR #651, opened 2019-11-06
Docker cpu limit

acopar:
Changes: 
- Update docker run command with --cpus option and remove --cpu-shares option.
This option is available since docker 1.13 and is set to unlimited by default. The cpus is set to the number of cores provided in process object. This limits most processes to single-threaded execution. 

- The default timeout is increased by factor 1.2. Ulimit option becomes unreliable on multithreaded processes, increase in timeout allows for some difference between realtime and measured time.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -18,6 +18,11 @@ Changed
   collections
 - **BACKWARD INCOMPATIBLE:** Remove ``Entity.descriptor_completed`` field
 
+Fixed
+-----
+- Fix docker executor command with ``--cpus`` limit option. This solves the
+  issue where process is killed before the timeout 30s is reached
+
 
 ===================
 19.1.0 - 2019-09-17
--- a/resolwe/flow/executors/docker/run.py
+++ b/resolwe/flow/executors/docker/run.py
@@ -62,9 +62,9 @@ class FlowExecutor(LocalFlowExecutor):
 
         # Set resource limits.
         limits = []
-        # Each core is equivalent to 1024 CPU shares. The default for Docker containers
-        # is 1024 shares (we don't need to explicitly set that).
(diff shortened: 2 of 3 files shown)


resolwe/flow/executors/docker/run.py
```diff
@@ -79,7 +79,10 @@ def _generate_container_name(self):
         # Set ulimits for interactive processes to prevent them from running too long.
         if self.process['scheduling_class'] == PROCESS_META['SCHEDULING_CLASS_INTERACTIVE']:
             # TODO: This is not very good as each child gets the same limit.
-            limits.append('--ulimit cpu={}'.format(limit_defaults.get('cpu_time_interactive', 30)))
+            # Note: Ulimit does not work as expected on multithreaded processes
+            # Limit is increased by factor 1.2 for processes with 2-8 threads.
+            # TODO: This should be changed for processes with over 8 threads.
+            limits.append('--ulimit cpu={}'.format(int(limit_defaults.get('cpu_time_interactive', 30)*1.2)))
```
dblenkus: Try to keep lines under 100 characters, e.g.:
```
            limits.append(
                '--ulimit cpu={}'.format(int(limit_defaults.get('cpu_time_interactive', 30) * 1.2))
            )
```
or (maybe even better):
```
            cpu_time_interactive = limit_defaults.get('cpu_time_interactive', 30)
            limits.append('--ulimit cpu={}'.format(int(cpu_time_interactive * 1.2)))
```

docs/CHANGELOG.rst
```diff
@@ -18,6 +18,11 @@ Changed
   collections
 - **BACKWARD INCOMPATIBLE:** Remove ``Entity.descriptor_completed`` field
 
+Fixed
+-----
+- Fix docker executor command with ``--cpus`` limit option. This solves the
```
dblenkus: This should go into the same commit as the change it describes.

PR #658, opened 2019-12-19
Use chunked delete for deleting Storage objects in purge

dblenkus:


resolwe/flow/models/base.py
```diff
@@ -12,6 +12,45 @@
 MAX_SLUG_RETRIES = 10
 
 
+# NOTE: This method is used in migrations.
+def delete_chunked(queryset, chunk_size=500):
```
mstajdohar: Why did you move the method out of the QuerySet?

mstajdohar: Because of migrations?

mstajdohar: Why didn't it work when called `DataQuerySet._delete_chunked `?

dblenkus: Custom queryset is not initialized in migrations. It feels less hacky to me to make a separate function than making the method static.

mstajdohar: > Custom queryset is not initialized in migrations

Thank you for the explanation.

PR #659, opened 2020-01-10
Move full-text search to PostgreSQL

dblenkus:


resolwe/flow/models/process.py
```diff
@@ -19,6 +20,13 @@ class Meta(BaseModel.Meta):
             ("owner_process", "Is owner of the process"),
         )
 
+        indexes = [
+            models.Index(name='idx_process_name', fields=['name']),
```
mstajdohar: Do we need an index on process version?

PR #660, opened 2020-01-17
Start using black

dblenkus:


pyproject.toml
```diff
@@ -4,3 +4,7 @@ requires = [
     "setuptools_scm >= 3.1.0, <4",
 ]
 build-backend = "setuptools.build_meta"
+
+[tool.black]
+target-version = ["py36"]
+skip-string-normalization = true
```
mstajdohar: Think whether to include string normalization or not
Think whether we should set the line length to 119
Add a line

setup.cfg
```diff
@@ -1,25 +1,25 @@
...
 known_django = django
 known_djangothird=django_priority_batch, rest_framework, guardian
 sections=FUTURE,STDLIB,THIRDPARTY,DJANGO,DJANGOTHIRD,FIRSTPARTY,LOCALFOLDER
+default_section = THIRDPARTY
 skip = migrations
 not_skip = __init__.py
+
+[flake8]
+; E,W - disable pycodestyle checks as they may conflict with black
+ignore = E,W
```
mstajdohar: Add a line

PR #662, opened 2020-01-21
Enable running Python processes with worflows as inputs

JureZmrzlikar:


docs/CHANGELOG.rst
```diff
@@ -15,6 +15,11 @@ Changed
 - Add ``username`` to ``current_user_permissions`` field of objects on API
 
 
+Fixed
+-----
+- Enable running Python processes with worflows as inputs
```
mstajdohar: Enable -> Support. Or just "Run Python...". Change the commit message too.

resolwe/process/fields.py
```diff
@@ -513,9 +516,17 @@ def __init__(self, data_id, field, cache):
...
             if field_type.startswith("list:"):
-                field = ListField(ALL_FIELDS_MAP[field_type[5:]]())
+                inner_field_class = ALL_FIELDS_MAP[field_type[5:]]
+                if issubclass(inner_field_class, DataField):
+                    field = ListField(inner_field_class(data_type=field_type[10:]))
+                else:
+                    field = ListField(inner_field_class())
             else:
-                field = ALL_FIELDS_MAP[field_type]()
+                field_class = ALL_FIELDS_MAP[field_type]
```
mstajdohar: This branch is not covered in tests is it?

JureZmrzlikar:
@mstajdohar Thanks, for your review. It made me think again and i decided to solve the problem in a bit more simplistic way (at least IMO). @jkokosar Can you please verify that this solution still works in your case?

mstajdohar:
Much better, the code is now more readable, thanks.

resolwe/process/fields.py
```diff
@@ -501,8 +501,11 @@ def __init__(self, data_id, field, cache):
         """
         super().__setattr__("_data_id", data_id)
 
+        if not cache:
```
dblenkus: You should set the default value of the `cache` to `None` if it is not required.

resolwe/process/fields.py
```diff
@@ -512,7 +515,13 @@ def __init__(self, data_id, field, cache):
 
             field_type = field_descriptor["type"].rstrip(":")
 
-            if field_type.startswith("list:"):
+            if field_type.startswith("list:data"):
+                # The only case where output field is of type
```
dblenkus: Why is that?

resolwe/process/fields.py
```diff
@@ -512,7 +515,13 @@ def __init__(self, data_id, field, cache):
 
             field_type = field_descriptor["type"].rstrip(":")
 
-            if field_type.startswith("list:"):
+            if field_type.startswith("list:data"):
```
dblenkus: Add colon at the end.

docs/CHANGELOG.rst
```diff
@@ -15,6 +15,11 @@ Changed
 - Add ``username`` to ``current_user_permissions`` field of objects on API
 
 
+Fixed
+-----
+- Support worflows as inputs to Python processes
```
dblenkus: If change log entry starts with the "Support", then it is not a fix. Move it under "Added" or rephrase it to "Fix workflows to work with..."

resolwe/process/fields.py
```diff
@@ -576,6 +585,8 @@ def __repr__(self):
 class DataField(Field):
     """Data object field."""
 
+    field_type = "data"
```
dblenkus: (This should go into a separate commit.)

resolwe/process/fields.py
JureZmrzlikar: Done.

resolwe/process/fields.py
JureZmrzlikar: This is just how we currently do it. There are no processes in resolwe/resolwe-bio that would have data objects as outputs (except for workflows).

resolwe/process/fields.py
JureZmrzlikar: Check two lines above. :-)

docs/CHANGELOG.rst
JureZmrzlikar: Done.

resolwe/process/fields.py
JureZmrzlikar: Why? This enables the fix to work...

resolwe/process/fields.py
JureZmrzlikar: > I don't like this solution, because it is not 100% correct :)

Why do you mean by "it is not 100% correct"?

resolwe/process/fields.py
JureZmrzlikar: :+1:

resolwe/process/fields.py
dblenkus: If someone uses type `list:dataid:`, it would match this condition instead of `elif`.

JureZmrzlikar: How about `if field_descriptor["type"].startswith("list:data:"):`?

dblenkus: Then it won't match fields of type `list:data:`, because final colon is striped in the line above.

JureZmrzlikar: It is not stripped in `field_descriptor["type"]`

dblenkus: Sorry, i haven't seen this change :) (`field_type` to `field_descriptor["type"]`).

Yes, that would also work, but you have to add a test for this or at least a comment.

PR #664, opened 2020-01-24
Update collection filtering fields

hadalin:


resolwe/flow/views/collection.py
```diff
@@ -51,14 +51,18 @@ class CollectionViewSet(
...
         "contributor",
+        "contributor_name",
         "owners",
+        "owners_name",
         "text",
         "tags",
     )
     filtering_map = {
-        "name": "name.raw",
+        "name": "name.ngrams",
```
JureZmrzlikar: Is this really necessary? Becouse this was changed some time ago, to adress the issue in SDK where User would serch for Collectin with name=="Foo" and would get all sorts of collections where the names would be "Foo bar", "Foo bar is soo cool", "My best word is foo", ... Which is highly undesired behaviour.

hadalin: It's needed in some cases. Can we create a separate name filter, something like `name_contains`?

JureZmrzlikar: Sure, that's fine for me.

PR #666, opened 2020-01-28
Support Python 3.8

dblenkus:


tox.ini
```diff
@@ -19,8 +19,9 @@ basepython = python3
 
 [testenv]
 basepython =
-    !py37: python3.6
+    !py{37,38}: python3.6
```
acopar: I suggest changing this to:
py36,docs,linters,packaging,migrations: python3.6

Tox parser does not work as one would expect. It will match all conditions and concatenate the results. For example, Python 3.7 will match conditions in lines 22 and 23, (!py38 and py37 conditions), therefore interpreter is defined as 'python3.6\npython3.7', which not a valid command. Similar happens for Python 3.8.

dblenkus: Thanks!

dblenkus:
Included in https://github.com/genialis/resolwe/pull/668

PR #670, opened 2020-02-21
Support extensions for API filters

dblenkus:


resolwe/composer.py
```diff
@@ -0,0 +1,50 @@
+"""Support for easier dynamic composition of type extensions."""
```
gregorjerse: This comment repeats itself in the class definition bellow.

resolwe/composer.py
```diff
@@ -0,0 +1,50 @@
...
+    """Support for easier dynamic composition of type extensions."""
+
+    def __init__(self):
+        """Construct new composer instance."""
+        self._extensions = {}
+
+    def _get_class_path(self, klass_or_instance):
+        """Return class path for a given class.
+
+        :param klass_or_instance: Class or instance of given class
```
gregorjerse: According to the code, this argument can also be string.

resolwe/composer.py
```diff
@@ -0,0 +1,50 @@
...
+        :return: String containing the class path
+        """
+        if inspect.isclass(klass_or_instance):
+            klass = "{}.{}".format(
```
gregorjerse: It would be nicer to have a method for constructing class name and use that one here and bellow. Maybe consider using __qualname__ instead of __name__  since it handles nested classes properly. But this is not needed in this situation.

resolwe/flow/filters.py
```diff
@@ -158,7 +161,24 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
 
-class BaseResolweFilter(CheckQueryParamsMixin, filters.FilterSet):
+class ResolweFilterMetaclass(FilterSetMetaclass):
+    """Metaclass for Resolwe filters."""
+
```
gregorjerse: I believe this code deserves some comment. At least explain that our extensions for the given class are injected as argument attrs in FilterSetMetaclass and that they are loaded in there.

This solution is very nice, but it depends on the actual implementation of django_filters package, not just its API. Maybe it would be nice to mention that in a comment for easier debugging if/when the implementation of django_filters changes.

resolwe/composer.py
```diff
@@ -0,0 +1,55 @@
...
+        self._extensions = {}
+
+    def _get_class_path(self, klass_or_instance):
+        """Return class path for a given class.
+
+        :param klass_or_instance: Class or instance of given class or
+            class path as a string
+        :return: String containing the class path
+        """
+        def construct_class_path(module, klass):
```
gregorjerse: This method could also just accept single argument klass (Class).

PR #672, opened 2020-02-25
Simplify full-text search indexes

dblenkus:


resolwe/flow/filters.py
```diff
@@ -354,3 +366,30 @@ class Meta(BaseResolweFilter.Meta):
...
+class OrderingFilter(DrfOrderingFilter):
+    """Order results by field specified in request.
+
+    If no ordering is given, default ordering is used. Default ordering
+    is skipped for views filtered by full-text search as it already
+    applies ordering by ranking.
+    """
+
+    def get_ordering(self, request, queryset, view):
+        """Return name of the filed by which resukts should be ordered.
```
gregorjerse: resukts should be results

resolwe/flow/filters.py
```diff
@@ -354,3 +366,30 @@ class Meta(BaseResolweFilter.Meta):
...
+
+        Ordering is skipped if ``None`` is returneed.
```
gregorjerse: returneed should be returned

PR #673, opened 2020-03-02
Use DictRelatedField for collection field in RelationSerializer

JureZmrzlikar:


JureZmrzlikar:
@hadalin This requires update of the fronend, how long will it take you?

resolwe/flow/tests/test_relations.py
```diff
@@ -385,20 +385,21 @@ def test_update(self):
...
     def test_update_different_user(self):
-        data = {"collection": self.collection_2.pk}
+        data = {"collection": {"id": self.collection_2.pk}}
         resp = self._patch(self.relation_group.pk, data, user=self.user)
         self.assertEqual(resp.status_code, status.HTTP_404_NOT_FOUND)
         self.relation_group.refresh_from_db()
         self.assertEqual(self.relation_group.collection, self.collection)
 
         assign_contributor_permissions(self.collection, self.user)
+        assign_contributor_permissions(self.collection_2, self.user)
```
dblenkus: Is this line required for this fix?

JureZmrzlikar: Yes, test fail otherwise.

dblenkus: What is the reason for this?

JureZmrzlikar: If you update the collection field in relation instance (from C1 to C2)...
Before: only permissions on C1 were checked (in views/relations.py) but not for C2
Now: Besides permissions for C1 olso permissions for C2 are checked (this happens in DictRelatedField, [here](https://github.com/genialis/resolwe/blob/master/resolwe/flow/serializers/fields.py#L68)).

PR #678, opened 2020-03-12
Support allow_custom_choice field property in Python processes

jkokosar:


dblenkus:
Merged in https://github.com/genialis/resolwe/pull/679

PR #680, opened 2020-03-27
Make filtering via foreign key more Django like

gregorjerse:
When foreign key does not exist return empty set instead of raising
validation exception.

Also when filtering using list of foreign keys do not raise validation
exception if some foreign keys in the list do not exist.


dblenkus:
Please add tests: https://github.com/genialis/resolwe/blob/master/resolwe/flow/tests/test_filtering.py

PR #681, opened 2020-03-27
Add data_count and status fields to Collection serializer

hadalin:


resolwe/flow/serializers/collection.py
```diff
@@ -18,6 +20,28 @@ class CollectionSerializer(ResolweBaseSerializer):
...
+        if collection.data_processing_count:
+            return Data.STATUS_PROCESSING
+        if collection.data_waiting_count:
+            return Data.STATUS_WAITING
+        if collection.data_resolving_count:
+            return Data.STATUS_RESOLVING
+        if collection.data_done_count == collection.data_count:
+            return Data.STATUS_DONE
+
+        return Data.STATUS_ERROR
```
dblenkus: In which case is this returned? If there is no specific case, I would prefer to return `None`.

resolwe/flow/serializers/collection.py
```diff
@@ -18,6 +20,28 @@ class CollectionSerializer(ResolweBaseSerializer):
...
+            return Data.STATUS_ERROR
+        if collection.data_uploading_count:
+            return Data.STATUS_UPLOADING
+        if collection.data_processing_count:
+            return Data.STATUS_PROCESSING
+        if collection.data_waiting_count:
+            return Data.STATUS_WAITING
+        if collection.data_resolving_count:
+            return Data.STATUS_RESOLVING
+        if collection.data_done_count == collection.data_count:
```
dblenkus: You can move this if to the top and remove `== collection.data_count` part as the status will be overridden if any other data exists.

resolwe/flow/tests/test_collections.py
```diff
@@ -0,0 +1,93 @@
...
+from django.contrib.auth.models import AnonymousUser
+from django.urls import reverse
+
+from guardian.shortcuts import assign_perm
+from rest_framework.test import APITestCase
+
+from resolwe.flow.models import Collection, Data, Process
+
+
+class CollectionTests(APITestCase):
```
dblenkus: There is already a [TestCollectionViewSetCase](https://github.com/genialis/resolwe/blob/b52d1223a98efe676758e6b65c4dd691696edb3d/resolwe/flow/tests/test_api.py#L431), so it would make sense to merge them.

resolwe/flow/serializers/collection.py
hadalin: This line of code should never be reached so I'm thinking that raising an exception would be more appropriate.

dblenkus: It will be reached if we introduce a new Data status. I think it would be better to return `None` than to raise an error in that case.

You can remove `if collection.data_count == 0:` at the top and return `None` here.

hadalin: If we introduce a new data status, then it would be nice to log if the status cannot be determined.

PR #682, opened 2020-03-30
Enable all top-level class definitions in Python processes

JureZmrzlikar:
Currently, process registration fails if there is a top level
class definition that does not inherit from
resolwe.process.runtime. This commit fixes this limitation.

Typical case of other top-level classes are custom classes needed
in the run method.

@dblenkus I'd really like to see this fix in the April release. Is it reasonable to expect that i will?

### DIFF
--- a/resolwe/process/parser.py
+++ b/resolwe/process/parser.py
@@ -1,6 +1,7 @@
 """Safe Resolwe process parser."""
 import ast
 import collections
+from inspect import isclass
 import json
 
 import asteval
@@ -173,7 +174,7 @@ class ProcessVisitor(ast.NodeVisitor):
             else:
                 continue
 
-            if issubclass(base, runtime.Process):
+            if isclass(base) and issubclass(base, runtime.Process):
                 break
         else:
             return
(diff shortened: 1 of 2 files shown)


resolwe/process/parser.py
```diff
@@ -173,6 +173,12 @@ def visit_ClassDef(self, node):
             else:
                 continue
 
+            # Value ``None`` can occur if there is another top level class
+            # definition in the module that does not inherit from classes
+            # defined in ``runtime``.
+            if base is None:
```
dblenkus: It would make more sense to put this condition in the if below:
```
if base is not None and issubclass(base, runtime.Process):
    break
```
Or to make it even more robust:
```
from inspect import isclass

if isclass(base) and issubclass(base, runtime.Process):
    break
```
This is also more self-explanatory, so you can also remove the comment.

JureZmrzlikar: Good point. Fixed.

JureZmrzlikar:
Tests fail for totally unrelated reason. Can someone with sufficient permissions on Travis rerun Py3.6 part of the build?

PR #700, opened 2020-04-24
Add entity_always_create in Process serializer

JureZmrzlikar:
This should be done long ago, but we realized only now that it is missing.


JureZmrzlikar:
Merge please.

PR #701, opened 2020-04-24
Add Python process template

romunov:
Starting to write a Python process can be daunting, especially for junior developers. While there are example processes in the test files, there is no comprehensive documentation on the subject yet.

Towards this end, I am committing a template of a Python process which would help developers add a process that already includes all necessary components. It also includes some comments on process specifics, available input/output fields, how to access variables, how to run external tools, raise errors etc.

Let me know if location of the template is appropriate within the example. One thing I did not do here is linking to the template from any documentation. If you think this should be done, please recommend from which document this would make most sense (i.e. `proc.rst`?).

### DIFF
--- a/docs/example/example/processes/template_py_process.py
+++ b/docs/example/example/processes/template_py_process.py
@@ -1,0 +1,277 @@
+"""Describe what tasks this Process accomplishes."""
+# Order of imported packages should be:
+# Standard libraries
+from pathlib import Path
+from shutil import copy2
+
+# Third party libraries
+from plumbum import TEE
+
+# Resolwe
+from resolwe.process import (
+    Process,
+    SchedulingClass,
+    Persistence,
+    Cmd,
+    StringField,
+    TextField,
+    BooleanField,
+    IntegerField,
+    FloatField,
+    DateField,
+    DateTimeField,
+    UrlField,
+    SecretField,
+    FileField,
+    FileHtmlField,
+    DirField,
+    JsonField,
+    ListField,
+    DataField,
+    GroupField,
+)
+
+# In-house libraries (i.e. resolwe_bio)
+# Local libraries
+
+
+class YourProcessName(Process):
+    """Process description in one line ends with a period.
+
+    Continue with a more elaborate description of what the process
+    does, short layout of the algorithm and what output(s) to expect.
+
+    Note that this is not a working process, but rather just shows
+    bits and pieces that can be used in your real process.
+    """
+
+    slug = "process-slug-name"
+    name = "Human readable name of the process"
+    process_type = "data:text_processing:"
+    version = "1.0.0"
+    # A category of the process, can be used in i.e. frontend app.
+    category = "Read Processing"
+    # Dynamically generated name of the resulting datum object. Lookup
+    # happens from Input class fields.
+    data_name = 'Data object ({{ string_field.file|default("?") }})'
+    scheduling_class = SchedulingClass.BATCH  # or .INTERACTIVE for short jobs
+    # or RAW or TMP, see documentation at
+    # https://resolwe.readthedocs.io/en/latest/proc.html#persistence
+    persistence = Persistence.CACHED
+    # Use below entity to make the created object also an entity of
+    # type sample. See
+    # https://resolwe.readthedocs.io/en/latest/proc.html#entity for
+    # more details.
+    # Fields descriptor_schema, input and always_create are optional.
+    entity = {
+        "type": "sample",  # name of your entity
+        # Use a descriptor schema slug if it differs from "type".
+        "descriptor_schema": "descriptor-schema-of-entity",
+        "input": "name of input that has the entity",
+        # If you want to create entity in all cases
+        "always_create": True,
+    }
+    requirements = {
+        "expression-engine": "jinja",
+        "executor": {
+            "docker": {
+                # Specify Docker image, may need to include a version.
+                "image": "resolwe/base:fedora-31"
+            }
+        },
+        "resources": {
+            "cores": 8,
+            "memory": 16384,
+            # Connects container to a network.
+            "network": True,
+        },
+    }
+
+    class Input:
+        """Input fields to process MergeData."""
+
+        string_field = StringField(
+            label="Labels are short and do not end in a period",
+            description="Description ends in a period.",
+            choices=[
+                ("computer_readable1", "Human readable 1"),
+                ("computer_readable2", "Human readable 2"),
+            ],
+            default="computer_readable1",
+            required=False,
+            hidden=False,
+            allow_custom_choice=True,
+        )
+        text_field = TextField(
+            label="Labels are short and do not end in a period",
+            description="Description ends in a period.",
+            default="default text",
+            required=False,
+            hidden=True,
+        )
+        boolean_field1 = BooleanField(
+            label="Labels are short and do not end in a period",
+            description="Note that description fields always end in a period.",
+            default=False,
+            required=True,
+            hidden=False,
+        )
+        integer_field = IntegerField(
+            label="Labels are short and do not end in a period",
+            description="Description ends in a period.",
+            default=1,
+        )
+        float_field = FloatField(
+            label="Labels are short and do not end in a period",
+            description="Description ends in a period.",
+            default=3.14,
+        )
+        date_field = DateField(
+            label="Labels are short and do not end in a period",
+            description="Description ends in a period.",
+            default="2020-04-20",
+        )
+        datetime_field = DateTimeField(
+            label="Labels are short and do not end in a period",
+            description="Description ends in a period.",
+            default="2020-04-20 12:16:00",
+        )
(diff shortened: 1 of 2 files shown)


docs/example/example/processes/template_py_process.py
```diff
@@ -0,0 +1,277 @@
+"""Describe what tasks this Process accomplishes."""
+# Order of imported packages should be:
+# future (Python 2)
```
dblenkus, line 3: There is no Python 2 anymore.

docs/example/example/processes/template_py_process.py
```diff
@@ -0,0 +1,277 @@
...
+# Standard libraries
+from pathlib import Path
+from shutil import copy2
+
+# Third party libraries
+from plumbum import TEE
+
+# resolwe
```
dblenkus, line 11: Resolwe should be capitalized.

docs/example/example/processes/template_py_process.py
```diff
@@ -0,0 +1,277 @@
...
+
+    Continue with a more elaborate description of what the process
+    does, short layout of the algorithm and what output(s) to expect.
+
+    Note that this is not a working process, but rather just shows
+    bits and pieces that can be used in your real process.
+    """
+
+    slug = "merge-reads"
+    name = "Merged reads"  # human readable name of the process
```
dblenkus, line 50: Capitalize comment. Put all comments above the line they refer to. And add a dot et the end.

docs/example/example/processes/template_py_process.py
```diff
@@ -0,0 +1,277 @@
...
+    process_type = "data:reads:merged:"
+    version = "1.0.0"
+    # See a list of existing categories in resolwe-bio documentation
```
dblenkus, line 53: This should not be resolwe-bio specific.

docs/example/example/processes/template_py_process.py
```diff
@@ -0,0 +1,277 @@
...
+    # https://resolwe-bio.readthedocs.io/en/latest/catalog-category.html
+    category = "Read Processing"
+    # Dynamically generated name of the resulting datum object. Lookup
```
dblenkus, line 56: I think it is more common to name the object "Data object", at least in Resolwe.

docs/example/example/processes/template_py_process.py
```diff
@@ -0,0 +1,277 @@
...
+    # happens from Input class fields.
+    data_name = '{{ string_field.file|default("?") }}'
+    scheduling_class = SchedulingClass.BATCH  # or .INTERACTIVE for short jobs
+    # or RAW or TMP, see documentation at
+    # https://resolwe.readthedocs.io/en/latest/proc.html#persistence
+    persistence = Persistence.CACHED
+    # Use below entity to make the created object also a sample.
```
dblenkus, line 63: sample -> entity of type sample

docs/example/example/processes/template_py_process.py
```diff
@@ -0,0 +1,277 @@
...
+        "always_create": True,
+    }
+    requirements = {
+        "expression-engine": "jinja",
+        "executor": {
+            "docker": {
+                # Refer to a valid Docker image. If using process
+                # within resolwe-bio, note that explicit version number
+                # as 'latest' is not permitted.
+                "image": "resolwebio/common:1.3.1"
```
dblenkus, line 81: Don't reference resolwe-bio, use `resolwe/base:fedora-31` or similar.

docs/example/example/processes/template_py_process.py
```diff
@@ -0,0 +1,277 @@
...
+            "memory": 16384,
+            # Connects container to a network.
+            "network": True,
+        },
+    }
+
+    class Input:
+        """Input fields to process MergeData."""
+
+        string_field = StringField(
```
dblenkus, line 95: There is a lot of other options, e.g. `allow_custom_choice`, `custom_choice_conditions`, `validate_regex`,...

docs/example/example/processes/template_py_process.py
```diff
@@ -0,0 +1,277 @@
...
+    def run(self, inputs, outputs):
+        """Enter your not so pseudo code below
+
+        Inputs are available through `inputs` variable. I.e., using
+        `inputs.list_fields` should yield a list.
+
+        """
+
+        # Below code extracts file name without extension.
+        # It below assumes 'seq:nucleotide' datatype. See individual
```
dblenkus, line 214: Use a generic process type.

docs/example/example/processes/template_py_process.py
```diff
@@ -0,0 +1,277 @@
...
+        # Possible values for various fields.
+        inputs.string_field  # string ('computer_readable1')
+        inputs.text_field  # string ('default text')
+        inputs.boolean_field  # boolean True or False
+        inputs.integer_field  # integer scalar (1)
+        inputs.float_field  # float scalar (3.14)
+        inputs.date_field  # `yyy-mm-dd` (2020-04-20)
+        inputs.datetime_field  # `yyy-mm-dd hh:mm:ss` (2020-04-20 12:16:00)
+        inputs.secret_field  # TODO: secret model
+        inputs.file_field  # string (relative) '/path/to/file'
```
dblenkus, line 229: If the path starts with a slash, it is not relative.

docs/example/example/processes/template_py_process.py
```diff
@@ -0,0 +1,277 @@
...
+        # Report errors, warning or info using:
+        self.error(f"Oops, an error occurred in {inputs.data_field.id}")
+        self.warning(f"Warning in {inputs.data_field.id}")
+        self.info(f"Pass info about {self.version}")
+
+        # You can move the progress bar using a float on scale 0.0 to 1.0.
+        self.progress(0.5)
+
+        # Specify all fields in Output class to `outputs`.
+        outputs.file_out = "./path/to/filename.fastq"
```
dblenkus, line 267: Use a generic file extension.

docs/example/example/processes/template_py_process.py
romunov: How does this sound?

```
data_name = 'Data object ({{ string_field.file|default("?") }})'
```

docs/example/example/processes/template_py_process.py
romunov: How does this sound?

```
# Below code extracts file name and omits .txt extension.
# Assuming Input has a variable called file_field, which in
# turn has field raw_file which holds a path to the file.
basename = Path(inputs.file_field.raw_file.path).name
assert basename.endswith(".txt")
name = basename[:-4]
```

docs/example/example/processes/template_py_process.py
romunov: Listing all of them for every Field would become rather verbose, especially for fields that are not used often. This would perhaps be better suited to be listed in each Field's documentation page.

I have added a few new arguments, like `placeholder`, `range` and `validate_regex` to a few entries.

dblenkus: Not listing all options for each field, but it would be nice to make one (or two) fields more verbose and list all available options.

docs/example/example/processes/template_py_process.py
```diff
@@ -0,0 +1,285 @@
...
+        "descriptor_schema": "descriptor-schema-of-entity",
+        "input": "name of input that has the entity",
+        # If you want to create entity in all cases
+        "always_create": True,
+    }
+    requirements = {
+        "expression-engine": "jinja",
+        "executor": {
+            "docker": {
+                # Refer to a valid Docker image. If using process
```
dblenkus, line 78: This is not resolwe-bio specific and depends on settings: https://github.com/genialis/resolwe/blob/640bda7f97c2afca1d2223dc038352e8364f9331/tests/settings.py#L147

docs/example/example/processes/template_py_process.py
```diff
@@ -0,0 +1,285 @@
...
+            description="Description ends in a period.",
+            choices=[
+                ("computer_readable1", "Human readable 1"),
+                ("computer_readable2", "Human readable 2"),
+            ],
+            default="computer_readable1",
+            required=False,
+            hidden=False,
+            allow_custom_choice=True,
+            custom_choice_conditions=[""],
```
dblenkus, line 106: Add an actual condition.

docs/example/example/processes/template_py_process.py
romunov: Comment made agnostic.

docs/example/example/processes/template_py_process.py
romunov: I believe `placeholder`, `range` and `validate_regex` do not work in Python processes yet so I've removed them. I have also made `string_field` more verbose as it includes all possible arguments.

docs/example/example/processes/template_py_process.py
romunov: Removed because this argument is not yet supported in Python process (I believe).

PR #702, opened 2020-04-28
Fix delete bug

gregorjerse:


resolwe/storage/models.py
```diff
@@ -202,6 +207,20 @@ def urls(self) -> List[str]:
...
+
+        The URLs are relative with respect to the connector base path.
+        """
+        return [os.path.join(self.url, file_) for file_ in self.urls]
+
+    def delete_data(self):
+        """Delete all data for this storage location."""
+        self.connector.delete(self.connector_urls)
+        # Remove base directory.
+        self.connector.delete([os.path.join(self.url, "")])
```
dblenkus, line 222: I would prefer to include this in `connector_urls`.

resolwe/storage/manager.py
```diff
@@ -32,17 +32,48 @@ def __init__(self, file_storage: FileStorage):
             ("data_slug", data_slug),
         ]
 
-    def _get_rules(self, connector_name: str, _type: str) -> dict:
+    def _get_matching_override_rule(
```
dblenkus, line 35: This function handles only `data_slug`, and `process_type` overrides, so there is no extra value in `self.override_priorities` as you need to update it for each new type.

I would recommend splitting this function in two, for each override type, and call them explicitly instead of using the `for` loop.

PR #706, opened 2020-05-04
Retry file transfer

gregorjerse:
when it fails. If transfer fails multiple times abort.

### DIFF
--- a/resolwe/storage/connectors/transfer.py
+++ b/resolwe/storage/connectors/transfer.py
@@ -6,11 +6,37 @@ from functools import partial
 from threading import Condition
 from typing import List, Optional
 
+import wrapt
+
 from .baseconnector import BaseStorageConnector
 from .exceptions import DataTransferError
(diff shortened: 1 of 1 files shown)


resolwe/storage/connectors/transfer.py
```diff
@@ -189,11 +215,15 @@ class Transfer:
...
-        self, from_connector: BaseStorageConnector, to_connector: BaseStorageConnector
+        self,
+        from_connector: BaseStorageConnector,
+        to_connector: BaseStorageConnector,
+        retry_count: int = 3,
     ):
         """Initialize transfer object."""
         self.from_connector = from_connector
         self.to_connector = to_connector
+        self._retry_count = retry_count
```
dblenkus, line 226: This variable looks obsolete

gregorjerse: True. I have removed it from the source.

PR #711, opened 2020-05-12
 Add move_to_collection method to Data viewset

hadalin:


resolwe/flow/models/data.py
```diff
@@ -230,6 +230,17 @@ def duplicate(
...
 
+    @transaction.atomic
+    def move_to_collection(self, destination_collection):
+        """Move data objects to destination collection."""
+        for datum in self:
+            datum.validate_collection(destination_collection)
+            datum.collection = destination_collection
+            if destination_collection:
+                datum.tags = destination_collection.tags
+                copy_permissions(destination_collection, datum)
```
hadalin, line 241: I've taken this from `Entity` model but noticed that if you change data collection by updating collection on data directly via 

```
request = {"collection": {"id": self.collection.pk}}
data_detail_viewset(request, pk=data.pk)
```

then I guess tags and permissions aren't copied. Should tags and permissions also be copied in the latter case?

JureZmrzlikar: I strongly vote for yes. As an direct restAPI user i would like that such intricacies are solved "for free". I guess this also removes some of the burden from frontend, right? Another use case is ReSDK, i'm sure users will be happy if they are not required to copy collection tags and permissions manually. 
Drawback is of course the lost of general ability to do one isolated "move". But do we need it anyway? Is there such a use case?

resolwe/flow/tests/test_api.py
```diff
@@ -367,6 +370,60 @@ def test_change_collection(self):
         data.refresh_from_db()
         self.assertEqual(data.collection, self.collection)
 
+    def test_move_to_collection(self):
```
dblenkus, line 373: Add a test to verify that the user is not able to move the data to a collection where he doesn't have permissions.

resolwe/flow/views/data.py
```diff
@@ -164,3 +165,34 @@ def parents(self, request, pk=None):
...
+    def _get_data(self, user, ids):
+        """Return data objects queryset based on provided ids."""
+        queryset = get_objects_for_user(
+            user, "view_data", Data.objects.filter(id__in=ids)
+        )
+        actual_ids = queryset.values_list("id", flat=True)
+        missing_ids = list(set(ids) - set(actual_ids))
+        if missing_ids:
+            raise exceptions.ParseError(
+                "Entities with the following ids not found: {}".format(
```
dblenkus, line 193: Entities -> data

resolwe/flow/models/data.py
```diff
@@ -711,6 +722,17 @@ def get_runtime_path(self, filename=None):
             prefix=settings.FLOW_EXECUTOR["RUNTIME_DIR"], filename=filename
         )
 
+    def validate_collection(self, collection):
+        """Validate whether data object can change collection."""
+        if getattr(self, "entity", None):
+            if getattr(self.entity.collection, "id", None) != getattr(
```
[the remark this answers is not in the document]
hadalin: I've taken it from [here](https://github.com/genialis/resolwe/pull/711/files#diff-85dabc7dcb33a71b4346421a5d041b48L100).

resolwe/flow/models/data.py
hadalin: @dblenkus what's your take on this?

hadalin: Or we can leave as it is and use `move_to_collection` to when we want to copy tags and permissions. You'll just put a single data object in the input list.

resolwe/flow/models/data.py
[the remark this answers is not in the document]
hadalin: But you're right, we don't need it.

resolwe/flow/models/data.py
dblenkus: I think that it is necessary to keep collection, entity, and data in sync. So it would be great if you can fix that.
Regarding backward compatibility, I think that it is better to remove/change it than to leave an option to put things out of sync.

hadalin: Like [this](https://github.com/genialis/resolwe/pull/711/commits/5932e9a6deda3bc1371baea42867312045ce52bc)?

resolwe/flow/models/data.py
```diff
@@ -726,8 +721,20 @@ def get_runtime_path(self, filename=None):
...
+        self.validate_change_collection(destination_collection)
+        self.collection = destination_collection
+        if destination_collection:
+            self.tags = destination_collection.tags
+            copy_permissions(destination_collection, self)
+        self.save()
+
+    def validate_change_collection(self, destination_collection):
         """Raise validation error if data object cannot change collection."""
+        if self.collection == destination_collection:
```
dblenkus, line 735: This should not be necessary as the method should be triggered only for an actual change.

resolwe/flow/serializers/data.py
```diff
@@ -97,5 +97,12 @@ def validate_process(self, process):
     def validate_collection(self, collection):
         """Verify that changing collection is done in the right place."""
         if self.instance:
-            self.instance.validate_change_collection()
+            self.instance.validate_change_collection(collection)
```
dblenkus, line 100: `if self.instance and self.instance.collection != collection:`

resolwe/flow/serializers/data.py
```diff
@@ -97,5 +97,12 @@ def validate_process(self, process):
...
         return collection
+
+    def update(self, instance, validated_data):
+        """Update."""
+        instance = super().update(instance, validated_data)
+        instance.move_to_collection(instance.collection)
```
dblenkus, line 106: This triggers a save each time it is called, so you should put it in an if:
```
if instance.collection != validated_data['collection']:
```
But to be able to do that, you will have to copy a few parts of the original `update` method instead of calling `super`.

resolwe/flow/tests/test_api.py
```diff
@@ -367,6 +370,89 @@ def test_change_collection(self):
...
+        force_authenticate(request, self.contributor)
+        response = self.move_to_collection_viewset(request)
+
+        self.assertEqual(
+            response.data["detail"],
+            "You do not have permission to perform this action.",
+        )
+
+        # It's possible to move data if source collection does
+        # not have add and edit permissions.
```
dblenkus, line 455: Remove "add". It should not be possible to move the data if you don't have edit permission on both collections.

resolwe/flow/serializers/data.py
hadalin: Why calling original `update`? Doesn't `validated_data` hold the previous collection?

dblenkus: No, `validated_data` holds the new state and `super` returns already updated object. So you have to get the access to [`instance`](https://github.com/encode/django-rest-framework/blob/aed74961ba03e3e6f53c468353f4e255eb788555/rest_framework/mixins.py#L65) in the original method. 
Not calling super will also save one save and thus improve performance.

hadalin: Oh, ok. Why not then just put pre-update collection in a variable `[pre_update_collection = instance.collection`](https://github.com/genialis/resolwe/pull/711/files#diff-85dabc7dcb33a71b4346421a5d041b48R105-R108).

dblenkus: Where will you get the `instance`? You have to duplicate at least one line from the original function, so it is nicer if you copy all five lines as it is that short.

dblenkus: Ok, I haven't noticed that it is passed in as an argument :)

PR #712, opened 2020-05-13
Report registration failure in ProcessTestCase

JureZmrzlikar:


resolwe/test/testcases/process.py
```diff
@@ -253,7 +253,10 @@ def setUp(self):
         """Initialize test data."""
         super().setUp()
 
-        self._register_schemas()
+        _, stderr = self._register_schemas()
+        stderr = stderr.getvalue()
+        if stderr:
+            raise ValueError(stderr)
```
dblenkus, line 259: This is not `ValueError`, more `RuntimeError`. Otherwise, it looks good.

dblenkus:
What about a change log?

resolwe/test/testcases/process.py
JureZmrzlikar: Fixed.

JureZmrzlikar:
> What about a change log?

Just added. ;-0

JureZmrzlikar:
Rebased, tests pass. Please merge.

PR #715, opened 2020-05-27
Log executor exceptions

gregorjerse:


resolwe/flow/executors/__main__.py
```diff
@@ -56,17 +83,24 @@ async def run_executor():
 
 
 if __name__ == "__main__":
+    # initialize logging ASAP.
```
dblenkus, line 86: Capitalize "Initialize".

resolwe/flow/executors/__main__.py
```diff
@@ -28,14 +28,41 @@
...
 from .logger import configure_logging
-from .protocol import ExecutorFiles
-from .transfer import transfer_data
+from .protocol import ExecutorProtocol
+
+
+def handle_exception(exc_type, exc_value, exc_traceback):
+    """Log unhandled exceptions."""
+    message = "".join(traceback.format_exception(exc_type, exc_value, exc_traceback))
+    logger.error("Unhandled exception in executor: {}".format(message),)
```
dblenkus, line 45: Would it make sense to put a new line after colon?

gregorjerse: Yes, and I have removed colon too.

resolwe/flow/executors/__main__.py
gregorjerse: OK.

PR #716, opened 2020-06-01
Make code Python3.6 compatible and add verify data method

gregorjerse:


resolwe/storage/models.py
```diff
@@ -243,6 +246,44 @@ def transfer_data(self, destination: "StorageLocation"):
...
+                        )
+                    )
+                    continue
+
+                # Hashes differ: log error and continue to check other hashes.
+                if referenced_path != hash_value:
+                    return_value = False
+                    logger.error(
+                        "ReferencedPath with id {} has wrong {} hash: {} instead of {}".format(
+                            referenced_path.id,
```
dblenkus, line 278: The string has 4 placeholders and 5 values to insert.

resolwe/storage/models.py
```diff
@@ -243,6 +246,44 @@ def transfer_data(self, destination: "StorageLocation"):
...
+        and cause network traffic, use with care.
+
+        :returns: True if data is OK, False otherwise. In case of failure
+        additional information about error is logged.
+        """
+        hash_types = ["md5", "crc32c", "awss3etag"]
+        return_value = True
+        for referenced_path in self.files:
+            url = self.get_path(filename=referenced_path.path)
+            hashes = dict(zip(hash_types, self.connector.get_hashes(url, hash_types)))
```
dblenkus, line 262: Is there any use-case for `get_hashes` where you don't need such dict? If not, the method can return it.

resolwe/storage/models.py
gregorjerse: Fixed.

resolwe/storage/models.py
gregorjerse: Not with new `transfer_objects` method, except one migration. I have changed method signature and the corresponding code.

resolwe/storage/connectors/baseconnector.py
```diff
@@ -135,6 +127,15 @@ def after_push(self, objects: List[dict], url: str) -> Optional[List[dict]]:
         """
         return objects
 
+    @validate_url
+    def check_url(self, url: Union[str, PathLike]) -> bool:
```
dblenkus, line 131: Squash this commit with the previous one.

PR #718, opened 2020-06-09
Feature/storage

gregorjerse:


resolwe/storage/connectors/transfer.py
```diff
@@ -116,14 +121,11 @@ def transfer_objects(
...
-            for i in range(0, n):
-                yield iterable[i::n]
-
-        max_threads = 50
-        number_of_chunks = min(len(objects_to_transfer), max_threads)
-        objects_chunks = chunks(objects_to_transfer, number_of_chunks)
+        futures = paralelize(
+            objects=objects_to_transfer,
+            worker=partial(self._transfer_chunk, url=url),
+            max_threads=max_threads,
```
dblenkus, line 127: Where is `max_threads`? Have you accidentally removed it?

gregorjerse: It is optional method argument, defaults to 10.

dblenkus: I see it is a bit confusing that you've decreased it from 50 to 10 and thus removed it.

PR #721, opened 2020-06-11
Manager reuse storage location

gregorjerse:


resolwe/storage/manager.py
```diff
@@ -183,12 +183,12 @@ def _process_file_storage(self, file_storage: FileStorage):
...
                     storage_location=storage_location, reason="Manager data transfer"
                 )
-                new_storage_location = StorageLocation.objects.create(
+                new_storage_location = StorageLocation.all_objects.get_or_create(
                     file_storage=file_storage,
                     url=storage_location.url,
                     connector_name=connector_name,
                     status=StorageLocation.STATUS_UPLOADING,
-                )
+                )[0]
```
dblenkus, line 194: I like to use `new_storage_location, _ = ...` in such cases.

PR #723, opened 2020-06-16
Feature/storage

gregorjerse:


resolwe/storage/manager.py
```diff
@@ -83,6 +84,11 @@ def _should_copy_to(self, connector_name: str) -> bool:
 
         rule_results["has_rules"] = len(copy_rules) > 0
 
+        data = self.file_storage.data.first()
+        rule_results["data_object_ok"] = (
+            data is not None and data.status == Data.STATUS_DONE
```
dblenkus, line 89: This only checks status for the first object. I would change it to something like:

```
rule_results["data_object_ok"] = self.file_storage.data.filter(status=Data.STATUS_DONE).exists()

gregorjerse: OK, nice idea.

PR #727, opened 2020-06-17
Fix duplicated bug and batch upload

gregorjerse:
Otherwise exception is thrown.

### DIFF
--- a/resolwe/flow/executors/__main__.py
+++ b/resolwe/flow/executors/__main__.py
@@ -29,13 +29,16 @@ using the python from the ``venv`` virtualenv.
 import argparse
 import asyncio
 import logging
+import os
 import sys
 import traceback
 from importlib import import_module
+from pathlib import Path
 
 from . import manager_commands
-from .global_settings import DATA, DATA_META
+from .global_settings import DATA, DATA_META, EXECUTOR_SETTINGS
 from .logger import configure_logging
+from .manager_commands import send_manager_command
 from .protocol import ExecutorProtocol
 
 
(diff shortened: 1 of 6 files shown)


resolwe/storage/connectors/googleconnector.py
```diff
@@ -52,13 +52,16 @@ def __getattr__(self, name):
...
-            with self.client.batch():
-                for delete_url in urls:
-                    blob = self.bucket.blob(
-                        os.fspath(self.base_path / url / delete_url)
-                    )
-                    if blob.exists():
+        # At most 1000 objects can be deleted at the same time.
+        max_chunk_length = 1000
+        for i in range(0, len(urls), max_chunk_length):
+            with suppress(NotFound):
```
dblenkus, line 58: Is it ok to put `suppress` here? Will be the rest of 1000 objects when one fails? Would it make more sense to put it only around `delete`?

resolwe/flow/executors/__main__.py
```diff
@@ -98,9 +117,16 @@ async def run_executor():
...
+        try:
+            # Try to obtains exclusive lock over stdout and jsonout files.
+            # When lock can not be obtained this task is probably a duplicate
+            # spawned by celery so lock error and skip processing.
+            log_file, json_file = get_stdout_json_file()
+            await transfer_data()
+            await run_executor(log_file, json_file)
+            await collect_files()
+        except FileExistsError:
+            logger.error("Stdout or jsonout file already exists, aborting.")
```
dblenkus, line 129: Do you need to send `ABORT` command to the listener to keep the semaphore in sync?

resolwe/storage/connectors/googleconnector.py
gregorjerse: The documentation is not really clear but according to my experience all existing objects in the batch is deleted and the exception for last non-existing element in a batch is raised.

resolwe/flow/executors/__main__.py
gregorjerse: I believe not since executors has not even started at this point and no messages have been exchanged.

gregorjerse: I will take another look into that.

dblenkus: It is increased here: https://github.com/genialis/resolwe/blob/fcbf83ac008d5fc039f4e0e69c00fc2e4b439fc0/resolwe/flow/managers/dispatcher.py#L429

gregorjerse: Yes, it is. So to keep the counters in sync the abort has to be called.

PR #728, opened 2020-06-29
Add referenced files to storage location

gregorjerse:
when it is downloaded by the executor.

### DIFF
--- a/resolwe/flow/managers/listener.py
+++ b/resolwe/flow/managers/listener.py
@@ -251,7 +251,7 @@ class ExecutorListener:
             data_id = obj[ExecutorProtocol.DATA_ID]
             data = Data.objects.get(pk=data_id)
         except Data.DoesNotExist:
-            logger.warning(
+            logger.error(
                 "Data object does not exist (handle_get_referenced_files).",
                 extra={"data_id": data_id,},
             )
@@ -501,7 +501,14 @@ class ExecutorListener:
                 obj, {ExecutorProtocol.RESULT: ExecutorProtocol.RESULT_ERROR},
             )
         else:
-            query.update(status=StorageLocation.STATUS_DONE)
+            # Add files to the downloaded storage location.
+            storage_location = query.get()
+            default_storage_location = (
+                storage_location.file_storage.default_storage_location
+            )
+            storage_location.files.add(*default_storage_location.files.all())
+            storage_location.status = StorageLocation.STATUS_DONE
+            storage_location.save()
             async_to_sync(self._send_reply)(
                 obj, {ExecutorProtocol.RESULT: ExecutorProtocol.RESULT_OK},
             )
@@ -527,7 +534,7 @@ class ExecutorListener:
         try:
             data = Data.objects.get(pk=data_id)
         except Data.DoesNotExist:
-            logger.warning(
+            logger.error(
                 "Data object does not exist (handle_missing_data_locations).",
                 extra={"data_id": data_id,},
             )
@@ -635,7 +642,6 @@ class ExecutorListener:
                     'command': 'storage_location_unlock',
                     'data_id': [id of the :class:`~resolwe.flow.models.Data`
                                object],
-                    'storage_location_id': id of storage location
                     'storage_access_log_id': id of access log to close.
                 }
         """
--- a/resolwe/storage/tests/test_listener.py
+++ b/resolwe/storage/tests/test_listener.py
@@ -1,0 +1,734 @@
+# pylint: disable=missing-docstring
+from pathlib import Path
+from unittest.mock import MagicMock, call, patch
+
+from resolwe.flow.managers.listener import ExecutorListener
+from resolwe.flow.managers.protocol import ExecutorProtocol
+from resolwe.flow.models import Data, DataDependency
+from resolwe.storage.models import (
+    AccessLog,
+    FileStorage,
+    ReferencedPath,
+    StorageLocation,
+)
+from resolwe.test import TestCase
+
+
+class ListenerTest(TestCase):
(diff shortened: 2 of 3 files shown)


resolwe/flow/managers/listener.py
```diff
@@ -501,7 +501,14 @@ def handle_download_finished(self, obj):
...
             )
         else:
-            query.update(status=StorageLocation.STATUS_DONE)
+            # Add files to the downloaded storage location.
+            storage_location = query.get()
+            default_storage_location = (
+                storage_location.file_storage.default_storage_location
+            )
+            storage_location.files.add(*default_storage_location.files.all())
+            storage_location.status = StorageLocation.STATUS_DONE
```
dblenkus, line 510: I think that `<m2m>.add` executes a query to add related objects, so you can keep `update` here instead of `save`.

resolwe/flow/managers/listener.py
```diff
@@ -501,7 +501,14 @@ def handle_download_finished(self, obj):
                 obj, {ExecutorProtocol.RESULT: ExecutorProtocol.RESULT_ERROR},
             )
         else:
-            query.update(status=StorageLocation.STATUS_DONE)
+            # Add files to the downloaded storage location.
+            storage_location = query.get()
+            default_storage_location = (
+                storage_location.file_storage.default_storage_location
```
dblenkus, line 507: The code (indentation) would be nicer if you rename `storage_location` to `location`, but also probably less consistent and self-explanatory.

resolwe/storage/tests/test_listener.py
```diff
@@ -0,0 +1,84 @@
...
+from resolwe.flow.managers.listener import ExecutorListener
+from resolwe.flow.managers.protocol import ExecutorProtocol
+from resolwe.storage.models import FileStorage, ReferencedPath, StorageLocation
+from resolwe.test import TestCase
+
+
+class ListenerTest(TestCase):
+    def setUp(self):
+        self.listener = ExecutorListener()
+        self.file_storage = FileStorage()
```
dblenkus, line 13: Add a new line above to separate infrastructure and data setup.

resolwe/storage/tests/test_listener.py
```diff
@@ -0,0 +1,84 @@
...
+from unittest.mock import MagicMock, patch
+
+from resolwe.flow.managers.listener import ExecutorListener
+from resolwe.flow.managers.protocol import ExecutorProtocol
+from resolwe.storage.models import FileStorage, ReferencedPath, StorageLocation
+from resolwe.test import TestCase
+
+
+class ListenerTest(TestCase):
+    def setUp(self):
```
dblenkus, line 11: Consider using [setUpTestData](https://docs.djangoproject.com/en/3.0/topics/testing/tools/#django.test.TestCase.setUpTestData) where relevant.

resolwe/storage/tests/test_listener.py
```diff
@@ -0,0 +1,84 @@
...
+                ):
+                    self.listener.handle_download_finished(obj)
+
+        async_to_sync.assert_called_once_with(send_reply)
+        send_wrapper.assert_called_once_with(
+            {"command": "download_finished", "data_id": -1, "storage_location_id": -2,},
+            {"result": "ER"},
+        )
+
+        storage_location.save()
```
dblenkus, line 53: Why do you need this save?

resolwe/storage/tests/test_listener.py
```diff
@@ -0,0 +1,84 @@
...
+        obj["storage_location_id"] = storage_location.id
+        send_wrapper = MagicMock()
```
dblenkus, line 55: You can use `<mock>. reset_mock` instead of creating a new mock.

resolwe/storage/tests/test_listener.py
```diff
@@ -0,0 +1,84 @@
...
+        )
+        obj = {
+            "command": ExecutorProtocol.DOWNLOAD_FINISHED,
+            "data_id": -1,
+            "storage_location_id": -2,
+        }
+        send_wrapper = MagicMock()
+        send_reply = MagicMock()
+        async_to_sync = MagicMock(return_value=send_wrapper)
+        with patch(
```
dblenkus, line 36: It would be more readable if you patch the whole function instead of using `with` block.

```
@patch(...)
def test(self, mock):
   ...
```

resolwe/storage/tests/test_listener.py
```diff
@@ -0,0 +1,84 @@
...
+        self.storage_location = StorageLocation(
+            file_storage=self.file_storage, connector_name="GCS", status="OK"
+        )
+        self.storage_location.save()
+        path = ReferencedPath.objects.create(
+            path="test.me", md5="md5", crc32c="crc", awss3etag="aws"
+        )
+        self.storage_location.files.add(path)
+
+    def test_download_finished(self):
```
dblenkus, line 24: This test is quite long and consists of two parts. Would it make sense to break it in half?

resolwe/storage/tests/test_listener.py
gregorjerse: Since it was actually not saved before. Now the code is a little cleaner in this regard.

resolwe/storage/tests/test_listener.py
gregorjerse: Done.

resolwe/storage/tests/test_listener.py
gregorjerse: OK.

resolwe/storage/tests/test_listener.py
gregorjerse: Done.

PR #730, opened 2020-07-07
Respect keep data settings in executor

gregorjerse:


resolwe/flow/tests/test_executors.py
```diff
@@ -359,3 +359,107 @@ def test_executor_fs_lock(self):
...
+        self.assertEqual(data.output["number"], 42)
+        self.assertEqual(data.status, Data.STATUS_DONE)
+        self.assertEqual(data.process_error, [])
+        # Make sure that the temporary file was not deleted.
+        self.assertTrue(tmp_file.exists())
+
+    @with_docker_executor
+    @tag_process("test-save-number")
+    @override_settings(FLOW_MANAGER_KEEP_DATA=False)
+    def test_executor_keep_data_false(self):
```
dblenkus, line 418: I think that this test is not necessary as this is the default behavior tested in the "purge" tests.

resolwe/flow/tests/test_executors.py
```diff
@@ -359,3 +359,107 @@ def test_executor_fs_lock(self):
         self.assertEqual(data.process_error, [])
         # Check that temporary file was not deleted.
         self.assertTrue(file.exists())
+
+    @with_docker_executor
+    @tag_process("test-save-number")
+    @override_settings(FLOW_MANAGER_KEEP_DATA=True)
+    def test_executor_keep_data_true(self):
```
dblenkus, line 366: It would be nicer to write a process that creates file and does not save it.

gregorjerse:
Duplicate.

PR #732, opened 2020-07-15
Relations in Python process

JureZmrzlikar:
@dblenkus , @gregorjerse This is heavily WIP, but it would be great if you could provide some general feedback regarding the approach i took.

### DIFF
--- a/resolwe/flow/models/utils.py
+++ b/resolwe/flow/models/utils.py
@@ -747,3 +747,63 @@ def referenced_files(data, include_descriptor=True):
     )
 
     return special_files + output_files + descriptor_files
+
+
+def get_collection_of_input_entities(data):
+    """Get collection that contains all "entity inputs" of a given data.
+
+    With "entity input", one refers to the inputs that are part of an entity.
+    """
+    # Prevent circular imports:
+    from resolwe.flow.models import Collection
+
+    data_ids = set()
+
+    for field_schema, fields in iterate_fields(data.input, data.process.input_schema):
+        name = field_schema["name"]
+        value = fields[name]
+        if "type" not in field_schema:
+            continue
+
+        if field_schema["type"].startswith("data:"):
+            value = [value]
+        elif not field_schema["type"].startswith("list:data:"):
+            continue
+
+        data_ids.update([val for val in value if val is not None])
+
+    collections = Collection.objects.filter(
+        data__in=list(data_ids), data__entity__isnull=False,
+    ).distinct()
+
+    if collections.count() != 1:
+        raise ValueError(
+            "Entity inputs should be part of exactly one collection. (not {})".format(
+                len(collections)
+            )
+        )
+
+    return collections.first()
+
+
+def serialize_collection_relations(collection):
+    """Serialize all relations of a collection."""
+    # Prevent circular imports:
+    from resolwe.flow.models import Relation, RelationPartition
+
+    serialized_relations = []
+    for relation in Relation.objects.filter(collection=collection):
+        serialized_relations.append(
+            {
+                "relation_id": relation.id,
+                "relation_type_name": relation.type.name,
+                "relation_type_ordered": relation.type.ordered,
+                "category": relation.category,
+                "unit": relation.unit,
+                "partitions": list(
+                    RelationPartition.objects.filter(relation=relation).values()
+                ),
+            }
+        )
+
+    return serialized_relations
--- a/resolwe/flow/models/utils.py
+++ b/resolwe/flow/models/utils.py
@@ -362,6 +362,7 @@ def hydrate_input_references(input_, input_schema, hydrate_values=True):
                 output["__type"] = data.process.type
                 output["__descriptor"] = data.descriptor
                 output["__name"] = getattr(data, "name", None)
+                output["__entity_id"] = getattr(data.entity, "id", None)
                 output["__entity_name"] = getattr(data.entity, "name", None)
                 output["__output_schema"] = data.process.output_schema
 
@@ -388,6 +389,7 @@ def hydrate_input_references(input_, input_schema, hydrate_values=True):
                     output["__type"] = data.process.type
                     output["__descriptor"] = data.descriptor
                     output["__name"] = getattr(data, "name", None)
+                    output["__entity_id"] = getattr(data.entity, "id", None)
                     output["__entity_name"] = getattr(data.entity, "name", None)
                     output["__output_schema"] = data.process.output_schema
 
--- a/resolwe/process/__main__.py
+++ b/resolwe/process/__main__.py
@@ -22,6 +22,7 @@ if __name__ == "__main__":
     parser.add_argument(
         "--requirements", type=str, help="Path to requirements JSON file"
     )
+    parser.add_argument("--relations", type=str, help="Path to relations JSON file")
     args = parser.parse_args()
 
     # Switch to target directory to import the module.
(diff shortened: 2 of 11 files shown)


JureZmrzlikar:
@dblenkus , @gregorjerse IMO code is ready for review. There are still some small things to be done, but the majority of it is done.

resolwe/flow/models/utils.py
```diff
@@ -747,3 +747,72 @@ def referenced_files(data, include_descriptor=True):
...
+        if field_schema["type"].startswith("data:"):
+            if value is None:
+                continue
+            try:
+                data = Data.objects.get(id=value)
+                if data.entity and data.collection:
+                    collections.add(data.collection.id)
+            except Data.DoesNotExist:
+                # This exception is copy-pasted from hydrate_input_references
+                # TODO: Think what is the case when data cannot be found? How process
```
gregorjerse, line 776: I believe this could happen:
- job is created, all input data is there
- job is waiting in queue, somebody deletes input data for the job
- job is started and input data is missing. 

If this is the case then aborting the job would be the right thing to do.

resolwe/flow/models/utils.py
```diff
@@ -747,3 +747,72 @@ def referenced_files(data, include_descriptor=True):
...
+                try:
+                    data = Data.objects.get(id=val)
+                    if data.entity and data.collection:
+                        collections.add(data.collection.id)
+                except Data.DoesNotExist:
+                    continue
+
+    if len(collections) != 1:
+        raise ValueError(
+            f"Entity inputs should be part of one and only one collection. (not {len(collections)})"
```
gregorjerse, line 793: Maybe "exactly one" instead of "one and only one"?

resolwe/flow/models/utils.py
JureZmrzlikar: But this should than also be fixed in `hydrate_input_references`, right?

JureZmrzlikar: Just to write this down: After some discussion, we agreed to just let this be as is. Will fix in the future if proven necessary.

resolwe/flow/models/utils.py
```diff
@@ -747,3 +747,68 @@ def referenced_files(data, include_descriptor=True):
...
+        if field_schema["type"].startswith("data:"):
+            if value is None:
+                continue
+            try:
+                data = Data.objects.get(id=value)
+                if data.entity and data.collection:
+                    collections.add(data.collection.id)
+            except Data.DoesNotExist:
+                continue
+        elif field_schema["type"].startswith("list:data:"):
```
dblenkus, line 776: Unifi both cases, e.g.
```python
if field_schema["type"].startswith("data:"):
    value = [value]

for val in value:
   ...

File: resolwe/flow/models/utils.py [hunk 4]
```diff
@@ -747,3 +747,68 @@ def referenced_files(data, include_descriptor=True):
...
+                try:
+                    data = Data.objects.get(id=val)
+                    if data.entity and data.collection:
+                        collections.add(data.collection.id)
+                except Data.DoesNotExist:
+                    continue
+
+    if len(collections) != 1:
+        raise ValueError(
+            f"Entity inputs should be part of exactly one collection. (not {len(collections)})"
```
dblenkus, line 789: We don't use f-strings in the codebase, so use `format` instead to keep it consistent.

resolwe/flow/models/utils.py
```diff
@@ -747,3 +747,68 @@ def referenced_files(data, include_descriptor=True):
...
+    for field_schema, fields in iterate_fields(data.input, data.process.input_schema):
+        name = field_schema["name"]
+        value = fields[name]
+        if "type" not in field_schema:
+            continue
+        if field_schema["type"].startswith("data:"):
+            if value is None:
+                continue
+            try:
+                data = Data.objects.get(id=value)
```
dblenkus, line 771: This is inefficient as a separate db query is made for each input and then an additional request is made for `data.collection`. You just need to collect data ids here and use them in a single request at the end of the function:
```python
collections = Collection.objects.filter(data__in=[data_ids])
if collections.count() != 1:
    raise

return collections.first()
```

resolwe/process/__main__.py
```diff
@@ -95,13 +96,24 @@
         print("ERROR: Requirements failed to load: {}".format(str(error)))
         sys.exit(1)
 
+    # Prepare relations.
+    relations = None
+    try:
+        if args.relations:
```
dblenkus, line 102: If this `if` needed as the parameter is required?

resolwe/process/fields.py
```diff
@@ -546,6 +595,10 @@ def _populate_cache(self):
 
     def _get(self, key):
         """Return given key from cache."""
+        # Relations are not stored in self._cache
```
dblenkus, line 598: Is there a reason why they are not stored in cache? I think it would be sensible as `self.relations` is not a trivial function.

resolwe/process/fields.py
```diff
@@ -582,6 +635,18 @@ def entity_name(self):
...
+    @property
+    def relations(self):
+        """Get this data's relations."""
+        relations = set()
+        for relation in [
+            RelationDescriptor.from_dict(data) for data in self._field.process.relations
+        ]:
+            for partition in relation.partitions:
+                if partition.entity_id == self.entity_id:
+                    relations.add(relation)
```
dblenkus, line 647: It would make sense to initialize `RelationDescriptor` here, not in the `for` loop.

resolwe/flow/models/utils.py
JureZmrzlikar: Good suggestion, done.

resolwe/flow/models/utils.py
JureZmrzlikar: Not totally true, since we check that for each data it also holds that `data.entity != None`. But you're right that this can be optimized. Will do.

resolwe/process/__main__.py
JureZmrzlikar: Currently, this parameter is always given, but it is not considered as required - AFAIK argparse considers parameters starting with `--` as optional and the one without `--` as mandatory.
[Link do related argparse docs.](https://docs.python.org/3/library/argparse.html#name-or-flags)
For safety (and consistency with `--requirements`) i think it is better to leave it as is.

resolwe/process/fields.py
JureZmrzlikar: True, will fix.

resolwe/process/fields.py
JureZmrzlikar: Fixed.

resolwe/process/__main__.py
dblenkus: Interesting, I didn't know that.

resolwe/process/fields.py
```diff
@@ -582,6 +637,20 @@ def entity_name(self):
...
+        return self._get("__relations")
+
+    def get_relations(self):
+        """Get this data's relations."""
+        relations = set()
+        for relation in self._field.process.relations:
+            for partition in relation["partitions"]:
+                if partition["entity_id"] == self.entity_id:
+                    relations.add(RelationDescriptor.from_dict(relation))
+        return list(relations)
```
dblenkus, line 652: How does Python compare `RelationDescriptors`? Do you have to add `__eq__` method to work as expected? I imagine that otherwise, they are compared by `id`, and even equal relations are treated as different.

JureZmrzlikar: Good catch! Yes, one needs to add `__eq__` and also `__hash__` methods for objects to be usable in sets. I've added `relation.id` to the "serializer" and used it in the above methods.

PR #733, opened 2020-07-21
Add relations to data inputs

dblenkus:


resolwe/flow/static/flow/fieldSchema.json
```diff
@@ -249,6 +250,53 @@
...
+                    "required": ["type", "npartitions"],
+                    "additionalProperties": false,
+                    "properties": {
+                        "type": {
+                            "type": "string"
+                        },
+                        "npartitions": {
+                            "anyOf": [
+                                {"type": "integer"},
+                                {"type": "string", "patteren": "^none|any$"}
```
mstajdohar, line 292: I propose we initially support defining a number 1 and not defining the property at all (not required).

resolwe/flow/static/flow/fieldSchema.json
```diff
@@ -448,6 +495,53 @@
...
```
mstajdohar, line 539: Change accordingly with what we decide above.

PR #736, opened 2020-08-03
Fix schema generation for list fields in Python processes

dblenkus:


dblenkus:
I've added a test, but I think the changelog is not needed, as it already includes this under "Add relations property to data: and list:data: fields to support relations on client".

PR #740, opened 2020-08-18
Improve process termination

dblenkus:
Improvments include:
* wait for (and kill) the process only if it was not yet terminated
* don't wait for process in Docker executor as it is already handled in
  the parent function
* don't log error when process is already terminated
* log error if Docker container removal fails

### DIFF
--- a/resolwe/flow/executors/local/run.py
+++ b/resolwe/flow/executors/local/run.py
@@ -4,6 +4,7 @@
 
 """
 import asyncio
+import contextlib
 import logging
 import os
 import shlex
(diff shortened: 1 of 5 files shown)


resolwe/flow/executors/local/run.py
```diff
@@ -61,12 +61,12 @@ async def terminate(self):
...
-            logger.exception("While terminating process with PID %s", self.proc.pid)
-
-        await asyncio.wait_for(self.proc.wait(), self.kill_delay)
-        if self.proc.returncode is None:
-            self.proc.kill()
-        await self.proc.wait()
+            # Process has already been terminated.
+            pass
+        else:
+            await asyncio.wait_for(self.proc.wait(), self.kill_delay)
```
gregorjerse, line 67: This line will raise asyncio.TimeoutError exception if timeout occured.
I suggest using suppress.

dblenkus: Good point! Fixed.

PR #741, opened 2020-08-21
Lock input locations early and unlock after processing is done

gregorjerse:


resolwe/flow/managers/dispatcher.py
```diff
@@ -836,6 +841,35 @@ async def communicate(self, data_id=None, run_sync=False, save_settings=True):
...
+                children_dependency__child=data,
+                children_dependency__kind=DataDependency.KIND_IO,
+            )
+            .exclude(location__isnull=True)
+            .distinct()
+        )
+
+        for parent in dependencies:
+            file_storage = parent.location
+            if file_storage.has_storage_location(STORAGE_LOCAL_CONNECTOR):
```
dblenkus, line 862: This makes 2 queries for each Data object, one to check if it exists and one to retrieve it. It would be better if you do this by filtering the queryset. Then you can also create all locks at once with `bulk_create`.

resolwe/flow/managers/listener.py
```diff
@@ -509,6 +509,16 @@ def handle_download_finished(self, obj):
             storage_location.files.add(*default_storage_location.files.all())
             storage_location.status = StorageLocation.STATUS_DONE
             storage_location.save()
+            # Add lock to it since it is required for processing. Do not change
+            # the reason since it is used to unlock storage locations after
```
dblenkus, line 529: "since it is used to unlock" - is this true or is it unlocked based on the cause?

resolwe/flow/managers/listener.py
```diff
@@ -231,6 +231,22 @@ def hydrate_spawned_files(self, exported_files_mapper, filename, data_id):
...
+    def unlock_all_inputs(self, data_id):
+        """Unlock all data objects that were locked by the given data.
+        
+        If exception occurs during unlocking we can not do much but ignore it.
+        """
+        with suppress(Exception):
+            query = AccessLog.objects.filter(cause_id=data_id)
+            if not query.exists():
+                # Log error and continue
+                logger.error(
```
dblenkus, line 243: Why is this an error? It is perfectly fine to have a Data object with no inputs.

resolwe/flow/managers/listener.py
```diff
@@ -683,6 +699,9 @@ def handle_annotate(self, obj):
         """
 
         def report_failure():
+            # Unlock all inputs.
```
dblenkus, line 702: This comment is probably not needed.

resolwe/flow/tests/test_manager.py
```diff
@@ -124,6 +125,14 @@ def test_dependencies(self):
         data_child1.refresh_from_db()
         data_child2.refresh_from_db()
         data_child3.refresh_from_db()
+
+        # Check locks are created in manager.
```
dblenkus, line 129: Tests should be in the same commit as the code that they test. Would it make sense to check the state of the access log, i.e. if it is locked or not?

resolwe/flow/managers/dispatcher.py
```diff
@@ -836,6 +841,35 @@ async def communicate(self, data_id=None, run_sync=False, save_settings=True):
...
+                # Lock the local storage location.
+                local_storage_location = file_storage.storage_locations.get(
+                    connector_name=STORAGE_LOCAL_CONNECTOR
+                )
+                AccessLog.objects.create(
```
dblenkus, line 867: Why do you only lock local storage locations? I think it would make sense to lock the local one if it exists and the default one otherwise. This would guarantee that the new Data object will have all inputs available.

dblenkus (approved):

PR #742, opened 2020-08-28
Faster duplicate

dblenkus:


resolwe/flow/tests/test_models.py
```diff
@@ -125,10 +125,14 @@ def test_name(self):
...
-            Data.objects.create(contributor=self.contributor, process=process)
+        Data.objects.bulk_create(
+            [
+                Data(name=str(i), contributor=self.contributor, process=process, size=0)
+                for i in range(25)
+            ]
+        )
 
-        Data.objects.all().delete_chunked(chunk_size=50)
+        Data.objects.all().delete_chunked(chunk_size=10)
```
gregorjerse, line 135: Just being curious: why change chunk size to 10?

resolwe/flow/models/utils/duplicate.py
```diff
@@ -0,0 +1,301 @@
...
+    for permission_cls in [GroupObjectPermission, UserObjectPermission]:
+
+        permissions_qs = permission_cls.objects.filter(
+            content_type=content_type, object_pk__in=pk_mapping.keys()
+        )
+        new_permissions = []
+
+        for permission in permissions_qs:
+            permission.pk = None
+            permission.object_pk = pk_mapping[int(permission.object_pk)]
```
gregorjerse, line 33: I believe this should be converted to string? Or is pk_mapping a map between int and string?

resolwe/flow/models/utils/duplicate.py
```diff
@@ -0,0 +1,301 @@
...
+    UserObjectPermission.objects.bulk_create(new_permissions, ignore_conflicts=True)
+
+
+def _rewire_foreign_key(objects, target, pk_mapping):
+    """"""
+    if not objects:
+        return
+
+    # Check that all objects are instances of the same class.
+    obj_class = {obj.__class__ for obj in objects}
```
gregorjerse, line 61: Maybe this is more readable:
```
obj_class = objects[0].__class__
if any(obj.__class__ != obj_class for obj in objects[1:]):
  raise ValueError("All objects must be instances of the same class.")
```

resolwe/flow/models/utils/duplicate.py
```diff
@@ -0,0 +1,301 @@
...
+    if len(obj_class) > 1:
+        print(obj_class)
+        raise ValueError("All objects must be instances of the same class.")
+    obj_class = obj_class.pop()
+
+    # Find fields pointing to the target model.
+    to_map = []
```
gregorjerse, line 68: This could be written as a list comprehension, but then it is quite long and (debatable) less maintanable.

resolwe/flow/models/utils/duplicate.py
```diff
@@ -0,0 +1,301 @@
...
+            to_map.append(field.name)
+
+    # Rewire references.
+    new_objects = []
+    for obj in objects:
+        new_obj = deepcopy(obj)
+        for field_name in to_map:
+            field_name = "{}_id".format(field_name)
+            value = getattr(obj, field_name)
+            if not value or value not in pk_mapping:
```
gregorjerse, line 80: I would use 

```
if value and value in pk_mapping:
    setattr(...)  
```
Or just `if value in pk_mapping` since None is probably not contained in the mapping.

resolwe/flow/models/utils/duplicate.py
```diff
@@ -0,0 +1,301 @@
...
+        new_objects.append(new_obj)
+
+    try:
+        # Add another atomic block to avoid corupting the main one.
+        with transaction.atomic():
+            model.objects.bulk_create(new_objects)
+    except IntegrityError:
+        # Probably a slug collision occured, try to create objects one by one.
+        for obj in new_objects:
+            obj.slug = None
```
gregorjerse, line 176: This will probably lead to the same error as the code above?
Sooner or later the same slug clash will happen as it did above?

resolwe/flow/models/utils/duplicate.py
```diff
@@ -0,0 +1,301 @@
...
+        Data,
+        DataDependency,
+        DataMigrationHistory,
+        Entity,
+    )
+
+    name_prefix = name_prefix or "Copy of"
+
+    if contributor is None:
+        raise ValueError("Contributor must be give.")
```
gregorjerse, line 215: `must be given. `

Maybe you could just use `if contributor` syntax to make it consistent with the rest of the code.

resolwe/flow/models/utils/duplicate.py
```diff
@@ -0,0 +1,301 @@
...
+            entities, contributor, name_prefix, entity_processor
+        )
+
+        entity_mapping = {old.pk: new for old, new in zip(entities, new_entities)}
+
+        entity_data_qs = Data.objects.filter(entity__in=entities)
+        if data is None:
+            data = entity_data_qs
+        else:
+            data = data.union(entity_data_qs)
```
gregorjerse, line 249: One could define 

`data = Data.objects.none()`

near line 222 and always just use 

`data = data.union(entity_data_qs)`.

resolwe/flow/models/utils/duplicate.py
```diff
@@ -0,0 +1,301 @@
...
+
+        inherit_entity = True
+
+    if data is not None:
```
gregorjerse, line 253: Another strength of the above approach is that you could use 

`if data` here since data would always be a queryset.

resolwe/flow/models/utils/duplicate.py
```diff
@@ -0,0 +1,301 @@
...
+        if data.exclude(status__in=[Data.STATUS_DONE, Data.STATUS_ERROR]).exists():
+            raise ValidationError(
+                "Data object must have done or error status to be duplicated."
+            )
```
gregorjerse, line 257: Is this not annoying for the end user? He should wait for all objects in the collection to be processed before being able to duplicate it. 

How about just excluding data objects in processing from the duplication? But then again user will probably expect all objects to be duplicated.

resolwe/flow/models/utils/duplicate.py
```diff
@@ -0,0 +1,301 @@
...
+
+        data_processor = process_data(
+            inherit_collection, inherit_entity, collection_mapping, entity_mapping
+        )
+        new_data = copy_objects(data, contributor, name_prefix, data_processor)
+
+        data_pks = data.values('pk')
```
gregorjerse, line 264: Do you need values here or values_list with flat=True?

gregorjerse (approved):

resolwe/flow/models/utils/duplicate.py
dblenkus: This is automatically handled by Django. But you are right that it is nicer to explicitly cast it to a string.

resolwe/flow/models/utils/duplicate.py
dblenkus: Slug is generated based only on slugs already in the database. This means that the second conflicting slug will be changed if the first one is already written in the database.

resolwe/flow/models/utils/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+
+    collections = Collection.objects.filter(
+        data__in=list(data_ids),
+        data__entity__isnull=False,
+    ).distinct()
+
+    if collections.count() != 1:
+        raise ValueError(
+            "Entity inputs should be part of exactly one collection. (not {})".format(
+                len(collections)
```
gregorjerse, line 57: Maybe use collections.count() instead of length?

resolwe/flow/models/utils/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    validate_process_types,
+    validate_schema,
+    validation_schema,
+)
+
+
+def get_collection_of_input_entities(data):
+    """Get collection that contains all "entity inputs" of a given data.
+
+    With "entity input", one refers to the inputs that are part of an entity.
```
gregorjerse, line 29: I would add a paragraph about ValueError that is raise when inputs are part of more than one collection.

resolwe/flow/models/utils/duplicate.py
```diff
@@ -0,0 +1,482 @@
...
+    """Check that user has permission on all given objects.
+
+    :param objects: A list of objects to be checked.
+    :type objects: list of `~django.db.models.Model`
+
+    :param str permission: Codename of the required permission.
+
+    :param contributor: A Django user that should have the permission.
+    :type contributor: `~django.contrib.auth.models.User`
+
```
gregorjerse, line 24: Maybe add something similar to the line bellow to the documentation.

`:raises ValidationError when user does not have all the required permissions.`

resolwe/flow/models/utils/duplicate.py
```diff
@@ -0,0 +1,482 @@
...
+def bulk_duplicate(
+    collections=None,
+    entities=None,
+    data=None,
+    contributor=None,
+    inherit_collection=False,
+    inherit_entity=False,
+    name_prefix=None,
+):
+    """Make a copy of given collection, entity ir data queryset.
```
gregorjerse, line 335: ir -> or

resolwe/flow/models/utils/duplicate.py
```diff
@@ -0,0 +1,482 @@
...
+
+    Exactly one of ``collections``, ``enteties`` and ``data`` parameters should
```
gregorjerse, line 337: Enteties -> entities (change also in the documentation bellow this comment).

resolwe/flow/models/utils/duplicate.py
```diff
@@ -0,0 +1,482 @@
...
+    be passed to the function and should respectively represent a queryset of
+    Collection, Entity and Data objects to be copied.
+
+    When copying Collections or Enteties, also contained objects (Enteties and
```
gregorjerse, line 341: also contained objects -> the contained objects.

gregorjerse (approved):

PR #743, opened 2020-09-03
Fix bulk_duplicate function

dblenkus:
- Skip pre-processing of objects
- Handle case where collection or entity is not set

### DIFF
--- a/resolwe/flow/models/utils/duplicate.py
+++ b/resolwe/flow/models/utils/duplicate.py
@@ -195,7 +195,7 @@ def process_entity(inherit_collection, collection_mapping):
         :rtype: `~resolwe.flow.models.Entity`
 
         """
-        if inherit_collection:
+        if inherit_collection and entity.collection:
             if entity.collection.id in collection_mapping:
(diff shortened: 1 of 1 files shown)


gregorjerse (approved):

resolwe/flow/models/utils/duplicate.py
```diff
@@ -240,13 +240,13 @@ def processor(datum):
 
         """
         if inherit_collection:
```
gregorjerse, line 242: Maybe you could just rewrite the first condition to `if inherit_collection and datum.collection`?

resolwe/flow/models/utils/duplicate.py
```diff
@@ -240,13 +240,13 @@ def processor(datum):
...
-            if datum.collection.id in collection_mapping:
+            if datum.collection and datum.collection.id in collection_mapping:
                 datum.collection = collection_mapping[datum.collection.pk]
         else:
             datum.collection = None
 
         if inherit_entity:
```
gregorjerse, line 248: Maybe you could just rewrite the first condition to `if inherit_entity and entity.collection`?

PR #744, opened 2020-09-10
Validate referenced input Data objects only when object is created

gregorjerse:


resolwe/flow/models/data.py
```diff
@@ -544,6 +544,11 @@ def save(self, render_name=False, *args, **kwargs):
                 self.input, self.process.slug, self.process.version
             )
 
+            validate_schema(
```
dblenkus, line 547: Why is this not in a single line?

resolwe/flow/managers/listener.py
```diff
@@ -373,7 +373,8 @@ def handle_referenced_files(self, obj):
                 if data.status == Data.STATUS_PROCESSING:
                     data.status = Data.STATUS_DONE
                     data.save()
-                    validate_data_object(data)
+                    # Do not raise error if input data object is missing.
```
dblenkus, line 376: Mention that we can skip missing data as the lock guarantee that the storage location is still there.

resolwe/flow/models/data.py
gregorjerse: Most probably my fault since I forgot to remove the last comma and let black handle the formatting.

resolwe/flow/managers/listener.py
gregorjerse: Done.

dblenkus (approved):

PR #746, opened 2020-09-18
Make relations visible to public user

gregorjerse:
if he has view permission on the respective collection.

### DIFF
--- a/resolwe/flow/tests/test_manager.py
+++ b/resolwe/flow/tests/test_manager.py
@@ -1,7 +1,10 @@
 # pylint: disable=missing-docstring
 import os
+from pathlib import Path
 
 from asgiref.sync import async_to_sync
+
+from django.test import override_settings
 
 from guardian.shortcuts import assign_perm
 
@@ -95,6 +98,17 @@ class TestManager(ProcessTestCase):
             {d.kind for d in data2.children_dependency.all()},
             {DataDependency.KIND_SUBPROCESS},
         )
+
+    def test_communicate(self):
+        """Make sure communicate can serialize settings.
+
+        When object of PosixPath is used in settings the call to communicate
+        fails due to object not being JSON serializable by default.
+        """
+        with override_settings(FLOW_EXECUTOR={"DATA_DIR": Path("/some/path")}):
+            async_to_sync(manager.communicate)(run_sync=True)
+        # Make sure original settings are saved.
+        async_to_sync(manager.communicate)(run_sync=True)
 
     def test_dependencies(self):
         """Test that manager handles dependencies correctly."""
--- a/resolwe/flow/tests/test_relations.py
+++ b/resolwe/flow/tests/test_relations.py
@@ -2,6 +2,7 @@
 import os
 
 from django.apps import apps
+from django.contrib.auth.models import AnonymousUser
 from django.core.management import call_command
 from django.db import DEFAULT_DB_ALIAS, connections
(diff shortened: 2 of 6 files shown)


resolwe/flow/tests/test_manager.py
```diff
@@ -96,6 +99,17 @@ def test_workflow(self):
...
 
+    def test_communicate(self):
+        """Make sure communicate can serialize settings.
+
+        When object of PosixPath is used in settings the call to communicate
+        fails due to object not being JSON serializable by default.
+        """
+        with override_settings(FLOW_EXECUTOR={"DATA_DIR": Path("/some/path")}):
+            async_to_sync(manager.communicate)(run_sync=True)
+        # Make sure original settings are saved.
```
dblenkus, line 110: Are settings not saved by default in tests? If not, that can cause a lot of problems.

resolwe/flow/tests/test_relations.py
```diff
@@ -469,9 +513,15 @@ def test_delete_superuser(self):
...
+    def test_delete_publicuser(self):
+        """Anonymous user must not be able to delete relations."""
+        resp = self._delete(self.relation_group.pk, user=AnonymousUser())
+        self.assertEqual(resp.status_code, status.HTTP_401_UNAUTHORIZED)
+        self.assertTrue(Relation.objects.filter(pk=self.relation_group.pk).exists())
+
     def test_delete_different_user(self):
         resp = self._delete(self.relation_group.pk, user=self.user)
-        self.assertEqual(resp.status_code, status.HTTP_404_NOT_FOUND)
+        self.assertEqual(resp.status_code, status.HTTP_401_UNAUTHORIZED)
```
dblenkus, line 524: Does the user have the view permission on the object? If not, 404 should be returned.

resolwe/flow/tests/test_manager.py
gregorjerse: They are not saved by default when communicate is called from channels (like from listener, for example). So if original settings are not saved back (the last line), then all other tests fail.

resolwe/flow/tests/test_relations.py
gregorjerse: He does not have the permission. I will change the returned code.

But is it not more appropriate to return unauthorized in this case? It is true that more information is revealed (the existence of object, for example) but I believe it is nice to return unauthorized when someone does not have the permission to view the object.

resolwe/flow/views/relation.py
```diff
@@ -93,11 +93,18 @@ def update(self, request, *args, **kwargs):
         the collection referenced in the ``Relation``.
         """
         instance = self.get_object()
-        if (
-            not request.user.has_perm("edit_collection", instance.collection)
-            and not request.user.is_superuser
-        ):
-            return Response(status=status.HTTP_401_UNAUTHORIZED)
+
+        if not request.user.is_superuser:
```
dblenkus, line 97: Put this check into a separate method and reuse it in the function below, e.g. something like `self.has_change_permission(instance, request.user)`.

dblenkus (commented):
Looks good; I only have one comment. And don't forget to add a changelog.

resolwe/flow/tests/test_manager.py
dblenkus: It would be great to force-update settings for each test. But that can be done in a separate pull request.

dblenkus (approved):

PR #748, opened 2020-09-22
Fix non-required ListField in Python processes

JureZmrzlikar:
Until now, non-required ListField would fail if default=[] was
not provided.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -14,6 +14,9 @@ Added
 -----
 - Support accessing Data name in Python processes through ``self.name``
 
+Fixed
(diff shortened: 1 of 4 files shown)


dblenkus:
I am not sure if this is the right solution, as an empty list can be an actual value. It would be better if input/output is not given at all in that case and is thus skipped at validation.

JureZmrzlikar:
> I am not sure if this is the right solution, as an empty list can be an actual value
IMO, empty list is the only sensible default for a list field... Also, this makes it consistent with `.yaml` processes.

dblenkus (approved):

PR #756, opened 2020-10-30
Support Python processes in Sphinx autoprocess directive

JureZmrzlikar:
cc: @romunov

### DIFF
--- a/resolwe/flow/utils/docs/autoprocess.py
+++ b/resolwe/flow/utils/docs/autoprocess.py
@@ -38,6 +38,7 @@ from docutils.parsers.rst import Directive
 from sphinx import addnodes
 
 from resolwe.flow.utils import iterate_schema
+from resolwe.process.parser import SafeParser
 
 PROCESS_CACHE = None
 
@@ -77,7 +78,7 @@ def get_processes(process_dir, base_source_uri):
         return PROCESS_CACHE
 
     all_process_files = []
-    process_file_extensions = ["*.yaml", "*.yml"]
+    process_file_extensions = ["*.yaml", "*.yml", "*.py"]
     for root, _, filenames in os.walk(process_dir):
         for extension in process_file_extensions:
             for filename in fnmatch.filter(filenames, extension):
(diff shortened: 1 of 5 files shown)


resolwe/flow/utils/docs/autoprocess.py
```diff
@@ -53,10 +55,24 @@ def get_process_definition_start(fname, slug):
     :rtype: int
 
     """
-    with open(fname) as file_:
-        for i, line in enumerate(file_):
-            if re.search(r"slug:\s*{}".format(slug), line):
-                return i + 1
+    if fname.endswith(".py"):
+        with open(fname) as handle:
+            for node in ast.walk(ast.parse(handle.read())):
```
gregorjerse, line 60: This might return bogus value for any class that has slug attribute set to the correct value.
I know this is very improbable but maybe consider checking if class we are visiting is a subclass of Process.

See https://github.com/genialis/resolwe/blob/7ac6f69bb47b76fe5a1aff421209b97d1ac15720/resolwe/process/parser.py#L165 for details.

gregorjerse (approved):

resolwe/flow/utils/docs/autoprocess.py
gregorjerse: Maybe it would be better to traverse file only once instead of twice (second pass is just to find the starting line of the class definition). 

1. Modify ``ProcessDescriptor.Metadata`` class to include the ``lineno`` attribute (by default None).

2.  Add the line  ``descriptor.metadata.lineno = node.lineno`` in ``descriptor.py`` after line https://github.com/genialis/resolwe/blob/7ac6f69bb47b76fe5a1aff421209b97d1ac15720/resolwe/process/parser.py#L180 .

3. Use ``p.lineno`` in ``autoprocess.py`` to get the line where the class is defined.

JureZmrzlikar: Thanks for suggesting this @gregorjerse , makes total sense. I've applied the suggested changes. Please re-review.

PR #773, opened 2021-01-04
Feature/protocol improvements

gregorjerse:


resolwe/flow/executors/socket_utils.py
```diff
@@ -567,6 +570,17 @@ def __init__(
         self.peer_status_changed = peer_status_changed
         self._heartbeat_messages_future: Optional[asyncio.Future] = None
         self._watchdog_future: Optional[asyncio.Future] = None
+        self._recycle_uuids_future: Optional[asyncio.Future] = None
+
+    async def _recycle_received_uuids(self, timeout: int = 600):
+        """Recycle the list of received messages uuids.
+
+        Move received messages to the old list and distard old ones.
```
dblenkus, line 578: distard -> discard

resolwe/flow/executors/constants.py
```diff
@@ -9,7 +9,7 @@
 SETTINGS_VOLUME = Path("/settings")
 SOCKETS_VOLUME = Path("/sockets")
 
-CONTAINER_TIMEOUT = 300
+CONTAINER_TIMEOUT = 600
```
dblenkus, line 12: What is the reasoning behind this change? I assume that 5 minutes should be more than enough for the container to start.

resolwe/flow/managers/listener.py
```diff
@@ -1153,30 +1157,56 @@ async def _process_command(
...
-            response_time = time.time() - received_message.sent_timestamp
-            self.logger.debug(__("Response time: {}", response_time))
-            await self.communicator.send_response(response, peer_identity)
-
-        except RuntimeError:
-            # When unable to send message just log the error. Peer will be
-            # removed by the watchdog later on.
-            self.logger.exception(
-                "Protocol: error sending response to {received_message}."
+        if self._parallel_commands_semaphore is None:
```
dblenkus, line 1161: Is there a reason why you initialize semaphore here and not in the `__init__`?

resolwe/flow/managers/workload_connectors/kubernetes.py
```diff
@@ -493,9 +493,18 @@ def start(self, data: Data):
         requests["cpu"] = limits.pop("cores")
         limits["cpu"] = requests["cpu"] + 1
 
-        requests["memory"] = limits["memory"] * (10 ** 6)
-        limits["memory"] += KUBERNETES_MEMORY_HARD_LIMIT_BUFFER
+        # The memory in the database is stored in megabytes but the kubertenes
+        # requires memory in bytes.
+        # We request 10% less memory than stored in the database asd set limit
```
dblenkus, line 498: asd -> and

resolwe/flow/managers/workload_connectors/kubernetes.py
```diff
@@ -493,9 +493,18 @@ def start(self, data: Data):
...
+        # at 10% more plus KUBERNETES_MEMORY_HARD_LIMIT_BUFFER. The processes
+        # usually require 16GB, 32GB... and since the node usualy has 64GB of
+        # memory and some of it is consumed by the system processes only one
+        # process process that requires 32GB can run on a node instead of 2.
+
+        requests["memory"] = 0.8 * limits["memory"]
```
dblenkus, line 504: Should this be `0.9`?

resolwe/flow/managers/workload_connectors/kubernetes.py
```diff
@@ -493,9 +493,18 @@ def start(self, data: Data):
...
+        limits["memory"] = 1.1 * limits["memory"] + KUBERNETES_MEMORY_HARD_LIMIT_BUFFER
         limits["memory"] *= 10 ** 6
+        requests["memory"] *= 10 ** 6
```
dblenkus, line 507: Is it ok to use `10 ** 6` here or would it be better to change it to `2 ** 20`?

setup.py
```diff
@@ -64,23 +64,20 @@
...
     ],
     extras_require={
         # XXX: Temporarily pin docutils to 0.15.2 (last before 0.16) since botocore3 depends
         # on docutils<0.16 and pip tries to install 0.16.
         "storage_s3": [
-            "boto3~=1.12.7",
+            "boto3~=1.16.47",
             "crcmod",
-            "docutils~=0.15.2",
+            #"docutils~=0.15.2",
```
dblenkus, line 76: Remove this if it is not needed.

resolwe/flow/executors/constants.py
gregorjerse: It was an actual event in kuberenetes when processing container took 20s over 5 minutes to start.
I believe the cause was downolading docker image but I am not sure.

resolwe/flow/executors/socket_utils.py
gregorjerse: OK.

resolwe/flow/managers/listener.py
gregorjerse: When semaphore is created it requires running event loop. If there is none, one is created.

If not being careful it can quickly happen that the semaphore is bound to a different event loop than the one that is used in listener. See 

https://stackoverflow.com/questions/55918048/asyncio-semaphore-runtimeerror-task-got-future-attached-to-a-different-loop

I just thought it is nicer to create semaphore when it is needed and avoid troubles with different event loops no matter how listener was initialized.

resolwe/flow/managers/workload_connectors/kubernetes.py
gregorjerse: OK.

resolwe/flow/managers/workload_connectors/kubernetes.py
gregorjerse: Yes.

resolwe/flow/managers/workload_connectors/kubernetes.py
gregorjerse: We have to decide. I believe (2 ** 20) in indeed a better choice.
But how about storage? Should we stick with the same units?

resolwe/flow/executors/constants.py
dblenkus: Ok, it explains the problem. In Docker executor, the process is divided into two parts, pulling the image and starting the container, where only the second part has a timeout. Is it possible to do something similar in Kubernetes?

resolwe/flow/managers/listener.py
dblenkus: Ok, thanks for the explanation.

dblenkus (approved):

PR #774, opened 2021-01-04
Add UriResolverView

JureZmrzlikar:
@dblenkus @gregorjerse Please take a look and comment whether the approach seems fine.

### DIFF
--- a/resolwe/storage/views.py
+++ b/resolwe/storage/views.py
@@ -1,9 +1,8 @@
 """Storage application views."""
-
 from datetime import datetime
 from distutils.util import strtobool
 from pathlib import Path
-from typing import Union
+from typing import List, Optional, Tuple, Union
 
 import pytz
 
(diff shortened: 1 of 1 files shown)


resolwe/storage/views.py
```diff
@@ -49,6 +49,37 @@ def _path_to_dict(self, referenced_path: ReferencedPath, base_path: Path):
...
+        ]
+
+    def _resolve_file(self, relative_path, file_storage):
+        """Resolve file URI and return signed url."""
+        # Redirect to the resource.
+        force_download = strtobool(self.request.GET.get("force_download", "false"))
+        return file_storage.default_storage_location.connector.presigned_url(
+            file_storage.subpath / relative_path, force_download=force_download
+        )
+
```
gregorjerse, line 82: The approach seems fine. However, I believe the code would be nicer of you add a method like

```
 def get_response(datum: Data, relative_path: str) -> Optional[(str, bool)]:
    """"Return a response + if it is file."""
    file_storage: FileStorage = datum.location
    if (
        relative_path == Path()
        or file_storage.files.filter(path=relative_path.as_posix() + "/").exists()
    ):
        return (self._resolve_dir(relative_path, file_storage), False)
    elif file_storage.files.filter(path=relative_path.as_posix()).exists():
        return (self._resolve_file(relative_path, file_storage), True)
    else:
        return None, False
```

Then use it in `DataBrowseView.get` as:

```
    def get(
        self, request: HttpRequest, *args: Union[str, int], **kwargs: Union[str, int]
    ) -> HttpResponse:
        """Giver URL either redirect to resource or serve a list of resources.
        If URL points to a file then redirect to a default storage location.
        If URL points to a directory then send a JSON encoded list of entires
        in this directory emulating nginx autoindex module.
        :param request: the given request. When the user requests URL
            http://platform-centos.genialis.local/data/1/..., then the value of
            parameter should be 1/... .
        """

        data_id = kwargs.get("data_id")
        relative_path = Path(kwargs.get("uri", ""))
        datum = self._get_datum(data_id)
        response, is_file = self.get_response(datum, relative_path)
        if is_file:
            return redirect(response)
       elif response:
            return JsonResponse(response, safe=False)
       else:
             raise PermissionDenied()
```
and in `UriResolverView` as

```
    def get(self, request, *args, **kwargs):
        """Get resolved and signed URLs for a given list of URIs."""

        response_data = {}

        for uri in request.GET.getlist("uris", []):
            match = re.match(r"(\d+)/(.+)", uri).group(1, 2)
            if not match:
                response_data[uri] = ""
                continue
            data_id = int(match.group(1))
            relative_path = match.group(2)
            datum = self._get_datum(data_id)
           
            response = self.get_response(datum, relative_path)[0]
            if response:
                response_data[uri] = response
            else:
                raise PermissionDenied()
        return HttpResponse(response_data)
```

JureZmrzlikar: Thanks @gregorjerse  this makes sense. I was thinking along those lines, but couldn't decide what was more ugly: (a) a repeating code or (2) a method that does (and return) two things: file/dir check and uri resolution. :-)

I've modified my proposal.

gregorjerse (approved):

PR #781, opened 2021-01-22
Bump Resolwe dependencies

gregorjerse:


setup.py
```diff
@@ -65,8 +65,8 @@
         "redis~=3.5.3",
         "resolwe-runtime-utils~=3.0.0",
         "shellescape~=3.8.1",
-        "Sphinx>=1.5.1, <1.7.0",
         "beautifulsoup4~=4.9.3"
+        "Sphinx",
```
dblenkus, line 67: Limit the version of the Sphinx.

dblenkus (commented):
Looks good.

setup.py
gregorjerse: OK, will pin it to the current stable version.

setup.py
```diff
@@ -41,11 +41,11 @@
     install_requires=[
         # XXX: Temporarily pin asgiref to 3.2.x since testing framework freezes
         # with version 3.3.x
-        "asgiref~=3.2.10",
+        "asgiref~=3.3.1",
```
dblenkus, line 42: Remove the comment above if it was solved.

dblenkus (approved):

PR #784, opened 2021-02-05
Build tagged version of communication container

gregorjerse:
when commit is tagged. The tag itself is used as a docker image tag.
If the commit is not tagged build the development version of the image.

Additional commit:
- Fix errors and code cleanup
- Explicit separation of development and stable tags
- Push to ECR instead of Docker hub

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -2,10 +2,14 @@ name: Build Docker Images
 
 on:
   push:
-    branches: master
-  tags:
-    - "[0-9]+.[0-9]+.[0-9]+*"
-  
+    branches:
+      - master
+    tags:
+      - "[0-9]+.[0-9]+.[0-9]+*"
+
+env:
+  AWS_REGISTRY: public.ecr.aws/genialis
+  AWS_REGION: us-east-1
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


dblenkus (approved):

.github/workflows/docker.yml
```diff
@@ -43,3 +46,29 @@ jobs:
...
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build and push ${{ matrix.image }} image
```
dblenkus, line 68: There is no `matrix` variable in this job.

.github/workflows/docker.yml
```diff
@@ -3,6 +3,9 @@ name: Build Docker Images
 on:
   push:
     branches: master
+  tags:
+    - "[0-9]+.[0-9]+.[0-9]+*"
```
dblenkus, line 7: Is this also triggered for alpha releases?

.github/workflows/ci.yml
```diff
@@ -75,24 +75,24 @@ jobs:
         with:
           file: .coverage
       
-  build:
-    runs-on: ubuntu-20.04
+  # build:
```
dblenkus, line 78: Uncomment this.

.github/workflows/docker.yml
acopar: Yes, if we only wanted stable releases, the regex would be without the asterix.

PR #786, opened 2021-02-09
Remove os.fspath to be compatible with Python 3.4+

gregorjerse:


dblenkus (approved):

resolwe/flow/management/commands/list_docker_images.py
```diff
@@ -96,18 +98,20 @@ def handle(self, *args, **options):
...
 
                 # Update set of pulled images.
                 with PULLED_IMAGES_LOCK:
                     PULLED_IMAGES.add(img)
 
-                if ret != 0:
-                    errmsg = "Failed to pull Docker image '{}'!".format(img)
+                if ret.returncode != 0:
+                    errmsg = "Failed to pull Docker image '{}': {}!".format(
+                        img, ret.stdout
```
dblenkus, line 122: Is stdout a single line? If not, I would prefer to add a new line (or two) before it and remove the exclamation mark.

dblenkus (commented):
Looks good, I only have one small suggestion.

resolwe/flow/management/commands/list_docker_images.py
gregorjerse: It depends. It could be of arbitrary length.

dblenkus (approved):

PR #792, opened 2021-02-16
Do not block async loop while uploading log files

gregorjerse:
also make sure the log files do not change while uploading else file
descriptor will close and transfer error will occur.

### DIFF
--- a/resolwe/flow/executors/docker/run.py
+++ b/resolwe/flow/executors/docker/run.py
@@ -472,7 +472,6 @@ class FlowExecutor(LocalFlowExecutor):
 
         loop = asyncio.get_event_loop()
         start_time = time.time()
-        logger.debug("Starting init container: %s", init_arguments)
         init_container = client.containers.run(**init_arguments)
         init_container_status = await loop.run_in_executor(None, init_container.wait)
 
@@ -483,10 +482,8 @@ class FlowExecutor(LocalFlowExecutor):
             )
             return
 
-        logger.debug("Starting communication container: %s", processing_arguments)
         communication_container = client.containers.run(**communication_arguments)
 
-        logger.debug("Starting processing container: %s", processing_arguments)
         processing_container = client.containers.run(**processing_arguments)
         end_time = time.time()
         logger.info(
--- a/resolwe/flow/executors/docker/run.py
+++ b/resolwe/flow/executors/docker/run.py
@@ -6,6 +6,7 @@
 """
 # pylint: disable=logging-format-interpolation
 import asyncio
+import functools
 import json
 import logging
 import os
@@ -13,7 +14,7 @@ import platform
 import time
 from contextlib import suppress
 from pathlib import Path
-from typing import Dict, Optional, Tuple, Union
+from typing import Dict, Optional, Tuple, Type, Union
 
 import docker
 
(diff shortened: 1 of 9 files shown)


resolwe/flow/executors/startup_processing_container.py
```diff
@@ -624,8 +618,8 @@ def start_processing_container():
     JSON_LOG_PATH.touch()
 
     # Replace the code bellow with asyncio.run when we no longer have to
-    # support python 3.6.
+    # support python 3.4+.
```
dblenkus, line 621: I prefer to keep `3.6` here as `async.run` was added in 3.7, so we can start using it once we drop support for 3.6.

resolwe/flow/executors/startup_processing_container.py
```diff
@@ -530,9 +530,10 @@ def _handle_processing_script_connection(self, reader, writer):
 
             message = yield from communicator.receive_data()
             while message:
-                with JSON_LOG_PATH.open("at") as json_file:
-                    json_file.write(json.dumps(message) + os.linesep)
-                    self.log_files_need_upload = True
+                with (yield from self._uploading_log_files_lock):
```
dblenkus, line 533: https://docs.python.org/3/library/asyncio-sync.html
> Changed in version 3.9: Acquiring a lock using await lock or yield from lock and/or with statement (with await lock, with (yield from lock)) was removed. Use async with lock instead.

resolwe/flow/executors/docker/run.py
```diff
@@ -32,6 +33,52 @@
...
+                except retry_exceptions as err:
+                    print(f"Got exception {err}")
+                    sleep = min(max_sleep, min_sleep * (2 ** retry))
+                    last_error = err
+            print("Giving retry up.")
+            raise last_error
+
+        return wrapper_retry
+
+    print("Retry wrapper.")
```
dblenkus, line 78: Remove print.

resolwe/flow/executors/docker/run.py
```diff
@@ -32,6 +33,52 @@
 logger = logging.getLogger(__name__)
 
 
+def retry(
```
dblenkus, line 36: We use (almost) the same decorator in storage manager. It would be nice to combine them if possible.

resolwe/flow/executors/startup_communication_container.py
```diff
@@ -549,13 +554,17 @@ async def start(self) -> int:
...
+            except RuntimeError as runtime_exception:
                 logger.exception("Error processing script.")
                 with suppress(Exception):
                     await self.listener_communicator.send_command(
                         Message.command(
                             "process_log",
-                            {"error": ["Runtime error in communication container."]},
+                            {
+                                "error": [
+                                    f"Runtime error in communication container: {runtime_exception}."
```
dblenkus, line 565: Add a new line before `{runtime_exception}`.

resolwe/flow/tests/test_executors.py
```diff
@@ -297,8 +294,8 @@ def start_processing(data):
         self.assertEqual(data.worker.status, Worker.STATUS_PROCESSING)
         data.worker.terminate()
 
-        # Give it max 5 seconds to terminate.
-        for _ in range(50):
+        # Give it max 10 seconds to terminate.
```
dblenkus, line 297: Improve the commit message to make it obvious that it reffers to tests.

resolwe/flow/tests/test_executors.py
```diff
@@ -267,6 +267,9 @@ def test_run_in_docker(self):
         data = self.run_process("test-docker")
         self.assertEqual(data.output["result"], "OK")
 
+    @unittest.skipIf(
```
dblenkus: This is not needed, as you remove it later anyway.

dblenkus (approved):

resolwe/flow/executors/docker/run.py
gregorjerse: Executor virtual environment does not have storage manager code available. I could use the one from the docker executor on both ends.

resolwe/flow/executors/startup_communication_container.py
gregorjerse: Done.

resolwe/flow/tests/test_executors.py
gregorjerse: Done.

resolwe/flow/tests/test_executors.py
gregorjerse: Done.

resolwe/flow/executors/docker/run.py
gregorjerse: Done.

dblenkus (approved):
Looks good!

PR #793, opened 2021-02-18
Storage manager speedup

gregorjerse:


resolwe/storage/manager.py
```diff
@@ -18,251 +18,307 @@
...
+    def _preprocess_delay(self, delay: Optional[int]) -> int:
+        """Preprocess delay according to rules.
 
-        :param connector_name: name of the storage connector.
-        :type connector_name: str
-
-        :return: True when data should be copied to the location managed by
-            connector connector_name.
+        1. When delay is None return 0. Entries without explicit delay are
+            copied immediately.
```
dblenkus, line 25: This method is also used for deleting locations, so it should not use "copy". Use "process" instead.

resolwe/storage/manager.py
```diff
@@ -18,251 +18,307 @@
...
+                open_logs=models.Exists(
+                    AccessLog.objects.filter(
+                        storage_location__file_storage=models.OuterRef("id"),
+                        storage_location__connector_name=connector_name,
+                        finished__isnull=True,
+                    )
+                )
+            )
+            .annotate(
+                transfered_lately=models.Exists(
```
dblenkus, line 160: Do you need this filter, or do `open_logs` and `accessed_lately` cover this use-case? If yes, you have to take care that `last_update` is set before the log is released.

resolwe/storage/manager.py
```diff
@@ -18,251 +18,307 @@
...
+                        )
+                        .annotate(
+                            priority=models.Case(
+                                *default_connector_whens,
+                                default=DEFAULT_CONNECTOR_PRIORITY,
+                                output_field=models.IntegerField(),
+                            )
+                        )
+                        .order_by("priority")
+                        .values_list("connector_name", flat=True)[:1]
```
dblenkus, line 202: Can you use `.first()` instead of `[:1]`?

resolwe/storage/manager.py
```diff
@@ -18,251 +18,307 @@
...
-
-        :return: True if storage_location should be deleted, False otherwise.
-        :rtype: bool
-        """
-        rules = self._get_rules(storage_location.connector_name, "delete")
-        rule_results = {}
-        rule_results["has_rules"] = len(rules) > 0
 
-        rule_results["not_locked"] = not storage_location.locked
+        default_connector_whens = [
```
dblenkus, line 120: It would be better to name this `connectors_priority_whens`.

dblenkus (commented):
It would be nice to have some tests for transfer logic.

resolwe/storage/manager.py
```diff
@@ -18,251 +18,307 @@
...
-        ).exists()
+        whens = [
+            models.When(
+                data__process__type__startswith=process_type,
+                then=timedelta(days=self._preprocess_delay(override_rule.get("delay"))),
+            )
+            for process_type, override_rule in process_overrides.items()
+        ]
+        whens += [
+            models.When(
```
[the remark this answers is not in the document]
gregorjerse: Done.

resolwe/storage/manager.py
gregorjerse: Done.

resolwe/storage/manager.py
gregorjerse: Unfortunately not. I had to refresh my memory but it has something to do with first being eager and not lazy and can not be used in such subquery.

resolwe/storage/manager.py
gregorjerse: OK.

resolwe/storage/manager.py
gregorjerse: Fixed.

resolwe/storage/models.py
```diff
@@ -92,6 +97,201 @@ def get_queryset(self) -> models.QuerySet:
...
+            .annotate(
+                all_locations_count=models.Count(
+                    "storage_locations",
+                    filter=models.Q(
+                        storage_locations__status=StorageLocation.STATUS_DONE
+                    ),
+                    distinct=True,
+                )
+            )
+            .annotate(
```
dblenkus, line 179: Can you merge all annotations and separate them with commas? Would that give the same result?

resolwe/storage/models.py
```diff
@@ -92,6 +97,201 @@ def get_queryset(self) -> models.QuerySet:
...
+                    ),
+                    models.Value(""),
+                )
+            )
+            .filter(
+                all_locations_count__gt=min_other_copies,
+                open_logs=False,
+                accessed_lately=False,
+            )
+            .filter(
```
dblenkus, line 227: Is there a reason for two separate filters?

dblenkus (approved):
Nice!

resolwe/storage/models.py
gregorjerse: I see no reason why not. I have merged the annotations into one (and also some filters).

resolwe/storage/models.py
gregorjerse: Just that I prefer to group logically connected (at least in my mind) filters together. I have merged them into one now.

PR #797, opened 2021-02-19
Allow to modify Entity in Python processes

gregorjerse:


resolwe/flow/managers/listener/python_process_plugin.py
```diff
@@ -99,6 +99,30 @@ class ExposeObjectPlugin(metaclass=abc.ABCMeta):
...
+            [full_permission_name],
+            object_,
+        )
+        if not filtered_object:
+            if object_:
+                raise RuntimeError(
+                    f"No permissions: {model._meta.model_name} with id {model_pk}."
+                )
+            else:
+                raise RuntimeError(
```
dblenkus, line 122: You can raise this exception before checking permissions.

dblenkus (commented):
It would be nice to add test for these features.

resolwe/flow/managers/listener/python_process_plugin.py
gregorjerse: OK, more efficient. I will change that and add tests.

gregorjerse:
> It would be nice to add test for these features.

Done.

resolwe/process/tests/test_python_process.py
```diff
@@ -389,6 +389,22 @@ def test_create_data(self):
         Collection.objects.get(name=collection_name)
         Data.objects.get(name=data_name)
 
+    @with_docker_executor
+    @tag_process("assign-entity-tags")
+    def test_assign_tags_entity(self):
+        """Assign tags to entity."""
+        data = self.run_process(
```
dblenkus, line 396: It would be nice to create another entity without edit permission and see what happens if the process tries to change it. Ideally, the listener should return an error message and a verbose error should be raised inside the process.

resolwe/flow/managers/listener/python_process_plugin.py
```diff
@@ -299,6 +299,15 @@ def can_update(
         if not_allowed_keys:
             raise RuntimeError(f"Not allowed to set {','.join(not_allowed_keys)}.")
 
+        print()
```
dblenkus, line 302: 🤯

dblenkus (changes requested):

resolwe/flow/managers/listener/python_process_plugin.py
gregorjerse: Removed.

dblenkus (approved):

PR #804, opened 2021-02-27
Handle exception when uploading data

gregorjerse:
When keep_data flag is set and there was error during processing the
exception was thrown that was not handled.

### DIFF
--- a/resolwe/flow/executors/startup_processing_container.py
+++ b/resolwe/flow/executors/startup_processing_container.py
@@ -221,9 +221,12 @@ class ProtocolHandler:
             yield from self.manager.upload_log_files()
 
             if KEEP_DATA:
-                yield from self.manager.send_file_descriptors(
-                    [str(file_) for file_ in Path("./").rglob("*") if file_.is_file()]
-                )
+                try:
+                    yield from self.manager.send_file_descriptors(
+                        [str(file) for file in Path("./").rglob("*") if file.is_file()]
+                    )
+                except:
+                    logger.exception("Error sending file descriptors (keep_data).")
 
             response = respond(message, "OK", self.return_code)
             yield from self.communicator.send_data(response)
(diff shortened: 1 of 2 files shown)


resolwe/flow/executors/startup_processing_container.py
```diff
@@ -221,9 +221,16 @@ def process_script(self, message):
             yield from self.manager.upload_log_files()
 
             if KEEP_DATA:
-                yield from self.manager.send_file_descriptors(
-                    [str(file_) for file_ in Path("./").rglob("*") if file_.is_file()]
-                )
+                try:
+                    yield from self.manager.send_file_descriptors(
+                        [
+                            str(file_)
```
dblenkus, line 227: `file` is not a keyword in Python 3, so there is no need for the underscore.

resolwe/flow/executors/startup_processing_container.py
```diff
@@ -500,7 +506,9 @@ def send_chunk(processing_filenames):
             )
             if self.upload_socket.recv(1) == b"0":
                 raise RuntimeError(
-                    "Error sending filenames: {}.".format(processing_filenames)
+                    "Error sending filenames: {} (comminication container response indicates error).".format(
```
dblenkus, line 509: Would it make sense to move the comment (in parenthesis) before the colon, just if the list of filenames is long?

dblenkus (approved):

resolwe/flow/executors/startup_processing_container.py
gregorjerse: Good point, much nicer code (in a single line).

resolwe/flow/executors/startup_processing_container.py
gregorjerse: I have rephrased the error message and split it into two rows.

PR #806, opened 2021-03-01
Feature/kubernetes improvements

gregorjerse:


dblenkus (approved):
Try to avoid using backticks in commit messages - use quotes instead.

PR #808, opened 2021-03-04
Add multipart upload to local and S3 connector

gregorjerse:


resolwe/storage/connectors/localconnector.py
```diff
@@ -48,6 +48,39 @@ def push(self, stream, url, chunk_size=BaseStorageConnector.CHUNK_SIZE):
...
 
+    def multipart_push_start(self, url, size=None):
+        """Start a multipart upload.
+
+        :returns: the upload id.
+        """
+        assert size is not None, f"{self}: size of multipart upload must be specified."
+        path = self.base_path / url
+        if not path.is_file():
+            with path.open("wb") as stream:
```
dblenkus, line 59: Should it fail if the file already exists?

resolwe/storage/connectors/s3connector.py
```diff
@@ -251,3 +293,57 @@ def presigned_url(self, url, expiration=60, force_download=False):
 
         # The response contains the presigned URL
         return response
+
+
+# {
```
dblenkus, line 298: Remove this.

resolwe/storage/connectors/localconnector.py
gregorjerse: This means we have a file name clash or browser re-posted the upload for some reason.

The first scenario is very unlikely but could have dire consequences (some data may be lost) and upload should be aborted. I will fix the code.

resolwe/storage/connectors/localconnector.py
```diff
@@ -50,6 +50,41 @@ def push(self, stream, url, chunk_size=BaseStorageConnector.CHUNK_SIZE):
...
+    def multipart_push_start(self, url, size=None):
+        """Start a multipart upload.
+
+        :returns: the upload id.
+        :raises AssertionError: when url exists or upload size is not given.
+        """
+        assert size is not None, f"{self}: size of multipart upload must be specified."
+        path = self.base_path / url
+        assert not path.exists(), f"Path {url} already exists, aborting upload."
+        if not path.is_file():
```
dblenkus, line 62: You can probably remove this as you already have `assert` above.

dblenkus (approved):

PR #809, opened 2021-03-08
Storage manager improvements

gregorjerse:


resolwe/storage/manager.py
```diff
@@ -28,6 +28,15 @@ def _lock_file_storage(self, file_storage_id: int) -> Optional[FileStorage]:
 
     def delete_single_location(self, file_storage: FileStorage, connector_name: str):
         """Delete given storage location."""
+        # Some other storage manager might have already processed this storage
+        # location before we did.
+        if not file_storage.storage_locations.filter(
+            connector_name=connector_name,
+            url=file_storage.subpath,
```
dblenkus, line 35: Why do you need to filter by url, as it is already linked to the file storage?

resolwe/storage/manager.py
```diff
@@ -46,23 +55,37 @@ def process_delete(self):
...
-                    )
-                    try:
-                        self.delete_single_location(file_storage, connector_name)
-                    except Exception:
-                        logger.exception(
-                            "Error deleting data from StorageLocation instance",
-                        )
+                    # Some other storage manager might have already processed this storage
+                    # location before we did.
+                    if not file_storage.storage_locations.filter(
```
dblenkus, line 60: This check is unnecessary; it only produces an extra query. It can fail only inside the `self.delete_single_location`, and it already does the same check.

resolwe/storage/manager.py
```diff
@@ -28,6 +28,15 @@ def _lock_file_storage(self, file_storage_id: int) -> Optional[FileStorage]:
 
     def delete_single_location(self, file_storage: FileStorage, connector_name: str):
         """Delete given storage location."""
+        # Some other storage manager might have already processed this storage
+        # location before we did.
+        if not file_storage.storage_locations.filter(
+            connector_name=connector_name,
+            url=file_storage.subpath,
+        ).exists():
```
dblenkus, line 36: This is still not atomic. You should do it like:
```py
delete_location = <queryset>.first()
if not delete_location:
    return
```

resolwe/storage/manager.py
```diff
@@ -79,7 +102,7 @@ def copy_single_location(self, file_storage: FileStorage, connector_name: str):
             storage_location.transfer_data(new_storage_location)
             new_storage_location.status = StorageLocation.STATUS_DONE
             new_storage_location.save()
-        except Exception:
```
dblenkus, line 82: Why have you removed this?

resolwe/storage/manager.py
```diff
@@ -46,23 +55,37 @@ def process_delete(self):
...
     def copy_single_location(self, file_storage: FileStorage, connector_name: str):
         """Copy given location to a given connector."""
+
+        # Some other storage manager might have already processed this storage
+        # location before we did.
+        if file_storage.storage_locations.filter(
+            connector_name=connector_name,
+            url=file_storage.subpath,
+            status=StorageLocation.STATUS_DONE,
+        ).exists():
```
dblenkus, line 81: Same as above. You should create the object instead of just checking that it does not exist.

resolwe/storage/manager.py
gregorjerse: Fixed.

resolwe/storage/manager.py
gregorjerse: Actually, I was not sure where exactly to place this check. In the end, I forgot to remove the redundant one :/

resolwe/storage/manager.py
gregorjerse: Are you sure about that? 

The FileStorage object is locked so no other manager can process it at the same time.

resolwe/storage/manager.py
gregorjerse: I believe it is clearer to just write `except` instead of `except Exception`. I know they are not 100% compatible but that should make no difference in this situation.

dblenkus: Actually, it does matter. Especially as this is a long-running task, and the except (without `Exception`) also catches `SystemExit` and `KeyboardInterrupt`, so the program cannot be interupted.

resolwe/storage/manager.py
dblenkus: You are right. But it is still nicer to make fewer queries.

dblenkus (approved):

PR #810, opened 2021-03-08
Handle very large inputs correctly

gregorjerse:
Increase EBS size by 10% to account for file system overhead on very
large input volumes.

### DIFF
--- a/resolwe/flow/managers/workload_connectors/kubernetes.py
+++ b/resolwe/flow/managers/workload_connectors/kubernetes.py
@@ -14,7 +14,6 @@ import shlex
 import time
 from base64 import b64encode
 from contextlib import suppress
-from math import ceil
 from pathlib import Path
 from typing import Any, Dict, List
 
@@ -484,6 +483,10 @@ class Connector(BaseConnector):
     def _data_inputs_size(self, data: Data, safety_buffer: int = 2 ** 30) -> int:
         """Get the size of data inputs.
 
+        Also add 10% of the volume size + 2GB as safety buffer. When having a
+        large filesystem with lots of files the filesystem overhead can be
+        significant so a fixed safety buffer alone is not sufficient.
+
         :returns: the size of the input data in bytes.
         """
         inputs_size = (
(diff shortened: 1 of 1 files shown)


resolwe/flow/managers/workload_connectors/kubernetes.py
```diff
@@ -484,6 +484,10 @@ def _fix_permissions(self) -> List[str]:
     def _data_inputs_size(self, data: Data, safety_buffer: int = 2 ** 30) -> int:
         """Get the size of data inputs.
 
+        Also add 2GB + 10% of the volume size as safety buffer. When having a
```
dblenkus, line 487: It would be easier to understand if you write "10% + 2GB".

resolwe/flow/managers/workload_connectors/kubernetes.py
```diff
@@ -495,7 +499,7 @@ def _data_inputs_size(self, data: Data, safety_buffer: int = 2 ** 30) -> int:
             .get("total_size")
         )
         assert isinstance(inputs_size, int)
-        return ceil((inputs_size + safety_buffer) / (2 ** 30)) * (2 ** 30)
+        return ceil(((inputs_size + safety_buffer) / (2 ** 30)) * 1.1) * (2 ** 30)
```
dblenkus, line 502: Where do you add 2GB?

dblenkus: Do you still need to round the number and then add extra space? It looks too complex; probably you can do only the latter.

dblenkus (commented):
Changelog.

resolwe/flow/managers/workload_connectors/kubernetes.py
gregorjerse: 1. Safety buffer equals 2 GB by default, it can be increased.
2. Actually the requirements are specified in bytes, and my intention was to round it to the nearest gigabyte. So I add both numbers, divide to get the number of gigabytes, ceil it to the nearest integer and multiply by 2**30 again. I will make it less complicated.

dblenkus: Obviously, it was too early, I haven't seen `safety_buffer` :). But still, what is the benefit of rounding it to a gigabyte? Wouldn't it work the same with the actual size?

gregorjerse: It most probably would. Andrej noticed that EBS size was always reported in gigabytes so I rounded it to the nearest gigabyte. But even so, 2GB of safety buffer and 10% of size should take care of any rounding problems.

dblenkus (approved):

PR #820, opened 2021-03-31
Add default value to PROJECT_ROOT if not present

gregorjerse:
to make build_sphinx command work

### DIFF
--- a/resolwe/storage/settings.py
+++ b/resolwe/storage/settings.py
@@ -2,13 +2,16 @@
 
 Used to provide s simple default configuration.
 """
+from pathlib import Path
+
 from django.conf import settings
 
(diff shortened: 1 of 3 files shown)


resolwe/storage/settings.py
```diff
@@ -2,13 +2,16 @@
 
 Used to provide s simple default configuration.
 """
+from pathlib import Path
+
 from django.conf import settings
 
+root = Path(getattr(settings, "PROJECT_ROOT", "/"))
```
dblenkus, line 9: It would be better to rename the variable to `project_root`.

dblenkus (approved):

resolwe/storage/settings.py
gregorjerse: OK.

PR #830, opened 2021-04-23
Feature/slurm

gregorjerse:


resolwe/flow/managers/workload_connectors/slurm.py
```diff
@@ -61,6 +61,9 @@ def submit(self, data, runtime_dir, argv):
                 script.write("#SBATCH --cpus-per-task={}\n".format(limits["cores"]))
                 if partition:
                     script.write("#SBATCH --partition={}\n".format(partition))
+                    script.write(
+                        "#SBATCH --output slurm-data-{}-job-%j.out\n".format(data.pk)
```
dblenkus, line 65: Should this be data id or location id?

gregorjerse: I believe it makes most sense it is the data id.

We need a way to connect slurm job to the data object  so we can check what the executor output for the data object. Previously it was inside subdirectory named after location url but I see no reason why not to just add data id here and skip one extra step (retrieving location url from data object id) when debugging.

dblenkus (approved):

PR #831, opened 2021-04-23
Feature/bcm selinux label

gregorjerse:


resolwe/flow/executors/run.py
```diff
@@ -53,7 +54,9 @@ def __init__(
         self.storage_url = Path(LOCATION_SUBPATH)
         self.runtime_dir = Path(SETTINGS["FLOW_VOLUMES"]["runtime"]["config"]["path"])
         self.communicator = communicator
-        self.listener_connection = listener_connection
+        self.listener_connection = list(listener_connection)
+        if platform.system() == "Darwin":
```
dblenkus, line 58: Why don't you use the following condition? In theory, it should also work on windows.
```py
if sys.platform not in ["linux", "linux2"]:
    self.listener_connection[0] = "host.docker.internal"
```

gregorjerse: Nice idea.

dblenkus (approved):

PR #832, opened 2021-04-24
Add test output management command

gregorjerse:
The command is not finished. There are some issues to resolve.

- some connectors (S3...) store only files and not directories, so directory tests will fail.
- just by iterating through schema it is impossible to tell if reference is a directory and what should that directory contain.
- I could consult the database to check if the entry is a directory or not. If it is I could also obtain the list of all entries in that directory. But let stick with basic file testing for now.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -6,6 +6,16 @@ All notable changes to this project are documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_.
 
 
+==========
+Unreleased
+==========
+
+Added
+-----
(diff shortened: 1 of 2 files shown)


gregorjerse:
We already have `validate_data` method on the `StorageLocation` object that checks the existance of every referenced path in the given `storage location`. 

Do we still need this check? It has some benefits (checks that the outputs are really there). If yes, I suggest we simplify it and only chech that the outputs of the `data` objects are referenced in the database. The health of the storage locations can then be checked with the `validate_data` if necessary.

dblenkus (approved):

PR #835, opened 2021-05-03
Fix heartbeat / timing bug

gregorjerse:


resolwe/storage/connectors/transfer.py
```diff
@@ -271,7 +271,7 @@ def future_done(stream_to_close, future):
                     executor.shutdown(wait=False)
 
             data_stream = CircularBuffer(
-                buffer_size=min(200 * 1024 * 1024, object_["size"])
+                buffer_size=min(200 * 1024 * 1024, object_["size"] + 1)
```
dblenkus, line 274: It would be cleaner in my opinion if you use:```py
max(min(200 * 1024 * 1024, object_["size"], 1))
```
Or at least add a comment to explain the offset.

dblenkus (approved):

resolwe/storage/connectors/transfer.py
gregorjerse: I was actually thinking about this. My opinion is that the additional max just complicates things. I will add a comment on why  `+1` is needed.

PR #836, opened 2021-05-04
Do not raise exception when exiting runlistener

gregorjerse:
Wait for the future started inside context manager instead of creating
the new one, which causes AssertionError to be thrown on script exit.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -6,6 +6,15 @@ All notable changes to this project are documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_.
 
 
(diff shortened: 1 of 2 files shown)


dblenkus (commented):
Using a private attribute doesn't look right. Is there any other way of handling this?

gregorjerse:
Sure. The previous way of doing it was to wait for the process to end in the exit stage of the context manager but I am not sure I like that approach... I will refactor the code.

gregorjerse:
A have changed code to await public asyncio event `` should_stop``, which basically does the same thing.

dblenkus (approved):

PR #838, opened 2021-05-05
Perform check if value is None correctly

gregorjerse:


AnzeLovse (approved):
Thanks, the process no longer fails if an optional `FileFields` is used in a `GroupField`.

dblenkus (approved):

PR #847, opened 2021-05-14
Transform docker image names with mapper

gregorjerse:
When docker image name starts with the key in the mapper the matching
part is replaced with the corresponding value.


dblenkus (approved):
Add a change log.

PR #855, opened 2021-06-10
When filtering Processes by slug retrieve the one with the latest version

gregorjerse:


jkokosar (commented):
Thank you for improving this functionality!

JureZmrzlikar:
Should `get_latest` be added also for DescriptorSchema model? I can see this handy in the future if we wish to set descriptor & descriptor schema within the process...

resolwe/process/tests/test_python_process.py
```diff
@@ -377,6 +377,37 @@ def test_get_collection(self):
...
+        self.assertEqual(data.output["process_pk"], process1.pk)
+
+        process2 = Process.objects.create(
+            contributor=self.admin,
+            persistence=Process.PERSISTENCE_TEMP,
+            version="1.0.1",
+            slug=process_slug,
+        )
+        data = self.run_process("get-latest-process", {"process_slug": process_slug})
+        self.assertEqual(data.output["process_pk"], process2.pk)
```
dblenkus, line 409: This can pass even if the method doesn't work as intended. I think it would be much more robust if you bump the version of `process1` to `2.0.0` and rerun the process to ensure it returns a different result.

dblenkus (approved):

PR #863, opened 2021-07-19
WIP: remove EFS

gregorjerse:


resolwe/flow/management/commands/prepare_runtime.py
```diff
@@ -16,6 +16,59 @@
...
+    :raises AssertionError: when destination_dir is not a directory or source
+        path of the Python process runtime can not be determined.
+    """
+    assert (
+        destination_dir.is_dir()
+    ), f"The path '{destination_dir}' must be a directory."
+
+    for runtime_class_name in settings.FLOW_PROCESSES_RUNTIMES:
+        module_name, class_name = runtime_class_name.rsplit(".", 1)
+        runtime_module = import_module(module_name)
```
dblenkus, line 56: I don't like that all modules are imported here, as they are configurable, and arbitrary code can be executed.

Can you do it with the following snippet?
```py
import importlib
spec = importlib.util.find_spec(module_name)
spec.origin
```

dblenkus (commented):
It would be great if we can remove the `prepare_runtime` management command. It only adds to the complexity and doesn't really solve any problem. 

However, we would need to figure out when to refresh the runtime cache or prepare the whole runtime each time from scratch. But, this is also not handled correctly in the current implementation.

resolwe/flow/management/commands/prepare_runtime.py
gregorjerse: Yes, that works like a charm. Thanks.

resolwe/process/communicator.py
```diff
@@ -4,7 +4,10 @@
 from pathlib import Path
 from typing import Any, Optional, Type
 
-from .socket_utils import Message, Response, receive_data, send_data
+try:
```
dblenkus, line 7: Add a comment why this is required.

dblenkus (approved):
LGTM

PR #865, opened 2021-07-20
Separate build and push workflow for stable and staging tags

plojyon:
- Build configuration for stable containers now only includes release tags, not pre-release tags (a* versions).
- Additional build configuration for staging matches all tags.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -13,6 +13,8 @@ env:
 
 jobs:
   build:
+    env:
+      TAGS:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


acopar (commented):
Good overall, there are some changes needed to remove debugging checks and some to improve clarity. See if it is possible to combine all three docker build steps into one, since only tags need to be changed. Perhaps tags can be prepared in advance and passed to the build as comma-separated strings.

acopar (approved):

acopar (commented):
Looks good. Please rebase and we can merge after deploy.

PR #877, opened 2021-09-06
Support assuming role in S3 connector

gregorjerse:


acopar (approved):

resolwe/storage/connectors/transfer.py
```diff
@@ -50,14 +50,22 @@
...
             return wrapped(*args, **kwargs)
-        except transfer_exceptions as err:
-            connection_err = err
-            sleep(ERROR_TIMEOUT)
-
-    raise connection_err
+        except transfer_exceptions:
+            # Log the exception on retry for inspection.
+            if retry != ERROR_MAX_RETRIES:
+                logger.exception(
```
dblenkus, line 59: It would be nice also to log the exception if the last retry fails.

resolwe/storage/connectors/transfer.py
```diff
@@ -284,6 +284,9 @@ def transfer(
             stream = to_connector.open_stream(to_base_url / to_url, "wb")
             from_connector.get(from_url, stream, chunk_size=chunk_size)
             stream.close()
+            # Store computed hashes as metadata for later use.
+            to_connector.set_hashes(to_base_url / to_url, hashes)
```
dblenkus, line 288: Merge this commit with the previous one, so it will be obvious that you've moved the same chunk of code.

PR #881, opened 2021-10-14
Add metrics for disk and memory usage

acopar:
On hold until we determine why this feature causes etag hashes to be out-of-sync on uploaded files.

IT-24


acopar:
Closed in favor of external node and pod metrics provided by kubelet and custom kube-exporter solutions.

PR #889, opened 2021-11-04
Fix/permissions

gregorjerse:


resolwe/permissions/models.py
```diff
@@ -260,8 +260,10 @@ def get_permission(self, user_or_group: UserOrGroup) -> Permission:
         if entity_name == "user" and entity.is_superuser:
             return Permission.highest()
 
-        permission_models = self.permissions.filter(**{entity_name: entity})
-        return permission_models[0].permission if permission_models else Permission.NONE
+        try:
+            return self.permissions.get(**{entity_name: entity}).permission
+        except PermissionModel.DoesNotExist:
```
dblenkus, line 267: Overhead with error handling in Python can be considerable, so it should be only used for rare errors.

gregorjerse: I believe this will be a rare error. The method is used mostly in tests and when setting permissions, not in common operations (such as filtering objects). 

I am aware that Python exception handling has overhead, but the previous code produced two queries to the database: first to check that a record exists and then another one to fetch its data. I would argue that exception handling overhead is lower than one request to the database.

I can also do some timing ;)

resolwe/permissions/models.py
```diff
@@ -142,12 +142,25 @@ def __str__(self) -> str:
         return self.name.lower()
 
 
+class PositivePermissions(models.Manager):
```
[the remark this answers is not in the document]
gregorjerse: Sure.

dblenkus (approved):

PR #892, opened 2021-11-15
Store only refecences to files in the data storage

gregorjerse:
so references to files in other storages (for example 'upload') are not
stored in the database.

### DIFF
--- a/resolwe/storage/migrations/0010_remove_export_paths.py
+++ b/resolwe/storage/migrations/0010_remove_export_paths.py
@@ -1,0 +1,40 @@
+from django.db import migrations
+
+
+def delete_nonexisting_paths(apps, schema_editor):
+    """Remove referenced paths that point to exported files.
+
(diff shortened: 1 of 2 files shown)


resolwe/storage/migrations/0010_remove_export_paths.py
```diff
@@ -0,0 +1,40 @@
...
+    """Remove referenced paths that point to exported files.
+
+    Due to a bug in communication container references to all files are stored
+    in the datase, not only to files in 'data' storage.
+
+    Specifically the database contains references to files in the 'upload'
+    storage, which are named 'export_#hex', where #hex is a 32 characters long
+    hexadecimal stored as string.
+
+    This scripts detects such paths and removes them if they do not exist it
```
dblenkus, line 14: it -> in

dblenkus (approved):

PR #894, opened 2021-11-26
Fix/move to collection

gregorjerse:


resolwe/flow/serializers/data.py
```diff
@@ -103,11 +105,12 @@ def validate_collection(self, collection):
             self.instance.validate_change_collection(collection)
         return collection
 
+    @transaction.atomic
     def update(self, instance, validated_data):
         """Update."""
-        pre_update_collection = instance.collection
+        must_update_collection = "collection" in validated_data
```
dblenkus, line 111: I would name this `update_collection` to make it shorter.

dblenkus (approved):

PR #901, opened 2022-01-07
New helm

gregorjerse:


resolwe/flow/executors/prepare.py
```diff
@@ -49,9 +49,9 @@ def extend_settings(self, data_id, files, secrets):
         """
         pass
 
-    def get_tools_paths(self):
+    def get_tools_paths(self, kubernetes=False):
```
dblenkus, line 52: Name variables after the functionality, not after technology (i.e. `use_cache` or something similar).

resolwe/flow/managers/dispatcher.py
```diff
@@ -303,14 +303,18 @@ def _get_listener_settings(
         listener_settings = getattr(settings, "FLOW_EXECUTOR", {}).get(
             "LISTENER_CONNECTION", {}
         )
+        container_listener_connection = getattr(
+            settings,
+            "COMMUNICATION_CONTAINER_LISTENER_CONNECTION",
+            {"local": "127.0.0.1"},
```
dblenkus, line 309: Try to avoid putting default values into the code. One option is to add a sanity check at the startup to verify that the setting is set, and the second option would be to add a dedicated file for default settings (e.g. `flow/setings.py`).

resolwe/flow/executors/docker/run.py
```diff
@@ -370,6 +377,7 @@ async def start(self):
         # container due to name clash.
         init_container_name = f"{self.container_name}-{_random_string()}-init"
 
+        # Init container must run as root to use named volumes.
```
dblenkus, line 378: Move this comment just above the `"user": "0:0"` line.

dblenkus (commented):
It would be great to merge all comment fixes into a single commit.

resolwe/flow/managers/consumer.py
```diff
@@ -90,3 +96,32 @@ async def control_event(self, message):
...
+        """Perform health check.
+
+        We are testing the channels layer and database layer. The channels
+        layer is already functioning if this method is called so we have to
+        perform database check.
+
+        If the check is successfull touch the file specified in the channels
+        message.
+        """
+        print("Healt check")
```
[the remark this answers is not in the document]
gregorjerse: I have replaced them with logger debug messages.

resolwe/flow/managers/consumer.py
```diff
@@ -90,3 +96,32 @@ async def control_event(self, message):
...
+        path = Path(message["file"])
+
+        if await self.check_database():
+            print("All OK")
```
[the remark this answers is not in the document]
gregorjerse: I have replaced them with logger debug messages.

resolwe/flow/executors/docker/run.py
gregorjerse: Done.

PR #903, opened 2022-02-01
Advanced filters

plojyon:


resolwe/flow/filters.py
```diff
@@ -269,15 +287,38 @@ class Meta(BaseResolweFilter.Meta):
 class CollectionFilter(BaseCollectionFilter):
     """Filter the Collection endpoint."""
 
+    sample_count = filters.NumberFilter(
+        method="count_samples", field_name="entity__count"
```
gregorjerse, line 291: In Resolwe the term `entity` should be used instead of `sample` . So change method `count_samples` to `count_entities`. This applies to the entire PR.

resolwe/flow/filters.py
```diff
@@ -269,15 +287,38 @@ class Meta(BaseResolweFilter.Meta):
...
+        method="count_samples", field_name="entity__count__lte"
+    )
+
     class Meta(BaseCollectionFilter.Meta):
         """Filter configuration."""
 
         model = Collection
 
+    def count_samples(self, queryset, name, value):
+        """Filter by the number of associated samples (entities)."""
```
gregorjerse, line 312: Use `entity` instead of `sample` also in documentation and comments.

resolwe/flow/filters.py
```diff
@@ -250,6 +266,8 @@ class BaseCollectionFilter(TextFilterMixin, UserFilterMixin, BaseResolweFilter):
     contributor_name = filters.CharFilter(method="filter_contributor_name")
     owners = filters.CharFilter(method="filter_owners")
     owners_name = filters.CharFilter(method="filter_owners_name")
+    shared_with_me = filters.BooleanFilter(method="filter_private")
+    group = filters.CharFilter(method="filter_for_group")
```
gregorjerse, line 270: How exactly is the group filter supposed to bo used?

resolwe/flow/filters.py
```diff
@@ -163,6 +163,22 @@ def filter_for_user(self, queryset, name, value):
...
         return queryset.filter_for_user(user, Permission.from_name(value))
 
+    def filter_for_group(self, queryset, name, value):
+        """Filter queryset by group."""
+        return queryset.filter(permission_group__permissions__group=value)
+
+    def filter_private(self, queryset, name, value):
+        """Return only elements that are not public (explicitly shared with me)."""
+
+        public_user = get_user_model().objects.get(username="public")
```
gregorjerse, line 173: Use helper method `get_anonymous_user` defined in `resolwe.permissions.models`. 

Hint: the anonymous user username is defined in settings and is not necessarily equal to the string `public`.

resolwe/flow/filters.py
```diff
@@ -163,6 +163,22 @@ def filter_for_user(self, queryset, name, value):
...
+        public_filter = Q(permission_group__permissions__user=public_user)
+
+        # if filter value is False, return only public elements
+        if value:
+            return queryset.filter(~public_filter)
```
gregorjerse, line 178: I am not sure but maybe it would be nicer to use exclude here instead of negating the Q object.

resolwe/flow/filters.py
```diff
@@ -163,6 +163,22 @@ def filter_for_user(self, queryset, name, value):
...
+        """Filter queryset by group."""
+        return queryset.filter(permission_group__permissions__group=value)
+
+    def filter_private(self, queryset, name, value):
+        """Return only elements that are not public (explicitly shared with me)."""
+
+        public_user = get_user_model().objects.get(username="public")
+        public_filter = Q(permission_group__permissions__user=public_user)
+
+        # if filter value is False, return only public elements
```
gregorjerse, line 176: I would prefer the comment to be associated with the row immediately below the comment.
So maybe `If filter value is True, return only private elements` would be more appropriate.

gregorjerse (changes requested):

resolwe/flow/filters.py
plojyon: User supplies the ID of the group to filter by. I meant to use `NumberFilter`, but on second thought, `ModelChoiceFilter` seems even more appropriate.

resolwe/flow/filters.py
```diff
@@ -163,6 +164,21 @@ def filter_for_user(self, queryset, name, value):
...
+    def filter_for_group(self, queryset, name, value):
+        """Filter queryset by group."""
+        return queryset.filter(permission_group__permissions__group=value)
+
+    def filter_private(self, queryset, name, value):
+        """Return only elements that are not public (explicitly shared with me)."""
+
+        public_filter = Q(permission_group__permissions__user=get_anonymous_user())
+
+        # exclude public elements if filter value is True
```
gregorjerse, line 176: Please capitalize.

gregorjerse (approved):
There are 2 minor things left:
1. Capitalize comment.
2. Change wording is commit message.

Please also rebase to latest Reslowe to fix linter problems.

PR #906, opened 2022-02-18
Feature/django32python310

gregorjerse:


resolwe/flow/executors/startup_processing_container.py
```diff
@@ -344,8 +344,13 @@ def __init__(self, loop):
         self.exported_files_mapper = dict()  # type: Dict[str, str] # noqa: F821
         self.upload_socket = None
         self.update_logs_future = None
-        self._uploading_log_files_lock = asyncio.Lock(loop=loop)
-        self._send_file_descriptors_lock = asyncio.Lock(loop=loop)
+        # Drop this when we stop supporting Python 3.7.
+        # This class has been implicitly getting the current running loop since 3.7.
+        lock_arguments = {}
+        if sys.version_info < (3, 7):
```
dblenkus, line 350: The comment says 3.7, but here it is 3.6.

tests/settings.py
```diff
@@ -59,6 +59,8 @@
     "resolwe.permissions.permissions.ResolwePermissionBackend",
 )
 
+DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
```
dblenkus, line 61: This is already the default value: https://docs.djangoproject.com/en/4.0/releases/3.2/#customizing-type-of-auto-created-primary-keys

setup.py
```diff
@@ -38,46 +38,43 @@
...
         # Storage requirement for computing hashes.
         "crcmod",
-        "kubernetes~=12.0.1",
-        "docker~=4.4.4",
+        "kubernetes~=21.7.0",
+        "docker~=5.0.3",
         "Django~=3.2.12",
-        "djangorestframework~=3.12.2",
-        "django-filter~=2.4.0",
-        "django-guardian~=2.3.0",
```
dblenkus, line 56: It would be nicer to move this line into the last commit.

gregorjerse: Ah, sure.

resolwe/flow/executors/startup_processing_container.py
gregorjerse: The comment is wrong, fixed.

tests/settings.py
dblenkus: Either of them is fine, but I would go with the second one to follow Django.

dblenkus (approved):

tests/settings.py
gregorjerse: Sure, that is better in the long run.

resolwe/process/models.py
```diff
@@ -394,6 +395,7 @@ def _get_field_data(self, field: Field) -> Any:
 
         :raises RuntimeError: when no data is received.
         """
+        print("Getting field data for field", field)
```
dblenkus, line 398: Remove the print.

PR #907, opened 2022-02-18
Get temporary credentials

gregorjerse:


resolwe/storage/views.py
```diff
@@ -19,12 +19,31 @@
...
+class TemporaryUploadCredentials(ViewSet):
+    """
+    Get the temporary credentials.
+    """
+
+    def list(self, request):
+        """Return the JSON representing temporary credentials."""
+        upload_connector = connectors.for_storage("upload")[0]
+        return Response(
+            upload_connector.temporary_credentials(str(request.user.id)),
```
dblenkus, line 42: I think it is not the best idea to use the user's ID for the unique prefix. What would happen if someone uploads two files with the same name?

resolwe/storage/connectors/s3connector.py
```diff
@@ -364,3 +371,59 @@ def presigned_url(self, url, expiration=60, force_download=False):
...
+        )
+        result = self.original_session.client("sts").assume_role(
+            RoleArn=role_arn or self.config["role_arn"],
+            RoleSessionName="upload-" + prefix,
+            DurationSeconds=duration,
+            Policy=policy,
+        )
+        result["bucket_name"] = self.bucket_name
+        result["region"] = self.config.get("region_name")
+        return result
```
dblenkus, line 429: How will the client know credentials for which service are this? It would be nice to also include a service identifier and any other data the client will need to perform the upload.

dblenkus (commented):
Don't forget to add change log.

resolwe/storage/connectors/s3connector.py
```diff
@@ -364,3 +371,59 @@ def presigned_url(self, url, expiration=60, force_download=False):
...
+
+        :raises KeyError: when no role_arn is gives as argument or connector
+            setting.
+
+        :returns: the dictionary with credentials, see
+        https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sts.html#STS.Client.assume_role
+        for detailed description.
+        """
+        # Policy to add access only to the objects named '/prefix/*'.
+        policy = (
```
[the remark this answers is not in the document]
gregorjerse: Good point, much cleaner.

resolwe/storage/views.py
gregorjerse: It depends on the implementation in the web browser.

I believe it is better to namespace permissions (so nothing from other users can be effected) and that web browser just prefixes a random filename with user id.

resolwe/storage/connectors/s3connector.py
gregorjerse: From the endpoint, which is meant exclusively for upload credentials and the fact we are using only AWS S3 buckets.

Were you thinking about adding some identifier so it would be clear the credentials are for AWS S3 service?

resolwe/storage/views.py
```diff
@@ -19,12 +20,55 @@
...
+        - credentials: the dictionary representing the set of credentials that
+          are used to upload data. This dictionary is specific to connector
+          type and may be empty.
+        """
+        connector_type_mapping = {
+            AwsS3Connector: "S3",
+            LocalFilesystemConnector: "local",
+        }
+        upload_connector = connectors.for_storage("upload")[0]
+        connector_type = connector_type_mapping[upload_connector.__class__]
```
dblenkus, line 60: Can you add the type as an attribute to the connector class?

dblenkus (approved):

resolwe/storage/connectors/s3connector.py
```diff
@@ -364,3 +371,56 @@ def presigned_url(self, url, expiration=60, force_download=False):
...
+                {
+                    "Sid": "AllowAllS3ActionsInUserFolder",
+                    "Effect": "Allow",
+                    "Action": ["s3:PutObject", "s3:AbortMultipartUpload"],
+                    "Resource": [f"arn:aws:s3:::{self.bucket_name}/{prefix}/*"],
+                },
+                {
+                    "Sid": "AllowUserKMSUsage",
+                    "Effect": "Allow",
+                    "Action": ["kms:*"],
```
dblenkus, line 416: List all required KMS permissions instead of using an asterisk.

resolwe/storage/connectors/s3connector.py
```diff
@@ -364,3 +371,56 @@ def presigned_url(self, url, expiration=60, force_download=False):
...
+                    "Resource": ["*"],
```
dblenkus, line 417: It would be great to give access only to a single key.

PR #912, opened 2022-03-09
Create nicer slug colision test

gregorjerse:
using mock object to mock pre_save method.

### DIFF
--- a/resolwe/flow/tests/test_executors.py
+++ b/resolwe/flow/tests/test_executors.py
@@ -6,18 +6,18 @@ import subprocess
 import sys
 import threading
 import unittest
-from concurrent.futures import ThreadPoolExecutor
 from contextlib import suppress
 from pathlib import Path
 from time import sleep
 from unittest import mock
+from unittest.mock import patch
 
 import zmq
 import zmq.asyncio
 from asgiref.sync import async_to_sync
 
 from django.conf import settings
-from django.db import connection, transaction
+from django.db import IntegrityError
 from django.test import override_settings
 
 from resolwe.flow.executors.prepare import BaseFlowExecutorPreparer
@@ -25,8 +25,10 @@ from resolwe.flow.executors.socket_utils import Message
 from resolwe.flow.executors.startup_processing_container import ProcessingManager
 from resolwe.flow.executors.zeromq_utils import ZMQCommunicator
 from resolwe.flow.managers import manager
+from resolwe.flow.managers.dispatcher import Manager
 from resolwe.flow.managers.utils import disable_auto_calls
 from resolwe.flow.models import Data, DataDependency, Process, Worker
+from resolwe.flow.models.fields import ResolweSlugField
 from resolwe.storage import settings as storage_settings
 from resolwe.test import (
     ProcessTestCase,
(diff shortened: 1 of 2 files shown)


resolwe/flow/tests/test_executors.py
```diff
@@ -344,31 +347,34 @@ def test_slug_colision(self):
         This checks for a bug in permission architecture that caused exception
         due to missing permission_group when slug colision occured.
         """
-        process = Process.objects.filter(slug="test-min").latest()
+        def pre_save(instance, add):
+            current_slug = slugs.pop(0)
+            if current_slug == IntegrityError:
```
dblenkus, line 352: It would be a bit more general to check whether the current slug is a subclass of `Exception`.

resolwe/flow/tests/test_executors.py
```diff
@@ -334,8 +336,9 @@ def test_run_in_docker(self):
         data = self.run_process("test-docker")
         self.assertEqual(data.output["result"], "OK")
 
-    @with_null_executor
-    def test_slug_colision(self):
+    @patch.object(Manager, "communicate")
+    @patch.object(ResolweSlugField, "pre_save")
+    def test_slug_colision(self, slug_field_mock, _):
```
dblenkus, line 341: Update the doc string.

resolwe/flow/tests/test_executors.py
```diff
@@ -344,31 +347,34 @@ def test_slug_colision(self):
...
+        slugs = ["process_slug", "first_data_slug", IntegrityError, "second_data_slug"]        
+        slug_field_mock.side_effect = pre_save
+
+        expected_slug = slugs[0]
+        process = Process.objects.create(
+            name="Dummy process", contributor=self.contributor
+        )
+        self.assertEqual(process.slug, expected_slug)
+
+        expected_slug = slugs[0]
```
dblenkus, line 366: This should probably be `[1]`?

resolwe/flow/tests/test_executors.py
```diff
@@ -344,31 +347,34 @@ def test_slug_colision(self):
...
+        data1 = Data.objects.create(
+            name="data", process=process, contributor=self.contributor
+        )
+        self.assertEqual(data1.slug, expected_slug)
+
+        expected_slug = slugs[1]  # Skip IntegrityError
```
dblenkus, line 372: And this `[3]`?

dblenkus (commented):
You can further simplify the test by mocking the method after creating the process.

resolwe/flow/tests/test_executors.py
gregorjerse: Nope, since the pre_save mock method actually removes the entries from the slugs list.

resolwe/flow/tests/test_executors.py
gregorjerse: Same as above. This one is 1 since there are two elements remaining: IntegrityError and the last slug.

resolwe/flow/tests/test_executors.py
gregorjerse: Sure.

resolwe/flow/tests/test_executors.py
gregorjerse: Done.

dblenkus (approved):

PR #913, opened 2022-03-11
WIP: Update migrations

gregorjerse:
Since table flow_storage contains a lot of data a migration to big int causes
errors on AWS RDS due to storage shortage. The new migration procedure
addresses that.


gregorjerse:
Let us wait until the migrations bug is backported to 3.2.12 or we upgrade Django to 4.x branch.

dblenkus (approved):

gregorjerse:
We use Django 4.2 now.

PR #914, opened 2022-03-14
Revert "Migrate id fields to BigAutoField"

gregorjerse:
This reverts commit aa7cab7414c75fd88134b86270736f9917664ff2.


resolwe/process/models.py
```diff
@@ -57,7 +57,6 @@ def to_output(self, value):
 
 
 FIELDS_MAP = {
-    "BigAutoField": IntegerField,
```
dblenkus, line 60: You can leave this in here.

gregorjerse: Thanks, nice catch.

PR #916, opened 2022-03-16
WIP: make it possible to extend process from another one

gregorjerse:
in the same file

### DIFF
--- a/resolwe/process/parser.py
+++ b/resolwe/process/parser.py
@@ -8,7 +8,7 @@ from inspect import isclass
 import asteval
 
 from . import runtime
-from .descriptor import Persistence, ProcessDescriptor, SchedulingClass
+from .descriptor import Persistence, ProcessDescriptor, SchedulingClass, ValidationError
 from .fields import get_available_fields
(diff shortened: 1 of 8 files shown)


resolwe/process/parser.py
```diff
@@ -111,6 +111,11 @@ def __init__(self, source):
         self.source = source
         self.processes = []
         self.base_classes = set()
+        # The defined classes list is a list of all processes defined in the
+        # given file we have seen so far.
+        # This allows us to define processes that inherit from other processes
+        # defined it the same file.
```
jkokosar, line 117: defined _in_

resolwe/process/parser.py
```diff
@@ -251,6 +265,9 @@ def base_classes(self):
 
         :return: A list of the base classes for the processes.
         """
+
+        print(self._source)
```
jkokosar, line 269: Remove the print

jkokosar (commented):
Add CHANGELOG entry

gregorjerse:
A different solution is needed.

PR #920, opened 2022-04-01
Do not return multiple versions of the same process

gregorjerse:
when checking permissions in python process


dblenkus (approved):
Add a change log.

PR #922, opened 2022-04-08
Fix misleading error message on file import

gregorjerse:
if status code indicating error was received from the upstream server

### DIFF
--- a/resolwe/process/fields.py
+++ b/resolwe/process/fields.py
@@ -624,9 +624,9 @@ class FileDescriptor:
                 else file_name
             )
 
-        # Large file download from Google Drive requires cookie and token.
         try:
             response = None
+            # Large file download from Google Drive requires cookie and token.
             if re.match(
                 r"^https://drive.google.com/[-A-Za-z0-9\+&@#/%?=~_|!:,.;]*[-A-Za-z0-9\+&@#/%=~_|]$",
                 src,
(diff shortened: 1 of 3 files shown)


resolwe/process/fields.py
```diff
@@ -624,13 +624,14 @@ def importUncompressed():
...
 
-        # Large file download from Google Drive requires cookie and token.
         try:
             response = None
+            # Large file download from Google Drive requires cookie and token.
             if re.match(
                 r"^https://drive.google.com/[-A-Za-z0-9\+&@#/%?=~_|!:,.;]*[-A-Za-z0-9\+&@#/%=~_|]$",
                 src,
             ):
+                print("Importing file from google drive.")
```
dblenkus, line 634: Remove print.

resolwe/process/fields.py
```diff
@@ -648,12 +649,14 @@ def importUncompressed():
                 r"^(https?|ftp)://[-A-Za-z0-9\+&@#/%?=~_|!:,.;]*[-A-Za-z0-9\+&@#/%=~_|]$",
                 src,
             ):
+                print("Importing file from http/ftp server")
```
dblenkus, line 652: Remove print.

resolwe/process/fields.py
gregorjerse: Done.

resolwe/process/fields.py
gregorjerse: Done.

dblenkus (approved):

PR #925, opened 2022-04-15
Send custom signal on bulk data create

gregorjerse:
Django signals are not triggered when bulk_create method is used.


dblenkus (commented):
Add a change log and tests.

dblenkus (commented):
It would be nice to follow Django's practice and name signals `pre_*` or `post_*`. Also, keep the terminology unified and use "duplicate" instead of "copy". So this signal should be `post_duplicate`.

dblenkus (approved):

PR #927, opened 2022-05-05
Build container also for arm64 architechure

gregorjerse:


dblenkus:
Closed in favor of #928.

PR #928, opened 2022-05-05
Build container also for arm64 architechure (branch)

gregorjerse:


gregorjerse:
Related to EXP-338 .

gregorjerse:
EXP-338

.github/workflows/docker.yml
```diff
@@ -38,51 +39,27 @@ jobs:
             suffix: ""
 
     steps:
-      - uses: actions/checkout@v2
```
dblenkus, line 41: Is there a reason why you moved these three actions after the login to ECR?

gregorjerse: Yes and no. While testing, the login to ECR failed a couple of times, and wanted to do it first to speed up the development.
I can change the order to the old one if necessary.

.github/workflows/docker.yml
```diff
@@ -52,37 +56,23 @@ jobs:
...
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@v3
         with:
           context: ${{ matrix.context }}
           file: ${{ matrix.file }}
-          tags: ${{ env.TAGS }}
-          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          platforms: linux/amd64,linux/arm64          
+          push: false
```
dblenkus, line 78: Remove this, and then we can merge it.

acopar (approved):

PR #932, opened 2022-06-07
Support custom user model in Python processes

gregorjerse:


AnzeLovse (commented):
Looks OK although I might be missing some details.

dblenkus (approved):

PR #939, opened 2022-07-17
Feature/test runner

gregorjerse:


resolwe/process/parser.py
```diff
@@ -111,9 +113,66 @@ def __init__(self, source):
...
+        # argument, so no separation is necessary.
+        argument_names = ("slug", "inputs")
+        arguments = [arg for arg in called_node.args]
+        all_arguments = {
+            name: value for name, value in zip_longest(argument_names, arguments)
+        }
+        all_arguments.update(
+            {keyword.arg: keyword.value for keyword in called_node.keywords}
+        )
+        assert "slug" in all_arguments, "The 'slug' is required argument"
```
dblenkus, line 137: This should be handled in process validation, not in a test runner. I suggest setting `slug` to `None` if it is not given and letting the code raise the `RuntimeError`.

resolwe/process/parser.py
```diff
@@ -111,9 +113,66 @@ def __init__(self, source):
...
+
+        :raises RuntimeError: when slug could not be determined (for instance when
+            the slug is being a variable).
+        """
+        # Both methods (run_process and get_latest) take argument slug as the first
+        # argument, so no separation is necessary.
+        argument_names = ("slug", "inputs")
+        arguments = [arg for arg in called_node.args]
+        all_arguments = {
+            name: value for name, value in zip_longest(argument_names, arguments)
```
dblenkus, line 132: This will work in most use-cases, but note that `run_process(inputs={...}, slug='...')` is also a valid syntax.

resolwe/process/parser.py
```diff
@@ -111,9 +113,66 @@ def __init__(self, source):
...
+        }
+        all_arguments.update(
+            {keyword.arg: keyword.value for keyword in called_node.keywords}
```
dblenkus, line 135: Is there any benefit of retrieving all parameters? It looks like returning the first element would be sufficient for this function (actually, as pointed out in the comment above, checking if `slug` is a keyword argument and returning the first argument otherwise).

resolwe/process/parser.py
```diff
@@ -111,9 +113,66 @@ def __init__(self, source):
...
+        :raises RuntimeError: when dependencies could not be determined. For
+            instance when process slug is a variable.
+
+        :returns: the dictionary where key is the process slug and the
+            corresponding value is the list of processes the process can spawn.
+        """
+        slugs = dict()
+        candidate_names = (
+            "Process.get_latest",
+            "BioProcess.get_latest",
```
dblenkus, line 158: This is an implementation detail of `resolwe_bio` and can be different for each package of processes.

resolwe/test_helpers/test_runner.py
```diff
@@ -583,57 +591,92 @@ def find_schemas(self, schema_path):
...
+            "python": lambda file_handler: [file_handler.read()],
+            "yaml": partial(yaml.load, Loader=yaml.FullLoader),
+        }
+        changed_schemas = {"python": [], "yaml": []}
+        schema_file_suffixes = {"python": (".py",), "yaml": (".yml", ".yaml")}
 
         for root, _, files in os.walk(schema_path):
             for schema_file in [os.path.join(root, fn) for fn in files]:
+                # Check for processes of both types: python and yaml.
+                for schema_type in ("python", "yaml"):
```
dblenkus, line 604: Use `schema_loader.keys()`.

resolwe/process/parser.py
gregorjerse: OK.

resolwe/process/parser.py
gregorjerse: Will this not work as expected? The zip_longest will be empty (only keyword arguments are given) and all_arguments will be empty after the first call, but the next one will update it with the keyword arguments.

resolwe/process/parser.py
gregorjerse: You are right. I have modified the method.

resolwe/process/parser.py
gregorjerse: I used FLOW_PROCESS_RUNTIMES settings where the registered runtimes are stored.

resolwe/test_helpers/test_runner.py
gregorjerse: Done.

dblenkus (approved):

PR #941, opened 2022-07-26
Add new model observers

plojyon:
Users can subscribe to observe model changes and will receive
real-time update notifications via a WebSocket.


resolwe/observers/models.py
```diff
@@ -0,0 +1,82 @@
...
+    """State of a model observer."""
+
+    CHANGE_TYPES = (
+        (CHANGE_TYPE_CREATE, "create"),
+        (CHANGE_TYPE_UPDATE, "update"),
+        (CHANGE_TYPE_DELETE, "delete"),
+    )
+
+    # Table of the observed resource
+    table = models.CharField(max_length=100)
```
dblenkus, line 26: Reference table by content type instead of the name as this is a more common approach in Django. Check how [GenericForeignKey](https://docs.djangoproject.com/en/4.0/ref/contrib/contenttypes/#django.contrib.contenttypes.fields.GenericForeignKey) does it.

resolwe/observers/models.py
```diff
@@ -0,0 +1,82 @@
...
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+    )
+    created = models.DateTimeField(auto_now_add=True)
+
+    # ID of the websocket session (can have multiple observers)
+    session_id = models.CharField(max_length=100)
+    # Unique ID for the client to remember which subscription a signal belongs to
+    subscription_id = models.CharField(
```
dblenkus, line 63: Use [UUIDField](https://docs.djangoproject.com/en/4.0/ref/models/fields/#uuidfield).

resolwe/observers/models.py
```diff
@@ -0,0 +1,82 @@
...
+        for id in resource_ids:
+            for change_type in change_types:
+                observer, _ = Observer.objects.get_or_create(
+                    table=table, resource_pk=id, change_type=change_type
+                )
+                self.observers.add(observer)
+
+    def delete(self):
+        """Clean up observers with no subscriptions before deleting self."""
+        for observer in self.observers.annotate(subs=Count("subscriptions")).filter(
```
dblenkus, line 78: This is not an atomic operation, so it can cause strange errors. Make sure to execute it as a single operation:
```py
self.observers.annotate(subs=Count("subscriptions")).filter(subs=1).delete()
```

resolwe/observers/mixins.py
```diff
@@ -0,0 +1,55 @@
...
+    Adds the /subscribe and /unsubscribe endpoints to the list view.
+    """
+
+    def _id_exists_for_user(self, id, user):
+        return self.get_queryset().filter_for_user(user).filter(pk=id).exists()
+
+    @action(detail=False, methods=["post", "get"])
+    def subscribe(self, request):
+        """Register an Observer for a resource."""
+        ids = dict(request.query_params).get("ids", [None])
```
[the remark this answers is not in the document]
plojyon: Is there a reason why `None` would be better? I expect `ids` to be a list of ids (when they're actually provided), so `[None]` makes more sense to me.
The `subscribe` method later accepts a list of ids to subscribe to, and subscribing to id=None counts as subscribing to the whole table.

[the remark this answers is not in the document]
dblenkus: As you already mentioned, `ids` is a list of ids, but there is no object with id `None`. You can use `[]` instead if it makes more sense.

resolwe/permissions/models.py
```diff
@@ -239,6 +241,21 @@ def get_permission(self, user_or_group: UserOrGroup) -> Permission:
...
             return Permission.NONE
 
+    def users_with_permission(self, permission: Permission) -> List[User]:
+        """Get a list of users with at least this permission level.
+
+        Calling this with Permission.NONE will return users for whom an explicit
+        PermissionModel with Permission.NONE exists.
+        """
+        filtered = self.permissions.filter(value__gte=permission)
+        users = set(
```
dblenkus, line 251: There is no need to transform the result into a set.

resolwe/observers/consumers.py
```diff
@@ -0,0 +1,72 @@
...
+        if change_type == CHANGE_TYPE_DELETE:
+            # The observed object was either deleted or the user lost permissions.
+            subscription = Subscription.objects.get(session_id=self.session_id)
+            observers = Observer.objects.filter(
+                content_type=content_type,
+                object_id=object_id,
+                # change_type = Any,
+            )
+            # Assure we don't stay subscribed to an illegal object.
+            for observer in observers:
```
dblenkus, line 62:
```py
subscription.observers.remove(*observers)
```

resolwe/observers/consumers.py
```diff
@@ -0,0 +1,72 @@
...
+            ),
+        )
+
+        if change_type == CHANGE_TYPE_DELETE:
+            # The observed object was either deleted or the user lost permissions.
+            subscription = Subscription.objects.get(session_id=self.session_id)
+            observers = Observer.objects.filter(
+                content_type=content_type,
+                object_id=object_id,
+                # change_type = Any,
```
dblenkus, line 59: Remove the comment.

resolwe/observers/consumers.py
```diff
@@ -0,0 +1,72 @@
...
+        """Handle an item update signal."""
+        content_type = ContentType.objects.get_for_id(msg["content_type_pk"])
+        object_id = msg["object_id"]
+        change_type = msg["change_type"]
+
+        interested = Observer.get_interested(
+            content_type=content_type, object_id=object_id, change_type=change_type
+        )
+        subscription_ids = map(
+            lambda x: x.hex,
```
dblenkus, line 44: Rename `x` to `subscription_id`.

resolwe/observers/mixins.py
```diff
@@ -0,0 +1,59 @@
...
+
+class ObservableMixin:
+    """A Mixin to make a model ViewSet observable.
+
+    Adds the /subscribe and /unsubscribe endpoints to the list view.
+    """
+
+    def user_has_permission(self, id, user):
+        """Verify that an object exists for a given user."""
+        return self.get_queryset().filter_for_user(user).filter(pk=id).exists()
```
dblenkus, line 22: Have you compared filtering orders (`.filter_for_user(user).filter(pk=id)` and `.filter(pk=id).filter_for_user(user)`? Do they produce the same database query?

plojyon: The query is similar enough, but apparently `filter_for_user` performs some subqueries in the background, so calling `filter` first should be better.

resolwe/observers/models.py
```diff
@@ -0,0 +1,144 @@
...
+    class Meta:
+        """Meta parameters for the Subscription model."""
+
+        indexes = [
+            models.Index(name="idx_subscription_session_id", fields=["session_id"]),
+        ]
+
+    def subscribe(self, content_type, resource_ids, change_types):
+        """Assign self to multiple observers at once."""
+        for id in resource_ids:
```
gregorjerse, line 146: I think it would be best to stick to the same terminology: if you are using `object_id` in the observer, use it also here (instead of using `resource_id`).

resolwe/observers/models.py
```diff
@@ -0,0 +1,144 @@
...
+            for change_type in change_types:
+                observer, _ = Observer.objects.get_or_create(
+                    content_type=content_type, object_id=id, change_type=change_type
+                )
+                self.observers.add(observer)
+
+    def delete(self):
+        """Clean up observers with no subscriptions before deleting self."""
+        self.observers.annotate(subs=Count("subscriptions")).filter(subs=1).delete()
```
gregorjerse, line 155: Is the correct filter`subs=1` or `subs=0`?
If the first one is the correct one, add some explanation.

resolwe/observers/models.py
```diff
@@ -0,0 +1,144 @@
...
+        """Assign self to multiple observers at once."""
+        for id in resource_ids:
+            for change_type in change_types:
+                observer, _ = Observer.objects.get_or_create(
+                    content_type=content_type, object_id=id, change_type=change_type
+                )
+                self.observers.add(observer)
+
+    def delete(self):
+        """Clean up observers with no subscriptions before deleting self."""
```
gregorjerse, line 154: I would prefer comments like

"""
Delete the given subscription.

Also, delete all observers to which only we are subscribed.
"""

The main purpose of this method is to delete subscriptions and deleting observers is an additional feature.

resolwe/observers/models.py
```diff
@@ -0,0 +1,144 @@
...
+        self.observers.annotate(subs=Count("subscriptions")).filter(subs=1).delete()
+        super().delete()
+
+    @classmethod
+    def notify(cls, session_id, instance, change_type):
+        """Register a callback to send a change notification on transaction commit."""
+        notification = {
+            "type": TYPE_ITEM_UPDATE,
+            "content_type_pk": ContentType.objects.get_for_model(type(instance)).pk,
```
gregorjerse, line 163: Remove `type`, it is not needed.

resolwe/observers/signals.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # Create signals will be caught when the PermissionModel is added.
+    if created:
+        return
+
+    Observer.observe_instance_changes(instance, CHANGE_TYPE_UPDATE)
+
+
+@dispatch.receiver(model_signals.pre_delete)
+def observe_model_deletion(sender, instance, **kwargs):
```
gregorjerse, line 75: Is it possible that objects is not really deleted after pre_delete is triggered?

resolwe/observers/consumers.py
```diff
@@ -0,0 +1,70 @@
...
+from .protocol import CHANGE_TYPE_DELETE, GROUP_SESSIONS
+
+
+class ClientConsumer(JsonWebsocketConsumer):
+    """Consumer for client communication."""
+
+    def websocket_connect(self, message):
+        """Handle establishing a WebSocket connection."""
+        session_id = self.scope["url_route"]["kwargs"]["session_id"]
+        self.session_id = session_id
```
gregorjerse, line 17: Is there one consumer instance per client connection or one consumer instance for all clients?

If the second is true, then `self.session_id` (which is client-specific) will always point to the session id of the client that performed the last connection.

resolwe/observers/consumers.py
```diff
@@ -0,0 +1,70 @@
...
+    def observers_item_update(self, msg):
+        """Handle an item update signal."""
+        content_type = ContentType.objects.get_for_id(msg["content_type_pk"])
+        object_id = msg["object_id"]
+        change_type = msg["change_type"]
+
+        interested = Observer.get_interested(
+            content_type=content_type, object_id=object_id, change_type=change_type
+        )
+        subscription_ids = map(
```
gregorjerse, line 43: Why not just use
```
subscription_ids = [ subscription_id.hex for subscription_id in Subscription.objects.filter(observers__in=interested)... ]
```

gregorjerse (changes requested):

resolwe/observers/consumers.py
plojyon: The first is true.

resolwe/observers/signals.py
plojyon: Yes, if the transaction is rolled back. That's why the actual notification is sent through `transaction.on_commit`, but that code lies two frames deeper, in the Subscription model.
https://github.com/genialis/resolwe/blob/23b4313e97b6e88d5a38a8dcfa0633ab0a0b7e5d/resolwe/observers/models.py#L176
The reason is that we're trying to collect information about the to-be-deleted object before it's actually deleted.

resolwe/observers/models.py
```diff
@@ -0,0 +1,144 @@
...
+)
+
+
+def get_random_uuid():
+    """Generate a random UUID in string format."""
+    return uuid.uuid4().hex
+
+
+class Observer(models.Model):
+    """State of a model observer."""
```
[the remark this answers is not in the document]
plojyon: It's kind of an abstract concept, so I wasn't really sure what to write. How about this:

    An observer for detecting changes in a model.

    Characterized by a db table or an instance of a model and a change type.
    Several subscriptions can subscribe to the same observer.

resolwe/observers/models.py
```diff
@@ -0,0 +1,165 @@
...
+    ) -> "QuerySet[Observer]":
+        """Find all observers watching for changes of a given item/table."""
+        query = Q(content_type=content_type)
+        if change_type is not None:
+            query &= Q(change_type=change_type)
+        query &= Q(object_id=object_id) | Q(object_id__isnull=True)
+        return cls.objects.filter(query)
+
+    @classmethod
+    def observe_instance_changes(cls, instance: any, change_type: str):
```
gregorjerse, line 71: You probably could use at least ``django.db.models.Model`` for the instance base class.

resolwe/observers/models.py
```diff
@@ -0,0 +1,165 @@
...
+
+        observers = Observer.get_interested(
+            change_type=change_type,
+            content_type=ContentType.objects.get_for_model(instance),
+            object_id=instance.pk,
+        )
+
+        # Forward the message to the appropriate groups.
+        for subscriber in Subscription.objects.filter(observers__in=observers):
+            if not instance.has_permission(Permission.VIEW, subscriber.user):
```
gregorjerse, line 82: Negate the condition .

```
if instance.has_permission(Permission.VIEW, subscriber.user):
  # Register on_commit callbacks to send the signals.
  Subscription.notify(subscriber.session_id, instance, change_type)
```

gregorjerse (changes requested):

resolwe/observers/mixins.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+
+    def user_has_permission(self, id: int, user: User) -> bool:
+        """Verify that an object exists for a given user."""
+        return self.get_queryset().filter(pk=id).filter_for_user(user).exists()
+
+    @action(detail=False, methods=["post"])
+    def subscribe(self, request: Request) -> Response:
+        """Register an Observer for a resource."""
+        ids = dict(request.query_params).get("ids", None)
```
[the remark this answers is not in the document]
plojyon: Getting values from a QueryParams object discards all but the first value. Example query:
```
?session_id=2&ids=1&ids=2&ids=3
```
produces
```
dict(query_params) -> {"session_id": [2], "ids": [1, 2, 3]}
query_params -> {"session_id": 2, "ids": 1}
```
In the instance of session_id, we're not expecting mutlitple values, so calling `get` on QueryParams is fine. 
This would also work for the second line:
`dict(request.query_params).get("session_id")[0]`
but it would throw a (more vague) IndexError in case session_id was missing.

PR #954, opened 2022-10-22
Processing stability / listener multiple

gregorjerse:


resolwe/flow/managers/listener/listener.py
```diff
@@ -69,22 +69,155 @@
...
 
 
-class Processor:
-    """Class respresents instance of one connected worker."""
+class RedisCache:
+    """The Redis cache."""
+
+    cached_fields_type = {
+        "id": int,
+        "slug": str,
```
dblenkus, line 69: This can change.

resolwe/flow/managers/listener/listener.py
```diff
@@ -69,22 +69,155 @@
...
+        "contributor": int,
+        "contributor_id": int,
+        "status": str,
```
dblenkus, line 72: This can change.

resolwe/flow/managers/listener/listener.py
```diff
@@ -69,22 +69,155 @@
...
+        "started": datetime,
+        "finished": datetime,
```
dblenkus, line 74: This can change if the current value is `None`.

resolwe/flow/managers/listener/listener.py
```diff
@@ -69,22 +69,155 @@
 logger = logging.getLogger(__name__)
 
 
-class Processor:
-    """Class respresents instance of one connected worker."""
+class RedisCache:
+    """The Redis cache."""
+
+    cached_fields_type = {
```
dblenkus, line 67: It would be nice to cache the outputs of finished processes.

resolwe/flow/managers/listener/listener.py
```diff
@@ -69,22 +69,155 @@
...
+
+        The query is run in a transaction.
+        """
+
+        def get_redis_data(pipeline: redis.client.Pipeline):
+            def get_redis_field(redis_key: str, field_name: str) -> Any:
+                """Get the redis data from the specific key."""
+                field_type = self.cached_fields_type[field_name]
+
+                if field_type in (int, str, datetime):
```
dblenkus, line 130: It would be nice to create a wrapper class for each field tye and move CRUD operations there.

resolwe/flow/executors/__main__.py
```diff
@@ -58,7 +58,7 @@ async def open_listener_connection(data_id, host, port, protocol) -> ZMQCommunic
     """Connect to the listener service."""
     zmq_context = zmq.asyncio.Context.instance()
     zmq_socket = zmq_context.socket(zmq.DEALER)
-    zmq_socket.setsockopt(zmq.IDENTITY, f"e_{data_id}".encode())
+    zmq_socket.setsockopt(zmq.IDENTITY, f"-{data_id}".encode())
```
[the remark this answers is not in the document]
gregorjerse: This code is mostly necessary for testing and slurm deploy. 

For listener to differentiate between connected peers they need a different ID.  It is quite fine that init container and communication container use the same ID, since they are not active at the same moment.

On slurm/testing, the executor starts on a possibly different machine, connects to the listener and bootstraps the docker container.  The executor also sends logs to the listener (how much time it took to start the docker container, report exceptions etc) and can therefore be active at the same time as containers. So it must use a different ID. Before I used `e_{data_id}`, but having minus makes code in the listener nicer.

resolwe/flow/executors/logger.py
```diff
@@ -40,8 +40,7 @@ def __init__(self, communicator, **kwargs):
     def emit(self, record):
         """Send log message to the listener."""
         asyncio.run_coroutine_threadsafe(
-            self.communicator.send_command(Message.command("log", self.format(record))),
-            self.loop,
+            self.communicator.log(self.format(record)), self.loop
```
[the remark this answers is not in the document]
gregorjerse: Done.

resolwe/flow/managers/listener/listener.py
gregorjerse: Sure. But can the change be made by something that is not listener? 
The listener uses method for saving changes that also updates the Redis cache.

resolwe/flow/managers/listener/listener.py
gregorjerse: I agree. When they are used in a pipeline, the output of already finished process is very likely to be used in the near future. But this would require changes in the RedisCache since output can be deeply nested and use various types for values.

How about using lru_cache to store the last 1000 outputs for example? The cache will be local to every listener instance, but there will probably not be a lot of them and the output of already finished processes can not change.

resolwe/flow/managers/listener/listener.py
gregorjerse: I think the code is simple and it serves the scope for which it is used. If the cache will be used for more than just caching some data values than I agree.

resolwe/flow/managers/listener/listener.py
dblenkus: You are right.

dblenkus (approved):

PR #972, opened 2022-12-22
Add set_permission to PermissionQuerySet

gregorjerse:


resolwe/permissions/models.py
```diff
@@ -362,6 +362,20 @@ def filter_for_user(
             user, groups, permission, public, with_superuser
         )
 
+    def set_permission(self, permission: Permission, user_or_group: UserOrGroup):
+        """Set the permission on the objects in the queryset."""
+        from resolwe.permissions.utils import get_identity  # Circular import
+
+        entity, entity_name = get_identity(user_or_group)
```
dblenkus, line 369: I prefer `entity_type` over `entity_name`. And it would be great to find another name instead of "entity" to avoid confusion.

dblenkus (approved):

resolwe/permissions/models.py
gregorjerse: I have changed `entity_name` to `entity_type`.

The better collective naming for `users` and `groups` still eludes me.

PR #984, opened 2023-02-03
Fix bug in superuser notification

gregorjerse:
superusers  were never notified when observable object permissions were changed since they always have all the permissions on every object.

The notifications are therefore sent to admins and regular users separately when object is created using the API.

### DIFF
--- a/resolwe/flow/views/mixins.py
+++ b/resolwe/flow/views/mixins.py
@@ -46,17 +46,13 @@ class ResolweCreateModelMixin(mixins.CreateModelMixin):
         """Create a resource."""
         with transaction.atomic():
             instance = serializer.save()
+            # Notify the users without explicit permissions (superusers).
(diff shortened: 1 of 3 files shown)


resolwe/flow/views/mixins.py
```diff
@@ -46,17 +46,13 @@ def perform_create(self, serializer):
         """Create a resource."""
         with transaction.atomic():
             instance = serializer.save()
+            # Notify the users without explicit permissions (superusers).
+            post_permission_changed.send(sender=type(instance), instance=instance)
```
dblenkus, line 50: Will this send double notifications if the contributor is also an administrator?

gregorjerse: It should not. Even if permission for admin are explicit, there was no change in permission level: first it was READ granted by implicit grant and afterwards by explicit permission entry in the database.

gregorjerse:
Deprecated due to #986 .

PR #993, opened 2023-03-20
Fix script retrieval timeout & replace asyncio loop with uvloop

gregorjerse:
EXP-551
EXP-593
EXP-607
EXP-612

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,6 +64,7 @@ setuptools.setup(
         "Sphinx~=4.3.2",
         "wrapt~=1.13.3",
         "pyzmq~=22.3.0",
+        "uvloop~=0.17.0",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -64,6 +64,7 @@
         "Sphinx~=4.3.2",
         "wrapt~=1.13.3",
         "pyzmq~=22.3.0",
+        "uvloop~=0.17.0",
```
dblenkus, line 67: You should add the requirement in the same commit as the code where you use it. That way, figuring out why the requirement was added is much easier.

gregorjerse: Fixed.

dblenkus (approved):

PR #995, opened 2023-03-27
Enable setting affinity for interactive/batch processes

gregorjerse:
EXP-565
EXP-567

### DIFF
--- a/resolwe/flow/managers/workload_connectors/kubernetes.py
+++ b/resolwe/flow/managers/workload_connectors/kubernetes.py
@@ -537,6 +537,32 @@ class Connector(BaseConnector):
             else kubernetes.config.load_kube_config()
         )
 
+    def optimize_job_scheduling(self, data: Data, job_description: dict):
+        """Optimize the scheduling by modifying the job description.
+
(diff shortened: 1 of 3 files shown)


resolwe/flow/managers/workload_connectors/kubernetes.py
```diff
@@ -537,6 +537,34 @@ def _load_kubernetes_config(self):
             else kubernetes.config.load_kube_config()
         )
 
+    def optimize_job_scheduling(self, data: Data, job_description: dict):
+        """Optimize the scheduling by modifying the job description.
+
+        Moodify the job description to be a better fit on the scheduler. Currently this
+        is only used to schedule interactive/non-interactive processes.
```
dblenkus, line 544: Scheduling for interactive/non-interactive processes is our use-case, but the idea behind it can change without changing the method and the doc string will become outdated. Use "schedule based on the scheduling class of the process" instead.

gregorjerse: Done.

dblenkus (approved):

PR #996, opened 2023-03-27
Make kubernetes overcommitment factors configurable

gregorjerse:
via the settings variable.

EXP-617
EXP-565

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -119,6 +119,12 @@ COMMUNICATION_CONTAINER_LISTENER_CONNECTION = {"local": "172.17.0.1"}
 #: case name of the process scheduling class.
 FLOW_KUBERNETES_AFFINITY = os.environ.get("RESOLWE_KUBERNETES_AFFINITY", None)
 
+# Add the settings to overcommit the CPU and memory based on the process
+# scheduling type. Example:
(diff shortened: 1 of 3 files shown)


tests/settings.py
```diff
@@ -117,7 +117,13 @@
 #:   {"batch": {"nodegroup": ["compute"]}, "interactive": {"nodegroup": ["interactive"]}}
 #: The keys in the scheduling_class dictionary must correspond to the lower
 #: case name of the process scheduling class.
-FLOW_KUBERNETES_AFFINITY = os.environ.get("RESOLWE_KUBERNETES_AFFINITY", None)
+FLOW_KUBERNETES_AFFINITY = None
```
dblenkus, line 120: Why is this changed in this commit?

dblenkus (commented):
Add a change log.

tests/settings.py
gregorjerse: You are right, it deserves a commit of its own. 

I made a stupid mistake to read a JSON encoded variable from environment directly, which would not work. Since we do not really need this in Resolwe I decided to add a default value in the settings file.

dblenkus (approved):

PR #998, opened 2023-03-31
Do not allow null values for object_id

gregorjerse:
EXP-620


resolwe/observers/models.py
```diff
@@ -112,12 +112,13 @@ class Observer(models.Model):
     Several subscriptions can subscribe to the same observer.
     """
 
+    ALL_IDS = 0  # The constant used for catchall object id.
```
dblenkus, line 115: Space is missing in the comment.

dblenkus (approved):

PR #1003, opened 2023-04-12
Expose annotation endpoints and send observer create on duplication

gregorjerse:
BUG-951

### DIFF
--- a/resolwe/api_urls.py
+++ b/resolwe/api_urls.py
@@ -4,6 +4,9 @@ from django.urls import include, path
 from rest_framework import routers
 
 from .flow.views import (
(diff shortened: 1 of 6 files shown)


resolwe/api_urls.py
```diff
@@ -25,6 +28,9 @@
 api_router.register(r"storage", StorageViewSet)
 api_router.register(r"upload_config", UploadConfig, basename="upload_config")
 api_router.register(r"task", BackgroundTaksViewSet, basename="backgroundtask")
+api_router.register(r"annotation_field", AnnotationFieldViewSet)
```
dblenkus, line 31: Add a change log.

PR #1011, opened 2023-05-08
Upgrade to Django 4.2

gregorjerse:


setup.py
```diff
@@ -38,42 +38,42 @@
...
         # Storage requirement for computing hashes.
         "crcmod",
-        "kubernetes~=21.7.0",
+        "kubernetes~=26.1.0",
         "docker~=6.1.1",
-        "Django~=3.2.12",
-        "djangorestframework~=3.13.1",
-        "django-filter~=21.1",
-        "django-versionfield~=1.0.2",
+        "Django~=4.2",
```
dblenkus, line 50: Pin the version to `4.2.0`, otherwise `~=` operator will also install version 4.2 if it will ever exist.

dblenkus (approved):

PR #1021, opened 2023-05-25
wip

gregorjerse:


resolwe/flow/models/annotations.py
```diff
@@ -334,6 +335,9 @@ class AnnotationPreset(AuditModel, PermissionObject):
     #: the fields belongint to this preset
     fields = models.ManyToManyField(AnnotationField, related_name="presets")
 
+    #: user that created the entry
```
dblenkus, line 338: entity -> preset

resolwe/flow/models/annotations.py
```diff
@@ -334,6 +335,9 @@ class AnnotationPreset(AuditModel, PermissionObject):
...
+    contributor = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.PROTECT)
```
dblenkus, line 339: It would make sense to use contributor serializer: https://github.com/genialis/resolwe/blob/b44518856653765ef3c4b4acb6124730a39680c7/resolwe/flow/serializers/base.py#L39

gregorjerse: This one is on a model, not a serializer.
I could use BaseResolweModel as a base but I feel preset does not need all the bells and whistles defined there.

dblenkus (approved):

PR #1022, opened 2023-05-25
Enable filtering entities by annotation values

gregorjerse:


dblenkus (commented):
Add a test for this.

resolwe/flow/filters.py
```diff
@@ -337,6 +339,63 @@ class EntityFilter(BaseCollectionFilter):
...
+            .values_list("type", flat=True)
+            .get()
+        )
+
+        if lookup_type not in allowed_lookups[field_type]:
+            raise ValueError(
+                f"Lookup '{lookup_type}' not supported for the field '{field_id}'."
+            )
+
+        post_processing = post_processing_map.get(field_type, lambda x: x)
```
dblenkus, line 385: Use `None` as default and execute the post processing only when it is defined.

resolwe/flow/filters.py
```diff
@@ -337,6 +339,63 @@ class EntityFilter(BaseCollectionFilter):
...
+        processing = lambda x: post_processing(
```
dblenkus, line 386: Don't use generic variable names.

resolwe/flow/filters.py
gregorjerse: Fixed.

resolwe/flow/filters.py
gregorjerse: Fixed.

dblenkus (approved):

PR #1044, opened 2023-09-08
History tracking

gregorjerse:


setup.py
```diff
@@ -96,6 +96,7 @@
             "isort>=5.12.0",
             "daphne>=4.0.0",
             "django-stubs>=4.2.4",
+            "djangorestframework-stubs[compatible-mypy]>=3.14.2",
```
dblenkus, line 99: This version should probably be the same as `djangorestframework`, so `~=3.14.0`.

resolwe/flow/models/base.py
```diff
@@ -12,6 +12,8 @@
 
 # Maximum number of slug-generation retries.
 MAX_SLUG_RETRIES = 10
+SLUG_LENGTH = 100
```
dblenkus, line 15: Prepend these constants with `MAX_`.

resolwe/flow/models/history.py
```diff
@@ -0,0 +1,193 @@
...
+from django.db import models
+from django.db.backends.postgresql.psycopg_any import DateTimeTZRange
+
+from .base import NAME_LENGTH, SLUG_LENGTH
+from .collection import Collection
+from .data import Data
+from .history_manager import HistoryMixin, history_manager
+
+
+class BaseHistory(models.Model):
```
dblenkus, line 18: Mark this class as abstract: https://docs.djangoproject.com/en/4.2/topics/db/models/#abstract-base-classes

resolwe/flow/models/history_manager.py
```diff
@@ -0,0 +1,142 @@
...
+                    if current_value != new_value:
+                        tracker.process_change(instance, field_name)
+            except RuntimeError:
+                tracker.process_change(instance, field_name)
+
+
+history_manager = HistoryManager()
+
+
+def post_delete_signal(sender, instance, **kwargs):
```
dblenkus, line 106: Move this to the `delete` method.

setup.py
gregorjerse: Not sure: `djangorestframework` has no `3.14.2` release, but unifying them will not hurt.

resolwe/flow/models/history_manager.py
gregorjerse: I was just about to do that when a paragraph

```The delete() method does a bulk delete and does not call any delete() methods on your models. It does, however, emit the [pre_delete](https://docs.djangoproject.com/en/4.2/ref/signals/#django.db.models.signals.pre_delete) and [post_delete](https://docs.djangoproject.com/en/4.2/ref/signals/#django.db.models.signals.post_delete) signals for all deleted objects (including cascaded deletions).```

from Django documentation changed my mind. However, I do not like to depend on signals to much so I will move this to a `delete` method. And keep in mind that bulk delete is not a good idea.

resolwe/flow/models/base.py
gregorjerse: Done.

resolwe/flow/models/history.py
gregorjerse: This is not possible: I have a reference to this class from `SlugHistory` for example. This enables me to use the same class to track history of multiple objects.

dblenkus (approved):

PR #1055, opened 2023-10-09
Add annotations property to entity

gregorjerse:
in Python processes. It can be used as dictionary to read/set entity annotations.


resolwe/flow/models/annotations.py
```diff
@@ -370,6 +370,15 @@ def __str__(self):
 class AnnotationValue(AuditModel):
     """The value of the annotation."""
 
+    class Meta:
```
dblenkus, line 373: This can/should go into a separate commit.

gregorjerse: Done.

dblenkus (approved):

dblenkus (approved):

PR #1069, opened 2023-11-13
Bulk create annotation values

gregorjerse:


resolwe/flow/serializers/annotations.py
```diff
@@ -80,6 +81,13 @@ class AnnotationsSerializer(serializers.Serializer):
     value = serializers.JSONField()
 
 
+class AnnotationsByPathSerializer(serializers.Serializer):
+    """Serializer that reads annotation field and its value."""
+
+    field_path = serializers.CharField(max_length=2 * ANNOTATION_NAME_LENGTH + 1)
```
dblenkus, line 87: Add a number where the number comes from.

resolwe/flow/views/annotations.py
```diff
@@ -75,25 +83,33 @@ class AnnotationValueViewSet(
...
+        if not all(
+            entity.has_permission(Permission.EDIT, request.user) for entity in entities
+        ):
             raise exceptions.PermissionDenied()
-        else:
-            raise exceptions.NotFound()
+        self.perform_create(serializer)
+        headers = self.get_success_headers(serializer.data)
+        return response.Response(
+            serializer.data, status=status.HTTP_201_CREATED, headers=headers
```
dblenkus, line 141: What will be the response for the singe object? The object itself or a list with one element?

resolwe/flow/views/annotations.py
```diff
@@ -108,8 +128,27 @@ def create(self, request, *args, **kwargs):
...
+
+    def put(self, request, *args, **kwargs):
+        """Update annotation value(s).
+
+        Create one or more annotation values. The request data can be either dict or
+        list (in case of bulk create).
+        """
+        # Call create, which calls _bulkcreate which will also allow update, since our
+        # method is PUT and not POST:
+        response = self.create(request, *args, **kwargs)
```
dblenkus, line 152: Extract the create logic into a separate method and do not modify the response.

resolwe/flow/views/annotations.py
```diff
@@ -83,21 +79,45 @@ class AnnotationValueViewSet(
...
+            try:
+                value.validate()
+            except ValidationError as e:
+                validation_errors.append(e)
+        if validation_errors:
+            raise ValidationError(validation_errors)
+
+    def _bulk_create(self, annotation_values: Iterable[AnnotationValue]):
+        """Bulk create or update annotation values."""
+        also_update = self.request.method == "PUT"
```
dblenkus, line 98: This should be passed from the `create`/`update` method.

resolwe/flow/views/annotations.py
```diff
@@ -83,21 +79,45 @@ class AnnotationValueViewSet(
...
+        inserted_values: list[AnnotationValue] = []
+        if also_update:
+            for annotation_value in annotation_values:
+                inserted_values.append(
+                    AnnotationValue.objects.update_or_create(
```
dblenkus, line 103: Can you use `bulk_create` and set `update_conflicts` to `True`?

resolwe/flow/views/annotations.py
```diff
@@ -108,8 +128,27 @@ def create(self, request, *args, **kwargs):
             entity.has_permission(Permission.EDIT, request.user) for entity in entities
         ):
             raise exceptions.PermissionDenied()
-        self.perform_create(serializer)
+
+        # Create annotation values.
```
dblenkus, line 132: Why don't you override the `perform_create` method to keep the structure consistent?

resolwe/flow/serializers/annotations.py
gregorjerse: You mean a description? Added.

resolwe/flow/views/annotations.py
gregorjerse: I could, but then it would not return primary keys of inserted/updated objects.

resolwe/flow/serializers/annotations.py
```diff
@@ -94,13 +98,63 @@ class Meta:
...
+        for annotations_exist, group in groupby(
+            validated_data,
+            lambda data: (data["field"], data["entity"]) in annotation_map,
+        ):
+            # To update or delete, depending if the value is None.
+            if annotations_exist:
+                for value in group:
+                    instance = annotation_map[(value["field"], value["entity"])]
+                    to_delete = value["_value"]["value"] is None
+                    if to_delete:
```
dblenkus, line 136: You can write the expression inline and avoid using another variable.

resolwe/flow/views/annotations.py
gregorjerse: OK.

resolwe/flow/views/annotations.py
gregorjerse: Old.

resolwe/flow/views/annotations.py
gregorjerse: Old.

resolwe/flow/views/annotations.py
gregorjerse: Refactored.

dblenkus (approved):

PR #1078, opened 2023-12-04
Propagate modify time from object to containers

gregorjerse:
Modifications propagate from:
* data -> entity -> collection
* annotation value -> entity
* relation -> collection

REID-2116


resolwe/flow/models/entity.py
```diff
@@ -377,6 +377,13 @@ def validate_annotations(self):
         if validation_errors:
             raise ValidationError(validation_errors)
 
+    def save(self, *args, **kwargs):
+        """Propagate the modified time to the collection."""
+        super().save(*args, **kwargs)
+        if self.collection:
+            self.collection.modified = self.modified
```
dblenkus, line 384: Does `self.modified` contain updated value?

resolwe/flow/models/data.py
```diff
@@ -570,6 +571,12 @@ def save(self, render_name=False, *args, **kwargs):
 
         with transaction.atomic():
             self._perform_save(*args, **kwargs)
+            if self.entity:
+                self.entity.modified = now()
```
dblenkus, line 575: Use `self.modified`.

resolwe/flow/models/data.py
```diff
@@ -570,6 +571,12 @@ def save(self, render_name=False, *args, **kwargs):
...
+                self.entity.save(update_fields=["modified"])
+            elif self.collection:
+                self.collection.modified = now()
```
dblenkus, line 578: `self.modified`

resolwe/flow/models/entity.py
gregorjerse: This is a good question. I belive it does, at least according to tests.

resolwe/flow/models/data.py
gregorjerse: Done.

resolwe/flow/models/data.py
gregorjerse: Done.

dblenkus (approved):

PR #1080, opened 2023-12-04
Fix crash when checking model-level permissions

gregorjerse:
The query is proxyed to Django default auth backend for staff users and denied for everybody else.

### DIFF
--- a/resolwe/permissions/permissions.py
+++ b/resolwe/permissions/permissions.py
@@ -35,7 +35,10 @@ class ResolwePermissionBackend:
         return None
 
     def has_perm(
-        self, user: User, permission: Permission, obj: Optional[models.Model] = None
+        self,
+        user: User,
+        permission: Permission | str,
+        obj: Optional[models.Model] = None,
     ) -> bool:
         """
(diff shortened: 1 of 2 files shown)


resolwe/permissions/permissions.py
```diff
@@ -48,9 +54,11 @@ def has_perm(
         if not user.is_active:
             return False
 
-        if not model_has_permissions(obj):
-            return True
+        # Object level permissions are handled by the django model backend.
```
dblenkus, line 57: This is in our case, but is not true in general

resolwe/permissions/permissions.py
```diff
@@ -18,6 +18,9 @@ class ResolwePermissionBackend:
     supports_anonymous_user = True
     supports_inactive_user = True
 
+    def __init__(self) -> None:
```
dblenkus, line 21: Why do you need this?

dblenkus (approved):

resolwe/permissions/permissions.py
gregorjerse: Forgot to remove.

PR #1098, opened 2024-02-05
Silently ignore nonexisting/without view permission objects when bulk deleting

gregorjerse:
REID-2208


resolwe/flow/views/mixins.py
```diff
@@ -166,13 +166,6 @@ def bulk_delete(self, request, *args, **kwargs):
                 f"Model {to_delete.model} does not support background deletion."
             )
 
-        permissions_to_delete = set(to_delete.values_list("pk", flat=True))
```
dblenkus, line 169: It is ok to skip objects that don't exist (user has no `VIEW` permission), but it should raise an error for objects with `VIEW` permission and no `EDIT` permission.

dblenkus (approved):

PR #1101, opened 2024-02-13
WIP Add unaccented full text search configuration + search in annotation

gregorjerse:
REID-2214


gregorjerse:
Full text search features must be defined before this can be fixed.

PR #1102, opened 2024-02-13
Add owner reference to PVC claim

gregorjerse:
And remove unnecessary cleanup code.


resolwe/flow/managers/workload_connectors/kubernetes.py
```diff
@@ -860,6 +863,32 @@ def start(self, data: Data, listener_connection: Tuple[str, str, str]):
...
+                        }
+                    ],
+                }
+            )
+
+        if patch:
+            # Patch the PVC with ownerReference. This ensures the PVC claim gets deleted
+            # when the job has completed or failed. The PVCs are deleted after
+            # ttlSecondsAfterFinished seconds, declared in Job spec.
+            core_api.patch_namespaced_persistent_volume_claim(
```
acopar, line 888: Patch is a list, containing typically one patch for each claim. I think you need to add iteration over all patches.

resolwe/flow/managers/workload_connectors/kubernetes.py
```diff
@@ -860,6 +863,32 @@ def start(self, data: Data, listener_connection: Tuple[str, str, str]):
             "It took {:.2f}s to send config to kubernetes".format(end_time - start_time)
         )
 
+        # Patch the PVC-s with owner references to the job.
+        patch = []
+        for claim_name in created_claim_names:
```
acopar, line 868: This is wrong, since it will add two copies of the same ownerReference. You only need a list containing one Job owner reference and apply it to each PVC claim.

acopar: To clarify, the list has to remain, since ownerReferences expects a list. The object in the kubectl yaml would look like this:
```
  ownerReferences:
  - apiVersion: batch/v1
    kind: Job
    name: job-name
    uid: some-uid
```

resolwe/flow/managers/workload_connectors/kubernetes.py
gregorjerse: Done.

resolwe/flow/managers/workload_connectors/kubernetes.py
gregorjerse: Done.

gregorjerse (commented):
Done.

acopar:
Please rebase

PR #1105, opened 2024-02-20
Migrate CI workflow to self hosted runners 

plojyon:
REID-2174


.github/workflows/ci.yml
```diff
@@ -12,90 +12,125 @@ on:
...
     steps:
       - uses: actions/checkout@v4
-      - name: Set up Python 3.11
-        uses: actions/setup-python@v5
-        with:
-          python-version: "3.11"
+
       - name: Install build
-        run: python -m pip install --user build
+        run: python3.12 -m pip install --user build
```
gregorjerse, line 123: Are you sure it is OK to use the Python version that was not used in tests to create a build?

gregorjerse (approved):

PR #1116, opened 2024-04-26
Use unaccented full-text-search configuration

gregorjerse:


resolwe/flow/migrations/0001_squashed_0043_full_text_search.py
```diff
@@ -27,6 +32,23 @@ class Migration(migrations.Migration):
     operations = [
         CITextExtension(),
         TrigramExtension(),
+        UnaccentExtension(),
+        migrations.RunSQL(
```
dblenkus, line 36: Why is this configuration duplicated? In general, you should leave existing migrations as is.

docs/CHANGELOG.rst
```diff
@@ -10,6 +10,10 @@ This project adheres to `Semantic Versioning <http://semver.org/>`_.
 Unreleased
 ==========
 
+Added
+-----
+- Add unaccented full text search configuration ``simple_unaccent``
```
dblenkus, line 15: This line feels unnecessary. Include it in the line below.

resolwe/flow/migrations/0001_squashed_0043_full_text_search.py
gregorjerse: I have removed it from the base migration.

docs/CHANGELOG.rst
gregorjerse: I have merged commits that create and use this configuration.

dblenkus (approved):

PR #1118, opened 2024-05-09
Permission refresh

gregorjerse:
Make it easier to add permissions to objects based on related objects with permissions.

For instance: the storage object is related to the data object so it can inherit the permissions from it.


resolwe/flow/views/annotations.py
```diff
@@ -52,6 +52,11 @@ class AnnotationPresetViewSet(
 
     def get_queryset(self):
         """Get the presets for the current user."""
+        qs = super().get_queryset()
+        print("Getting qs")
```
dblenkus: Remove prints.

dblenkus (approved):

PR #1123, opened 2024-05-27
Add encryption between listener and processes

gregorjerse:


resolwe/flow/executors/zeromq_utils.py
```diff
@@ -72,3 +73,20 @@ def __init__(
...
+    _instance = None
+
+    @classmethod
+    def has_instance(cls):
+        """Check if the instance exists."""
+        return cls._instance is not None
+
+    def __new__(cls):
+        """Overide new contructor to make it singleton."""
+        if cls._instance is None:
```
dblenkus: Use `if not cls.has_instance():`.

resolwe/flow/managers/listener/listener.py
```diff
@@ -560,11 +559,12 @@ def __init__(
         if zmq_socket is None:
             zmq_context: zmq.asyncio.Context = zmq.asyncio.Context.instance()
             # Start the authenticator.
-            auth = AsyncioAuthenticator(zmq_context)
-            auth.start()
-            auth.configure_curve_callback(
-                domain="*", credentials_provider=CurveCallback()
-            )
+            if ZMQAuthenticator.has_instance() is None:
```
dblenkus, line 622: Why do you check whether the class was initialized? It defeats the purpose of making the class a singleton.

resolwe/flow/managers/workload_connectors/local.py
```diff
@@ -35,5 +37,12 @@ def submit(self, data: Data, argv):
                 repr(argv),
             )
         )
+        process_environment = os.environ.copy()
+        print("LPK" * 10, LISTENER_PUBLIC_KEY)
```
dblenkus, line 41: Remove print.

gregorjerse: Fixed.

resolwe/flow/managers/listener/listener.py
gregorjerse: This was added since tests were failing due to the env bug. Removed.

dblenkus (approved):
Add a change log.

PR #1135, opened 2024-07-01
Listener metrics

gregorjerse:


resolwe/flow/managers/listener/metrics.py
```diff
@@ -0,0 +1,179 @@
...
+from resolwe.flow.executors.socket_utils import (
+    Message,
+    MessageProcessingCallback,
+    MessageProcessingEventType,
+)
+
+logger = logging.getLogger(__name__)
+
+# How often to export metrics (in milliseconds).
+METRICS_EXPORT_INTERVAL = 60000
```
acopar, line 23: We typically use granularity of 30s for other metrics.

resolwe/flow/managers/listener/metrics.py
```diff
@@ -0,0 +1,179 @@
...
+            name=f"{self.metrics_prefix}_messages_processing",
+            description="Number of currently processing messages.",
+        )
+        self._preparation_time_histogram = meter.create_histogram(
+            name=f"{self.metrics_prefix}_preparation_duration",
+            description="Response time for messages.",
+            unit="seconds",
+        )
+        self._response_time_histogram = meter.create_histogram(
+            name=f"{self.metrics_prefix}_response_time_duration",
```
acopar, line 87: I would use either `time` or `duration`, not `time_duration` because it seems redundant. Note that in grafana, `seconds` (unit name), will be appended as well.

resolwe/flow/managers/listener/metrics.py
gregorjerse: Changed to 30 seconds.

resolwe/flow/managers/listener/metrics.py
gregorjerse: OK, I have decided on `duration`.

dblenkus (approved):

PR #1138, opened 2024-07-10
Add background job details to audit log

gregorjerse:


gregorjerse:
This will be solved in the background worker itself to also handle any errors encountered during the background job.

PR #1163, opened 2024-11-07
Feature/entity status

gregorjerse:


resolwe/flow/models/base.py
```diff
@@ -18,6 +20,40 @@
...
+    PROCESSING = "PR"
+    #: data object is done
+    DONE = "OK"
+    #: data object is in error state
+    ERROR = "ER"
+    #: data object is in dirty state
+    DIRTY = "DR"
+
+    @classmethod
+    def collection_status_order(cls):
```
dblenkus, line 45: This is not only for collections but also for entities. And, it could also be used to sort data objects themselves.

dblenkus (approved):

resolwe/flow/models/base.py
gregorjerse: I agree... Should I rename it to the `status_sorting_order` ?

dblenkus: It is already in `DataStatus` class, so I would prefer just `sort_order`.

gregorjerse: OK.

PR #1196, opened 2025-01-31
Speed up retrieving latest version of objects

gregorjerse:
REID-2683
REID-2648

when using history


gregorjerse:
Deprecated, see #1198

PR #1202, opened 2025-02-18
Dynamic resources

plojyon:
REID-2461

### DIFF
--- a/resolwe/flow/models/process.py
+++ b/resolwe/flow/models/process.py
@@ -9,6 +9,7 @@ from django.core.validators import RegexValidator
 from django.db import models
 
 from resolwe.permissions.models import PermissionObject
+from resolwe.process.resources import PROCESS_RESOURCES
 
 from .base import BaseManagerWithoutVersion, BaseModel
 from .data import Data
@@ -221,11 +222,21 @@ class Process(BaseModel, PermissionObject):
             if environment_settings.get(resource, {}).get(self.slug)
         }
 
+        dynamic_resources = {}
+        if data is not None:
+            estimators = PROCESS_RESOURCES.get(self.slug, {})
+            dynamic_resources = {
+                resource: estimators[resource](data)
+                for resource in resources
+                if resource in estimators
+            }
+
         # Gather requirements for all resources from all sources.
         # The order of requirements determines their priority.
         resources_map = ChainMap(
             data.process_resources if data is not None else {},
             environment_resources,
+            dynamic_resources,
             self.requirements.get("resources", {}),
             getattr(settings, "FLOW_PROCESS_RESOURCE_DEFAULTS", {}),
             fallback,
--- a/resolwe/flow/models/process.py
+++ b/resolwe/flow/models/process.py
@@ -204,7 +204,7 @@ class Process(BaseModel, PermissionObject):
               not otherwise specified in the resource requirements.
             - ``cores``: Core count limit. Defaults to 1.
             - ``storage``: Size (in gibibytes) of temporary volume used for
-              processing in kubernetes. Defaults to 200.
+              processing in kubernetes. Defaults to 10.
 
         :rtype: dict
         """
--- a/resolwe/process/resources.py
+++ b/resolwe/process/resources.py
@@ -1,0 +1,16 @@
+"""Resource consumption estimators for processes."""
+
+from collections import defaultdict
+
+PROCESS_RESOURCES = defaultdict(dict)
+
+
+def estimator(process_slug, resource):
+    """Register resource consumption estimator."""
+
+    def decorator(func):
+        """Register resource consumption estimator."""
+        PROCESS_RESOURCES[process_slug][resource] = func
+        return func
+
+    return decorator
--- a/resolwe/process/tests/test_python_process.py
+++ b/resolwe/process/tests/test_python_process.py
@@ -1,4 +1,5 @@
 # pylint: disable=missing-docstring
+import math
 import os
 import sys
 import unittest
@@ -686,7 +687,7 @@ class PythonProcessRequirementsTest(ProcessTestCase):
                 ]
             )
             # Assert iterate method was called the correct number of times.
-            self.assertEqual(calls, Process.objects.count() // chunk_size)
+            self.assertEqual(calls, math.ceil(Process.objects.count() / chunk_size))
             # Assert the result is correct.
             process_slugs = Process.objects.all().values_list("slug", flat=True)
             self.assertCountEqual(process_slugs, data.output["process_slugs"])
--- a/resolwe/process/tests/test_python_process.py
+++ b/resolwe/process/tests/test_python_process.py
@@ -30,6 +30,7 @@ from resolwe.flow.models.annotations import (
     AnnotationValue,
 )
 from resolwe.permissions.models import Permission, get_anonymous_user
+from resolwe.process import resources
 from resolwe.test import (
     ProcessTestCase,
     tag_process,
@@ -661,6 +662,33 @@ class PythonProcessRequirementsTest(ProcessTestCase):
         self.assertEqual(data.output["cores"], 3)
         self.assertEqual(data.output["memory"], 50000)
         self.assertEqual(data.output["storage"], 500)
+
+    @with_docker_executor
+    @tag_process("test-python-process-requirements2")
+    def test_dynamic_resources(self):
+        data = self.run_process("test-python-process-requirements2")
+        data.refresh_from_db()
+        self.assertEqual(data.status, "OK")
+
+        self.assertEqual(data.output["cores"], 2)
+        self.assertEqual(data.output["memory"], 4096)
+        self.assertEqual(data.output["storage"], 200)
+
+        @resources.estimator("test-python-process-requirements2", "cores")
+        def cores_estimator(data):
+            return 3
+
+        @resources.estimator("test-python-process-requirements2", "storage")
+        def storage_estimator(data):
+            return 3.5e9
+
+        data = self.run_process("test-python-process-requirements2")
+        data.refresh_from_db()
+        self.assertEqual(data.status, "OK")
+
+        self.assertEqual(data.output["cores"], 3)
+        self.assertEqual(data.output["memory"], 4096)  # default
+        self.assertEqual(data.output["storage"], 3.5e9)
 
     @with_docker_executor
     @tag_process("test-python-process-iterate")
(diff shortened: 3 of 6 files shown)


resolwe/flow/models/utils/dynamic_resources.py
```diff
@@ -0,0 +1,103 @@
...
+        "LOAD_CONST",  # constants
+        "LOAD_NAME",  # variables
+        "CALL_FUNCTION",  # provided functions, e.g. log
+        "RETURN_VALUE",  # implicit return
+        "RETURN_CONST",  # implicit return (alt.)
+    }
+    try:
+        compiled = compile(expr, "", "eval")
+    except SyntaxError:
+        raise ValueError(f"{expr} is not a valid expression")
```
gregorjerse, line 34: Add quotes arount `expr`.

resolwe/flow/models/utils/dynamic_resources.py
```diff
@@ -0,0 +1,103 @@
...
+
+    for instruction in dis.get_instructions(expr):
+        opcode = dis.opname[instruction.opcode]
+        if opcode not in allowed_codes:
+            raise ValueError(f"Operation {opcode} not allowed (in {expr})")
```
gregorjerse, line 39: This only returns the first not-allowed code.

How about something like

```
if not_allowed := ({instruction.opname for instruction in dis.get_instructions(expr)} - allowed_codes):
    raise ValueError(f"Operations '{not_allowed}' are not allowed in '{expr}'.")
```

resolwe/flow/models/utils/dynamic_resources.py
```diff
@@ -0,0 +1,103 @@
...
+
+    scope = {"log": math.log, "exp": math.exp, "sqrt": math.sqrt}
```
gregorjerse, line 41: ```scope = {...} | variables |{"__builtins__":{}} ```

resolwe/flow/models/process.py
```diff
@@ -226,7 +228,7 @@ def get_resource_limits(self, data: Optional[Data] = None):
         resources_map = ChainMap(
             data.process_resources if data is not None else {},
             environment_resources,
-            self.requirements.get("resources", {}),
+            get_dynamic_resource_limits(data) if data is not None else {},
```
gregorjerse, line 231: Missing the process argument.

resolwe/flow/models/utils/dynamic_resources.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # Get variables used in the expression.
+        used_variables = {
+            node.id for node in ast.walk(ast.parse(formula)) if type(node) is ast.Name
+        }
+
+        # Find the values of the used variables.
+        try:
+            variables = {
+                var: _resolve_value(var, data.input, process.input_schema)
```
gregorjerse, line 93: Use `variable` instead of `var`.

resolwe/flow/models/utils/dynamic_resources.py
```diff
@@ -0,0 +1,103 @@
...
+                for var in used_variables
+            }
+        except ValueError as exc:
+            raise ValueError(f"{exc} in requirements for {resource}: {formula}")
+
+        amount = _eval_dynamic_requirements(process, formula, variables)
```
gregorjerse, line 99:
```
if amount := ...:
```

resolwe/flow/models/utils/dynamic_resources.py
```diff
@@ -0,0 +1,103 @@
...
+    return field
+
+
+def _resolve_value(name: str, input_values: dict, input_schema: list):
+    """Resolve the value of a variable in the expression."""
+
+    try:
+        keys = name.split("__")
+        value = reduce(dict.get, keys, input_values)
+        type_ = reduce(schema_get, keys, input_schema)["type"]
```
gregorjerse, line 61: The method `dict_dot` may come in handy.

https://github.com/genialis/resolwe/blob/03b54cc0630b2d37f6dfa3c6cf912ed6908ff7d4/resolwe/flow/utils/__init__.py#L52

resolwe/flow/models/utils/dynamic_resources.py
```diff
@@ -0,0 +1,103 @@
...
+    except (KeyError, StopIteration):
+        raise ValueError(f"Unrecognized variable {name}")
+
+    # More types may be supported as needed.
+    if type_ == "data:":
+        return Data.objects.get(pk=value).size
```
gregorjerse, line 67: To avoid reading all the data and creating the entire object just get the size from the database.

`Data.objecs.filter(pk=value).values_list("size", flat=True).get()`

gregorjerse: Maybe you could use the `match` statement?
```
match type_:
    case "data:":
        return Data...
    case "basic:integer" | "basic_float" ... : 
        return value
    case _:
        raise ...
```

resolwe/process/tests/test_python_process.py
```diff
@@ -661,6 +661,27 @@ def test_resource_data_override(self):
...
+                "integer_input": 3,
+                "my_group": {"foo": "burek", "subgroup": {"bar": -2}},
+            },
+        )
+        data.refresh_from_db()
+        self.assertEqual(data.status, "OK")
+
+        self.assertEqual(data.output["cores"], 1)
+        self.assertEqual(data.output["memory"], 120)
+        self.assertEqual(data.output["storage"], 3036)
```
gregorjerse, line 683: 3306 = ?

resolwe/flow/models/utils/dynamic_resources.py
```diff
@@ -0,0 +1,103 @@
...
+    elif type_ in ("basic:integer:", "basic:float:", "basic:boolean:"):
+        return value
+
+    raise ValueError(f"Unsupported type {type_}")
+
+
+def get_dynamic_resource_limits(process: Process, data: Data):
```
gregorjerse, line 74: Not sure if this belongs here or in the process class. 

I do not see any added value by having these methods in the `utils` file, since they will probably only be used in the `process.py`  model file.

resolwe/flow/models/utils/dynamic_resources.py
```diff
@@ -0,0 +1,103 @@
...
+import dis
+import math
+from functools import reduce
+
+from resolwe.flow.models import Data, Process
+
+
+def _eval_dynamic_requirements(expr: str, variables: dict):
+    """Evaluate an arithmetic expression."""
+    allowed_codes = {
```
gregorjerse, line 13: I am just curious: how would you implement requirements for the `STAR` aligner process using these codes?

resolwe/flow/models/process.py
```diff
@@ -226,7 +228,7 @@ def get_resource_limits(self, data: Optional[Data] = None):
         resources_map = ChainMap(
             data.process_resources if data is not None else {},
             environment_resources,
-            self.requirements.get("resources", {}),
```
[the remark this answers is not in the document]
plojyon: This field is handled by the new function, `get_dynamic_resource_limits`. It can now be a string expression.

resolwe/flow/models/utils/dynamic_resources.py
plojyon: It can't replace the second `reduce`, and I'd prefer to keep the symmetry of the two lines.

resolwe/flow/models/utils/dynamic_resources.py
plojyon: While it's true the functions won't be reused, I feel keeping the expression evaluation logic separate simplifies maintenance by isolating complex calculation logic from the core ORM functionalities of the `Process` class, as they are only marginally related.

resolwe/flow/models/utils/dynamic_resources.py
plojyon: Top 3 processes:

# rnaseq-vc-preprocess
`10995780188 * bam - 189023487527`

# vc-gatk4-hc
`1.007 * alignment - 12.4`

# alignment-star
`(genome == 14477317829)*(reads + 1.5e10) + (genome != 14477317829)*16`

resolwe/flow/models/process.py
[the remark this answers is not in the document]
gregorjerse: My comment was connected with a fact that only the data object was passed to the `get_dynamic_resource_limits` , which was clearly a bug which is now fixed.

resolwe/flow/models/utils/dynamic_resources.py
gregorjerse: These formulas actually look like black magick to me. My hopes are bioinformaticians can add resources by themselves and that they are readable even after a year (when all the magick numbers have evaporated from my memory).

resolwe/flow/models/utils/dynamic_resources.py
gregorjerse: OK.

resolwe/flow/models/utils/dynamic_resources.py
```diff
@@ -0,0 +1,111 @@
...
+        "exp": math.exp,
+        "sqrt": math.sqrt,
+        "__builtins__": {},
+    } | variables
+    return eval(compiled, scope)
+
+
+def schema_get(schema: list, key: str) -> dict | list:
+    """Return a schema field with the given name."""
+    field = next(iter(filter(lambda field: field["name"] == key, schema)))
```
gregorjerse, line 56: The `filter` statement returns an iterator, so `iter` is redundant.

resolwe/process/resources.py
```diff
@@ -0,0 +1,16 @@
+"""Resource consumption estimators for processes."""
+
+from collections import defaultdict
+
+PROCESS_RESOURCES = defaultdict(dict)
+
+
+def estimator(process_slug, resource):
```
gregorjerse, line 8: In the rest of the code the `wrapt.decorator` is usually used to avoid the inner method.

resolwe/process/tests/test_python_process.py
```diff
@@ -686,7 +718,7 @@ def test_python_process_iterate(self):
                 ]
             )
             # Assert iterate method was called the correct number of times.
-            self.assertEqual(calls, Process.objects.count() // chunk_size)
+            self.assertEqual(calls, math.ceil(Process.objects.count() / chunk_size))
```
gregorjerse, line 721: Why is this change necessary? If it fixes a bug it does belong into a new commit.

resolwe/process/resources.py
plojyon: does not apply here

PR #1216, opened 2025-05-08
Generic assign annotation to collection

gregorjerse:
Make method general so it can be used to assign prediction fields also.

REID-2753


resolwe/flow/serializers/annotations.py
```diff
@@ -77,7 +77,6 @@ class AnnotationFieldDictSerializer(serializers.Serializer):
     annotation_fields = PrimaryKeyDictRelatedField(
         queryset=AnnotationField.objects.all(), many=True
     )
-    confirm_action = serializers.BooleanField(default=False)
```
dblenkus, line 80: Add a change log.

gregorjerse: Added.

dblenkus (approved):

